PR #236, opened 2025-03-17
Fix resume payload path validation in loop executor

MarceloRGonc:
Fixes OPS-1360.

## Additional Notes

When resuming a workflow with a loop, we need to validate whether we are resuming one iteration or a block outside the loop. This PR fixes the validation if a loop's actionName belongs to the path to be resumed or not.

The path is a string like these:
- 'step_1,0.step_3' -> resume a iteration
- 'step_10' -> resume a block


bigfluffycookie (approved):
And this works no matter how nested the flow is?

MarceloRGonc:
> And this works no matter how nested the flow is?

Yes. The problem was that if we had a workflow like the one below. If the workflow was paused at block 'step_23', upon resuming, we were resuming the loop ('step_2'), even though the loop had already finished.

- loop: step_2
  - ...
- step_23

PR #245, opened 2025-03-18
Add action to return the link to an OpenOps table

rita-gorokhod:
Fixes OPS-1373
image

### DIFF
--- a/packages/blocks/openops-tables/src/actions/get-table-url-action.ts
+++ b/packages/blocks/openops-tables/src/actions/get-table-url-action.ts
@@ -1,0 +1,27 @@
+import { BlockAuth, createAction } from '@openops/blocks-framework';
+import {
+  getTableIdByTableName,
+  openopsTablesDropdownProperty,
+} from '@openops/common';
+import { SharedSystemProp, system } from '@openops/server-shared';
+
+const DEFAULT_DATABASE_ID = 1;
+
+export const getTableUrlAction = createAction({
+  auth: BlockAuth.None(),
+  name: 'get_table_url',
+  description: 'Get a shareable URL for the provided table.',
+  displayName: 'Get Table URL',
+  props: {
+    tableName: openopsTablesDropdownProperty(),
+  },
+  async run(context) {
+    const tableName = context.propsValue.tableName as unknown as string;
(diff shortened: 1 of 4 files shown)


packages/blocks/openops-tables/src/actions/get-table-url-action.ts
```diff
@@ -0,0 +1,27 @@
...
+  props: {
+    tableName: openopsTablesDropdownProperty(),
+  },
+  async run(context) {
+    const tableName = context.propsValue.tableName as unknown as string;
+    const tableId = await getTableIdByTableName(tableName);
+    const baseUrl = system.getOrThrow(SharedSystemProp.FRONTEND_URL);
+
+    return (
+      baseUrl + `/tables?path=/database/${DEFAULT_DATABASE_ID}/table/${tableId}`
```
rita-gorokhod, line 24: same logic as we do
https://github.com/openops-cloud/openops/blob/ad603dad8ae2d77d8d65f79577814105088bb735/packages/react-ui/src/app/routes/home-demo/index.tsx#L44

This will allow sending a link on slack that leads to our full app, and the table opened
image

packages/blocks/openops-tables/src/actions/get-table-url-action.ts
```diff
@@ -0,0 +1,27 @@
+import { BlockAuth, createAction } from '@openops/blocks-framework';
+import {
+  getTableIdByTableName,
+  openopsTablesDropdownProperty,
+} from '@openops/common';
+import { SharedSystemProp, system } from '@openops/server-shared';
+
+const DEFAULT_DATABASE_ID = 1;
```
MarceloRGonc, line 8: We have a method to get the default database id "getDefaultDatabaseId".

MarceloRGonc (approved):

packages/blocks/openops-tables/src/actions/get-table-url-action.ts
rita-gorokhod: As we discussed offline, it will do a request, and I want to avoid doing that in the blocks until we have caching for it and/or support multiple dbs in tables.

PR #249, opened 2025-03-18
Add copy paste to step context menu

alexandrudanpop:
Fixes OPS-860

## Additional Notes

Extracted context menu to a new component.

## Testing Checklist

Check all that apply:

- [x] Changes are backwards compatible with any existing data

image

### DIFF
--- a/packages/react-ui/src/app/features/builder/flow-canvas/context-menu/canvas-context-menu.tsx
+++ b/packages/react-ui/src/app/features/builder/flow-canvas/context-menu/canvas-context-menu.tsx
@@ -1,3 +1,4 @@
+import { flagsHooks } from '@/app/common/hooks/flags-hooks';
 import {
   Button,
   DropdownMenuContent,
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/app/features/builder/flow-canvas/context-menu/canvas-context-menu.tsx
```diff
@@ -0,0 +1,196 @@
...
+      return applyOperationAndPushToHistory(
+        {
+          type: FlowOperationType.DUPLICATE_ACTION,
+          request: {
+            stepName: data.step.name,
+          },
+        },
+        () => toast(UNSAVED_CHANGES_TOAST),
+      );
+    };
```
alexandrudanpop, line 86: moved from step node

packages/react-ui/src/app/features/builder/flow-canvas/nodes/step-node.tsx
```diff
@@ -171,6 +122,10 @@ const WorkflowStepNode = React.memo(
       e.stopPropagation();
     };
 
+    const blockSelectorOperation = isAction
+      ? FlowOperationType.UPDATE_ACTION
+      : FlowOperationType.UPDATE_TRIGGER;
```
alexandrudanpop, line 127: this was previously a ref, and updated from context menu, I don't think we need that behavior since we can just compute the operation here

PR #250, opened 2025-03-18
Add action for azure advisor get cost recommendations

bigfluffycookie:
Fixes OPS-1380

## Additional Notes

Adds a new action to fetch cost recommendations from azure advisor via the CLI.

image


packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,87 @@
...
+  auth: azureAuth,
+  name: 'advisor',
+  description: 'Get Azure Advisor Cost Recommendations',
+  displayName: 'Get Advisor Cost Recommendations',
+  props: {
+    useHostSession: useHostSession,
+    subscriptions: subDropdown,
+    ids: Property.Array({
+      displayName: 'Resource IDs',
+      description:
```
bigfluffycookie, line 17: copied the description from azure directly

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,87 @@
...
+        context.propsValue.subscriptions?.['subDropdown'],
+      );
+
+      try {
+        const jsonObject = JSON.parse(result);
+        return jsonObject;
+      } catch (error) {
+        return result;
+      }
+    } catch (error) {
```
bigfluffycookie, line 51: i know this is very similar to the other action, but as we arent sure if we will keep using the cli in the backend or if we will use the azure API in the future i didnt feel like it makes sense yet to merge the logic, if we add a third action that needs this we can do it then

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,87 @@
...
+      throw new Error(message);
+    }
+  },
+});
+
+function buildCommand(propsValue: any) {
+  const { ids, resourceGroup } = propsValue;
+  let command = `az advisor recommendation list --category 'cost' --output json`;
+
+  if (ids && ids.length > 0 && resourceGroup) {
```
MarceloRGonc, line 72: You can do 'ids?.length'

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,87 @@
...
+    throw new Error(
+      'Resource IDs and Resource Group cannot be specified together. Please only specify one of them.',
+    );
+  }
+
+  if (ids && ids.length > 0) {
```
MarceloRGonc, line 78: Same here.

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const advisorAction = createAction({
+  auth: azureAuth,
+  name: 'advisor',
+  description: 'Get Azure Advisor Cost Recommendations',
+  displayName: 'Get Advisor Cost Recommendations',
+  props: {
+    useHostSession: useHostSession,
+    subscriptions: subDropdown,
+    ids: Property.Array({
```
MarceloRGonc, line 15: I think it's better to call it 'resourceIds'

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,87 @@
...
+      description:
+        'One or more resource IDs (space-delimited). If provided, no other "Resource Id" arguments should be specified.',
+      required: false,
+    }),
+    resourceGroup: Property.ShortText({
+      displayName: 'Resource Group',
+      description: 'Name of a resource group.',
+      required: false,
+    }),
+    dryRun: dryRunCheckBox(),
```
MarceloRGonc, line 26: Was it discussed if we need a dry run on this action?

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
rita-gorokhod: I have to say I don't get it.

bigfluffycookie: yea i think what they mean is no other resource related arguments should be filled out. 

Maybe "If provided, cannot also provide 'resource group' " and the same for resource group?

MarceloRGonc (approved):

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
```diff
@@ -0,0 +1,81 @@
...
+        return result;
+      }
+    } catch (error) {
+      logger.error('Failed to fetch Azure cost recommendations', {
+        command: command,
+        error: error,
+      });
+      let message =
+        'An error occurred while fetching Azure cost recommendations: ';
+      if (String(error).includes('login --service-principal')) {
```
rita-gorokhod, line 52: kinda ugly for now but I see your point, let's see once we get a third action that needs it

rita-gorokhod (commented):
See comments

packages/blocks/azure/src/lib/actions/azure-advisor-action.ts
bigfluffycookie: It is mutually exclusive, but i fear a toggle might be a bit confusing no? Wed still have to add a description of why we toggle it

rita-gorokhod: Hmm, I imagined something like "Filter by Resource IDs" vs "Filter by Resource Group".

bigfluffycookie: both can be empty yes

PR #255, opened 2025-03-18
Add getOrAdd method to redisWrapper

MarceloRGonc:
Part of OPS-1323.

## Additional Notes
This PR adds a wrapper so we can cache the results of the method. It's protected against race conditions.

## Testing Checklist

Check all that apply:


packages/server/shared/src/lib/cache/memory-wrapper.ts
```diff
@@ -41,7 +41,17 @@ async function getSerializedObject<T>(key: string): Promise<T | null> {
   return result ? (JSON.parse(result) as T) : null;
 }
 
+async function getOrSet<T, Args extends unknown[]>(
```
MarceloRGonc, line 44: We don't need this on the memory wrapper

packages/server/shared/src/lib/cache/redis-lock.ts
```diff
@@ -0,0 +1,88 @@
...
+const DEFAULT_RETRY_COUNT = 35;
+const DEFAULT_RETRY_DELAY = 700;
+const DEFAULT_JITTER = 300;
+
+type RedlockRetryConfig = {
+  retryDelay: number;
+  retryJitter: number;
+};
+
+const generateRedlockRetryConfig = (
```
MarceloRGonc, line 18: The timeout depends on the function we are calling, so I added a way to set a custom timeout. However, with that, we need the retry configuration to also change based on the timeout value.

packages/server/shared/src/lib/cache/redis-wrapper.ts
```diff
@@ -40,6 +42,35 @@ async function setSerializedObject<T>(
   await setKey(key, JSON.stringify(obj), expireInSeconds);
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+async function getOrSet<T, Args extends any[]>(
```
MarceloRGonc, line 46: If you have an awesome suggestion for the name of the method feel free to suggest.

packages/server/shared/src/lib/cache/redis-wrapper.ts
```diff
@@ -40,6 +42,35 @@ async function setSerializedObject<T>(
...
+  key: string,
+  fn: (...args: Args) => Promise<T>,
+  args: Args,
+  expireInSeconds?: number,
+): Promise<T> {
+  const value = await getSerializedObject<T>(key);
+
+  if (value) {
```
rita-gorokhod, line 54: value !== undefined

Otherwise this doesn't work for values like false, 0 and ''

packages/server/shared/src/lib/cache/redis-wrapper.ts
rita-gorokhod: getOrAdd ? cosmetics

packages/server/shared/src/lib/cache/redis-wrapper.ts
```diff
@@ -40,6 +42,35 @@ async function setSerializedObject<T>(
...
+    return value;
+  }
+
+  let lock: Lock | undefined;
+  try {
+    lock = await acquireRedisLock(`lock:${key}`);
+    const value = await getSerializedObject<T>(key);
+    if (value) {
```
rita-gorokhod, line 62: same here

packages/server/shared/src/lib/cache/redis-wrapper.ts
MarceloRGonc: Changed. I'm checking for null; that's the value returned by the cache.

rita-gorokhod (approved):
LGTM but the config values are hard coded and it's going to create a problem if someone changes the env vars for timeout, and not know this class exists.

rita-gorokhod (approved):

PR #259, opened 2025-03-19
Handle workflow nodes selection

rSnapkoOpenOps:
Fixes OPS-1378.

## Testing Checklist

Check all that apply:

https://www.loom.com/share/3a101d8018b74ef28c8f8c88da04bde3?sid=697994a0-282f-4214-9d7e-70e1f58581c2


packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -315,6 +315,15 @@ function getSplitBranches(step: SplitActionSchema): SplitBranch[] {
...
 
+function truncateFlow(firstStep: Step, lastStepName: string): Step {
+  if (firstStep.name !== lastStepName && firstStep.nextAction) {
+    truncateFlow(firstStep.nextAction, lastStepName);
+  } else {
+    firstStep.nextAction = undefined;
+  }
+  return firstStep;
+}
+
```
rSnapkoOpenOps, line 326: Truncates next actions by action name

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -30,12 +47,56 @@ export const CanvasContextProvider = ({
...
+      setSelectedNodes(ev.nodes as WorkflowNode[]);
+    }
+  };
+
+  const onSelectionEnd = useCallback(() => {
+    const firstStep = selectedNodes[0]?.data.step;
+    if (!firstStep) return;
+
+    const topLevelSteps = flowHelper.getAllStepsAtFirstLevel(firstStep);
+    if (!topLevelSteps.length) return;
```
rSnapkoOpenOps, line 61: Get all top level steps relative to the first selected step

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -30,12 +47,56 @@ export const CanvasContextProvider = ({
...
+
+    const lastSelectedIndex = selectedNodes.reduceRight(
+      (foundIndex, node, i) =>
+        foundIndex === -1 &&
+        topLevelSteps.some((step) => step.name === node.data.step?.name)
+          ? i
+          : foundIndex,
+      -1,
+    );
```
rSnapkoOpenOps, line 70: Find last selected item on top level

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -30,12 +47,56 @@ export const CanvasContextProvider = ({
...
+      lastSelectedIndex !== -1
+        ? topLevelSteps.slice(0, lastSelectedIndex + 1)
+        : topLevelSteps;
+
+    if (!selectedSteps.length) return;
+
+    const truncatedFlow = flowHelper.truncateFlow(
+      cloneDeep(selectedSteps[0]),
+      selectedSteps[selectedSteps.length - 1].name,
+    );
```
rSnapkoOpenOps, line 82: This can be used in future as copied value

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -30,12 +47,56 @@ export const CanvasContextProvider = ({
...
+
+    const selectedStepNames = flowHelper
+      .getAllSteps(truncatedFlow)
+      .map((step) => step.name);
```
rSnapkoOpenOps, line 86: Calculated all selected steps with children

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -17,6 +32,8 @@ export const CanvasContextProvider = ({
   children: ReactNode;
 }) => {
   const [panningMode, setPanningMode] = useState<PanningMode>('grab');
+  const [selectedNodes, setSelectedNodes] = useState<WorkflowNode[]>([]);
```
alexandrudanpop, line 35: do we need to store the full WorkflowNode or only the step name?
Also, is there a problem that we don't update the selectedNodes when the workflow changes? (example on deletes)

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -315,6 +315,15 @@ function getSplitBranches(step: SplitActionSchema): SplitBranch[] {
   }));
 }
 
+function truncateFlow(firstStep: Step, lastStepName: string): Step {
+  if (firstStep.name !== lastStepName && firstStep.nextAction) {
+    truncateFlow(firstStep.nextAction, lastStepName);
```
cezudas, line 320: It calls truncateFlow but does not use its return value. Can you return anyway?

rSnapkoOpenOps: Yes, it's fine. We care about top level step only

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -30,12 +47,58 @@ export const CanvasContextProvider = ({
...
+      cloneDeep(selectedSteps[0]),
+      selectedSteps[selectedSteps.length - 1].name,
+    );
+
+    const selectedStepNames = flowHelper
+      .getAllSteps(truncatedFlow)
+      .map((step) => step.name);
+
+    state.addSelectedNodes(selectedStepNames);
+    setSelectedActions([]);
```
rSnapkoOpenOps, line 90: Clean actions on selection end

alexandrudanpop: if we are cleaning this after the selection end can we use ref instead to avoid re-renders? Currently, it's re-rendering quite a lot since we setState in `onSelectionChange`.

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -30,12 +47,58 @@ export const CanvasContextProvider = ({
     return 'grab';
   }, [panningMode, shiftPressed, spacePressed]);
 
+  const onSelectionChange = (ev: OnSelectionChangeParams) => {
+    if (ev.nodes.length) {
+      setSelectedActions(
+        ev.nodes.map((node) => node.data.step).filter(Boolean) as Action[],
+      );
+    }
+  };
```
alexandrudanpop, line 56: we should wrap in `useCallback`

alexandrudanpop (approved):

PR #261, opened 2025-03-19
Add token refresh logic

bigfluffycookie:
Part of OPS-1205

Adds logic to refresh token if it has expired. This is needed for OAUTH2 apps like teams, which has a token expiry of 1 hour.

### DIFF
--- a/packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
+++ b/packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
@@ -443,6 +443,7 @@ function needRefresh(connection: AppConnection): boolean {
   switch (connection.value.type) {
     case AppConnectionType.PLATFORM_OAUTH2:
     case AppConnectionType.CLOUD_OAUTH2:
+      return oauth2Util.isExpired(connection.value);
     case AppConnectionType.OAUTH2:
       return oauth2Util.isExpired(connection.value);
     default:
--- a/packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
+++ b/packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
@@ -31,21 +31,32 @@ async function refresh({
     tokenUrl: connectionValue.token_url,
   };
 
-  const oauthProxyUrl = system.get<string>(
-    SharedSystemProp.INTERNAL_OAUTH_PROXY_URL,
-  );
+  logger.info(`Refreshing token via Lambda for block: ${blockName}`);
 
-  const response = (
-    await axios.post(`${oauthProxyUrl}/refresh`, requestBody, {
-      timeout: 10000,
-    })
-  ).data;
-  return {
-    ...connectionValue,
-    ...response,
-    props: connectionValue.props,
-    type: AppConnectionType.CLOUD_OAUTH2,
-  };
+  try {
+    const oauthProxyUrl = system.get<string>(
+      SharedSystemProp.INTERNAL_OAUTH_PROXY_URL,
+    );
+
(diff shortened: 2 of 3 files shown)


packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
```diff
@@ -443,6 +443,7 @@ function needRefresh(connection: AppConnection): boolean {
   switch (connection.value.type) {
     case AppConnectionType.PLATFORM_OAUTH2:
     case AppConnectionType.CLOUD_OAUTH2:
+      return oauth2Util.isExpired(connection.value);
```
bigfluffycookie, line 446: this is mostly the important change

packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
```diff
@@ -31,21 +31,32 @@ async function refresh({
     tokenUrl: connectionValue.token_url,
   };
 
-  const oauthProxyUrl = system.get<string>(
-    SharedSystemProp.INTERNAL_OAUTH_PROXY_URL,
-  );
+  logger.info(`Refreshing token via Lambda for block: ${blockName}`);
 
-  const response = (
```
rita-gorokhod, line 38: so were we just never calling it?

packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
```diff
@@ -31,21 +31,32 @@ async function refresh({
...
+    ).data;
+
+    return {
+      ...connectionValue,
+      ...response,
+      props: connectionValue.props,
+      type: AppConnectionType.CLOUD_OAUTH2,
+    };
+  } catch (e: unknown) {
+    logger.error('Error refreshing token via Lambda', e);
```
rita-gorokhod, line 54: Error refreshing token for block

and pass the block name in the log too

packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
```diff
@@ -31,21 +31,32 @@ async function refresh({
...
+    throw new ApplicationError({
+      code: ErrorCode.INVALID_CLOUD_REFRESH,
+      params: { blockName },
```
rita-gorokhod, line 57: why not add the e as well?

rita-gorokhod (approved):

packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
bigfluffycookie: we were, but the refresh itself had an empty implementation

packages/server/api/src/app/app-connection/app-connection-service/oauth2/services/cloud-oauth2-service.ts
bigfluffycookie: hmm idk, we dont do it for the claim part either, from what ive seen with the application errors we dont usually add the error to it

packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
rita-gorokhod: you don't need this, the switch has a fallthrough

bigfluffycookie:
I missed the cascading switch, so none of the changes are really needed. Gonna close this and just do the changes in the lambda.

PR #263, opened 2025-03-19
Reduce the number of requests made to Tables

MarceloRGonc:
Part of OPS-1323

## Additional Notes
When we run the update records inside a loop we were doing a lot of requests to tables and receiving the same response. With this PR I'm adding cache to these requests.


bigfluffycookie (commented):
Tests?

packages/blocks/openops-tables/src/actions/delete-record-action.ts
```diff
@@ -25,11 +26,23 @@ export const deleteRecordAction = createAction({
   },
   async run(context) {
     const tableName = context.propsValue.tableName as unknown as string;
-    const tableId = await getTableIdByTableName(tableName);
+    const tableCacheKey = `${context.run.executionCorrelationId}-table-${tableName}`;
+    const tableId = await wrapWithCacheGuard(
+      tableCacheKey,
+      getTableIdByTableName,
```
rita-gorokhod, line 32: proposal: why not do these changes inside the tables API, i.e. in getTableIdByTableName.

i.e. TablesClient, that is a class that gets context.run.executionCorrelationId in the ctor. WDYT? too much?

bigfluffycookie (commented):
Sorry, i didnt mean "Fix tests", i meant please add tests for the new behavior, the table actions are all well tested so I'd like to keep it that way

packages/blocks/openops-tables/src/actions/delete-record-action.ts
MarceloRGonc: We are also using the Tables API in the properties. And in that cases we don't want to use the cache.

We can infer the key from the method and parameters, and in that case we just need the executionCorrelationId or other value to append

packages/blocks/openops-tables/src/actions/update-record-action.ts
```diff
@@ -27,7 +28,7 @@ export const updateRecordAction = createAction({
         'The primary key value of the row to update. If the row does not exist, a new row will be created.',
       required: true,
       refreshers: ['tableName'],
-      props: async ({ tableName }) => {
+      props: async ({ tableName, context }) => {
```
rita-gorokhod, line 31: need it here?

packages/engine/src/lib/handler/context/engine-constants.ts
```diff
@@ -102,7 +103,7 @@ export class EngineConstants {
       input.flowVersion.displayName,
       input.flowVersion.id,
       input.flowVersion.state,
-      'test-run',
+      openOpsId(),
```
rita-gorokhod, line 106: superstore-pray

rita-gorokhod (approved):
🚀

PR #264, opened 2025-03-19
Add "PASTE" workflow action

rSnapkoOpenOps:
Fixes OPS-855.
https://www.loom.com/share/cdcbb460c60f41899ff6240ccaf9e822?sid=5f731557-4710-421a-8366-9a57430a7f3e

## Testing Checklist

Check all that apply:


packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -955,18 +984,18 @@ function duplicateStep(
     });
     return step;
   });
-  let finalFlow = addAction(flowVersionWithArtifacts, {
+  let finalFlow = addAction(flowVersion, {
```
bigfluffycookie, line 987: why was this change necessary? I forgot the difference between flowVersion and flowVersionWithArtifact

rSnapkoOpenOps: A new function, `duplicateStepCascading`, has been introduced with the parameter `flowVersion`.

The logic for this function was copied from the existing `duplicateStep` function, where the corresponding parameter was named `flowVersionWithArtifacts`.

Both parameters are of the `FlowVersion` type.

bigfluffycookie (approved):

PR #265, opened 2025-03-19
Move the telemetry tracker to the server

MarceloRGonc:
Fixes OPS-1402.


bigfluffycookie (approved):

packages/server/api/src/app/core/security/authn/access-token-authn-handler.ts
```diff
@@ -66,21 +66,3 @@ export class AccessTokenAuthnHandler extends BaseSecurityHandler {
     return accessToken;
   }
 }
-
-async function getTrackEventsConfigForUser(userId: string): Promise<string> {
```
rita-gorokhod, line 70: next time please point out if it's moved as is or not :)

packages/server/api/src/app/flows/step-run/step-run-service.ts
```diff
@@ -50,14 +51,18 @@ export const stepRunService = {
...
-    sendWorkflowTestBlockEvent({
-      flowId: flowVersion.flowId,
-      success: result.success,
-      projectId,
-      duration,
-      userId,
-      step,
-    });
+    const trackEventsEnabled = await userService.getTrackEventsConfig(userId);
+    sendWorkflowTestBlockEvent(
```
rita-gorokhod, line 55: instead of passing trackEventsEnabled, I think we can just do an IF here and not call sendWorkflowTestBlockEvent

packages/server/shared/src/lib/telemetry/telemetry.ts
```diff
@@ -99,7 +99,13 @@ function enrichEventLabels(event: TelemetryEvent): Timeseries {
...
 
-function isTelemetryEnabledForCurrentUser(): boolean {
+function isTelemetryEnabledForCurrentUser(
+  trackEventsEnabled?: boolean,
+): boolean {
+  if (trackEventsEnabled) {
+    return true;
+  }
+
   return requestContext.get('trackEvents' as never) === 'true';
```
rita-gorokhod, line 109: question: why not just put userId on every telemetry event as a mandatory param, and call getTrackEventsConfig?

MarceloRGonc: Changed

packages/server/api/test/unit/telemetry.test.ts
```diff
@@ -5,52 +5,71 @@ const axiosMock = {
...
 
 describe('telemetry', () => {
   let getEnvironmentId: jest.Mock;
-  let telemetry: typeof import('../src/lib/telemetry/telemetry').telemetry;
+  let telemetry: typeof import('../../src/app/telemetry/telemetry').telemetry;
 
   beforeEach(() => {
     jest.clearAllMocks();
-    jest.resetModules();
+    // jest.resetModules();
```
bigfluffycookie, line 63: ?

packages/server/api/src/main.ts
```diff
@@ -68,7 +68,10 @@ const main = async (): Promise<void> => {
 
   const app = await setupServer();
 
-  setStopHandlers(app);
+  setStopHandlers(app, async () => {
```
bigfluffycookie, line 71: why this change?

bigfluffycookie (approved):
Looks good just 2 small comments.

packages/engine/src/api-handler.ts
MarceloRGonc, line 1: We don't need telemetry on the engine.

packages/server/api/src/app/flows/flow-run/flow-run-side-effects.ts
MarceloRGonc, line 1: Removed the WorkflowExecution event

packages/server/api/src/app/flows/flow/flow.controller.ts
MarceloRGonc, line 1: An update should always be associated to a user

packages/server/api/src/app/telemetry/event-models/workflow.ts
```diff
@@ -148,29 +145,6 @@ export function sendWorkflowTestBlockEvent(eventParams: {
   });
 }
 
-export function sendWorkflowExecutionEvent(
```
MarceloRGonc, line 151: Not able to fill the userId

packages/server/api/src/app/telemetry/telemetry-event.ts
```diff
@@ -7,7 +7,7 @@ import { UserBase, UserEventName } from './event-models/users';
 
 type EventName = WorkflowEventName | ConnectionEventName | UserEventName;
 
-type EventLabels<T> = Record<string, string> & { userId?: string } & (
+type EventLabels<T> = Record<string, string> & { userId: string } & (
```
MarceloRGonc, line 10: UserId should be mandatory

packages/server/api/src/app/telemetry/telemetry.ts
```diff
@@ -36,34 +40,35 @@ export const telemetry = {
     startMetricsCollector();
   },
   trackEvent(event: TelemetryEvent): void {
-    try {
-      const isEnable = isTelemetryEnabledForCurrentUser();
-      if (!isEnable) {
-        return;
-      }
+    isTelemetryEnabledForCurrentUser(event.labels.userId)
```
MarceloRGonc, line 43: The isTelemetryEnabledForCurrentUser is async. But we can't wait for the metric to be sent

packages/server/api/src/main.ts
MarceloRGonc: The setStopHandlers does not have access to the telemetry. But we need to flush the queue before exit.

packages/server/shared/src/lib/stop-handlers.ts
```diff
@@ -2,10 +2,12 @@
 import { FastifyInstance } from 'fastify';
 import { logger, sendLogs } from './logger/index';
 import { SharedSystemProp, system } from './system';
-import { telemetry } from './telemetry/telemetry';
 
 let shuttingDown = false;
-const stop = async (app: FastifyInstance): Promise<void> => {
+const stop = async (
+  app: FastifyInstance,
+  cleanup?: () => Promise<void>,
```
MarceloRGonc, line 9: Adding a method to be called before exit

packages/server/api/src/app/telemetry/telemetry.ts
```diff
@@ -99,8 +99,14 @@ function enrichEventLabels(event: TelemetryEvent): Timeseries {
...
-  return requestContext.get('trackEvents' as never) === 'true';
+async function isTelemetryEnabledForCurrentUser(
+  userId: string,
+): Promise<boolean> {
+  if (!userId) {
+    return false;
+  }
+
+  return (await userService.getTrackEventsConfig(userId)) === 'true';
 }
```
MarceloRGonc, line 111: Check if the userId as the tracking enable.

PR #269, opened 2025-03-19
Remove array input from list operations action

rita-gorokhod:
Fixes OPS-448

An existing template imports and runs correctly: 
image


packages/blocks/list-operations/src/lib/actions/extract-from-action.ts
```diff
@@ -1,12 +1,13 @@
...
+import { getItemsAsArray } from '../utils';
 
 export const extractFromListAction = createAction({
   auth: BlockAuth.None(),
   name: 'extract_from_list_action',
   description: 'Extract items from a list with a given key',
   displayName: 'Extract From List',
   props: {
-    listItems: Property.Array({
+    listItems: Property.ShortText({
```
bigfluffycookie, line 10: shouldnt it be long text?

packages/blocks/list-operations/src/lib/actions/to-dictionary-action.ts
```diff
@@ -1,12 +1,13 @@
...
+import { getItemsAsArray } from '../utils';
 
 export const toMapAction = createAction({
   auth: BlockAuth.None(),
   name: 'to_map_action',
   description: 'Map items by a given key',
   displayName: 'Map list items',
   props: {
-    listItems: Property.Array({
+    listItems: Property.ShortText({
```
bigfluffycookie, line 10: same here, would except long text

bigfluffycookie (commented):
questions

packages/blocks/list-operations/src/lib/actions/extract-from-action.ts
bigfluffycookie: Hm i think we might aswell support it, i was just thinking bc if its a text field i'd expect to be able to also manually use it

rita-gorokhod: fair point I'll change

rita-gorokhod:
@bigfluffycookie fixed
image

bigfluffycookie (approved):

PR #271, opened 2025-03-20
Add workflow copy event handler

rSnapkoOpenOps:
Fixes OPS-852.

## Additional Notes

This PR adds listener for clipboard copy action. 

If you want to test it add logger to `copy` function

## Testing Checklist

Check all that apply:

image


packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -96,6 +109,21 @@ export const CanvasContextProvider = ({
     setSelectedActions([]);
   }, [selectedActions, state]);
 
+  const copy = useDebounceCallback(() => {
```
cezudas, line 112: Optional (my preference): maybe lines 112-126 + ine 52 can be extracted in a custom hook.

alexandrudanpop, line 112: we will need the copy function for the context menus

alexandrudanpop: since it depends on the `truncatedFlowRef` it might be problematic to extract without extracting the other logic as well

rSnapkoOpenOps: done

rSnapkoOpenOps:
https://www.loom.com/share/4167e45763024150a1438ff881c7ba8b?sid=1c4eed1b-2dfb-40f4-acde-3b837a4bff4e

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -96,14 +120,66 @@ export const CanvasContextProvider = ({
...
+      .then(() => {
+        copyPasteToast({
+          success: true,
+          isCopy: true,
+          itemsCounter: actionCounter,
+        });
+      })
+      .catch(() => {
+        copyPasteToast({
+          success: true,
```
alexandrudanpop, line 160: I think this should be `success: false`

alexandrudanpop (approved):

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
rSnapkoOpenOps: Correct, thanks!

PR #282, opened 2025-03-24
Fix browser back button in workflow builder page

alexandrudanpop:
Fixes OPS-1435

​In React Router, the `setSearchParams` function updates the query string parameters in the URL. This function performs a "PUSH" operation by default, adding a new entry to the browser's history stack.

### DIFF
--- a/packages/react-ui/src/app/routes/flows/id/index.tsx
+++ b/packages/react-ui/src/app/routes/flows/id/index.tsx
@@ -20,7 +20,12 @@ const FlowBuilderPage = () => {
     );
 
     if (viewOnly !== 'true' && viewOnly !== 'false') {
-      setSearchParams({ viewOnly: 'true' });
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/app/routes/flows/id/index.tsx
```diff
@@ -20,7 +20,12 @@ const FlowBuilderPage = () => {
     );
 
     if (viewOnly !== 'true' && viewOnly !== 'false') {
-      setSearchParams({ viewOnly: 'true' });
+      setSearchParams(
+        { viewOnly: 'true' },
+        {
+          replace: true,
+        },
+      );
```
alexandrudanpop, line 28: the main issue was we were setting this in the useEffect

cezudas (approved):
lgtm

PR #283, opened 2025-03-24
Add Google Cloud block boilerplate

MarceloRGonc:
Part of OPS-1427.


packages/blocks/google-cloud/project.json
```diff
@@ -0,0 +1,54 @@
...
+      "generatorOptions": {
+        "packageRoot": "dist/{projectRoot}",
+        "currentVersionResolver": "git-tag"
+      }
+    }
+  },
+  "tags": [],
+  "targets": {
+    "build": {
+      "executor": "nx:noop",
```
MarceloRGonc, line 17: The block will not be available in the UI

packages/blocks/google-cloud/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import { BlockAuth, createBlock } from '@openops/blocks-framework';
+
+export const googleCloud = createBlock({
+  displayName: 'Google Cloud',
+  auth: BlockAuth.None(),
```
MarceloRGonc, line 5: It will be changed later

bigfluffycookie (approved):

packages/blocks/google-cloud/project.json
rita-gorokhod: "nx:noop" means that it will not compile
it needs to be `"executor": "@nx/js:tsc"`
@MarceloRGonc but why is it like this? did you change the create-block template? @bigfluffycookie where did you get this code from?

MarceloRGonc: As this PR only adds the boilerplate, I'm turning off the creation of the block in the dist folder. To have it on the dist folder, we need to change this line to "executor": "@nx/js:tsc",

rita-gorokhod (approved):

packages/blocks/google-cloud/project.json
bigfluffycookie: but i mean does it matter? Aslong as you dont add an action to the index file it wont show the block anyways

PR #284, opened 2025-03-24
Extract common logic in cli block

bigfluffycookie:
Fixes OPS-1436

Since we will be adding a third CLI block, I thought it made sense to extract some common logic, especially the error logging and json parsing.

### DIFF
--- a/packages/blocks/azure/src/lib/azure-cli.ts
+++ b/packages/blocks/azure/src/lib/azure-cli.ts
@@ -37,7 +37,18 @@ export async function runCommand(
 }
 
 async function login(credentials: any, envVars: any) {
-  const loginCommand = `login --service-principal --username ${credentials.clientId} --password ${credentials.clientSecret} --tenant ${credentials.tenantId}`;
+  try {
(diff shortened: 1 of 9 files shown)


MarceloRGonc (approved):

packages/blocks/azure/src/lib/azure-cli.ts
```diff
@@ -37,7 +37,18 @@ export async function runCommand(
...
-  const loginCommand = `login --service-principal --username ${credentials.clientId} --password ${credentials.clientSecret} --tenant ${credentials.tenantId}`;
-
-  return await runCliCommand(loginCommand, 'az', envVars);
+  try {
+    const loginCommand = `login --service-principal --username ${credentials.clientId} --password ${credentials.clientSecret} --tenant ${credentials.tenantId}`;
+
+    return await runCliCommand(loginCommand, 'az', envVars);
+  } catch (error) {
+    let message = 'Error while login into azure: ';
+    if (String(error).includes('login --service-principal')) {
```
rita-gorokhod, line 46: you don't need the IF here,you already know it's the login command

bigfluffycookie: Hmm yea i thought about that, but i think it can still have a different error, like it could fail for a different reason inside run command

PR #288, opened 2025-03-24
Add auth to Google Cloud block

MarceloRGonc:
Part of OPS-1427.


bigfluffycookie (approved):

packages/blocks/framework/src/lib/property/authentication/custom-auth-prop.ts
```diff
@@ -1,5 +1,5 @@
 import { Type } from '@sinclair/typebox';
-import { StaticPropsValue } from '..';
+import { FileProperty, StaticPropsValue } from '..';
```
rita-gorokhod, line 2: is this used? can revert

PR #298, opened 2025-03-25
Add gcloud cli action

bigfluffycookie:
Part of OPS-1427

image


packages/blocks/google-cloud/src/lib/google-cloud-cli.ts
```diff
@@ -0,0 +1,48 @@
...
+      envVars,
+    );
+  }
+
+  return await runCliCommand(command, 'gcloud', envVars);
+}
+
+async function loginGCPWithKeyObject(keyObject: string, envVars: any) {
+  const tmpKeyPath = path.join(os.tmpdir(), 'gcp-key.json');
+  await fs.writeFile(tmpKeyPath, keyObject);
```
bigfluffycookie, line 43: This is necessary because gcloud cli expects a file, not a json object. So until we implement a file input for the connection screen this is a workaround

packages/blocks/google-cloud/src/lib/google-cloud-cli.ts
```diff
@@ -0,0 +1,48 @@
...
+  command: string,
+  auth: any,
+  shouldUseHostCredentials: boolean,
+  project?: string,
+): Promise<string> {
+  const envVars: Record<string, string> = {
+    PATH: process.env['PATH'] || '',
+  };
+
+  const processGoogleCloudConfigDir = process.env['CLOUDSDK_CONFIG'];
```
bigfluffycookie, line 16: I'll need to make a follow up pr to actually add the env variable in all the places similarly to what @MarceloRGonc did with AZURE_CONFIG_DIR

packages/blocks/google-cloud/src/lib/common-properties.ts
```diff
@@ -0,0 +1,55 @@
...
+        options: projects.map(({ name, projectId }) => ({
+          label: name,
+          value: projectId,
+        })),
+      };
+    } catch (error) {
+      return {
+        disabled: true,
+        options: [],
+        placeholder: `Error fetching projects: ${error}`,
```
rita-gorokhod, line 51: Alex added the "error" property, instead of using it in the placeholder 
https://github.com/openops-cloud/openops/blob/main/packages/blocks/azure/src/lib/common-properties.ts#L46

packages/blocks/google-cloud/src/lib/google-cloud-cli.ts
rita-gorokhod: Ew, sad. Thanks.

packages/blocks/google-cloud/src/lib/google-cloud-cli.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  return await runCliCommand(command, 'gcloud', envVars);
+}
+
+async function loginGCPWithKeyObject(keyObject: string, envVars: any) {
+  const tmpKeyPath = path.join(os.tmpdir(), 'gcp-key.json');
+  await fs.writeFile(tmpKeyPath, keyObject);
+
+  const loginCommand = `gcloud auth activate-service-account --key-file=${tmpKeyPath}`;
```
rita-gorokhod, line 45: do we know if it's a service account? what if I want my own credentials, with `gcloud auth login` ?

rita-gorokhod (commented):
Questions

packages/blocks/google-cloud/src/lib/google-cloud-cli.ts
MarceloRGonc: Wrap this with the useTempFile method. We need to delete the file after we use it

packages/blocks/google-cloud/src/lib/common-properties.ts
```diff
@@ -0,0 +1,56 @@
...
+          label: name,
+          value: projectId,
+        })),
+      };
+    } catch (error) {
+      return {
+        disabled: true,
+        options: [],
+        placeholder: `Error fetching projects`,
+        error: `${error}`,
```
rita-gorokhod, line 52: Can just be error: error, or just error

rita-gorokhod (approved):

PR #310, opened 2025-03-26
Add feature flag to clipboard context

rSnapkoOpenOps:
Fixes OPS-1461.

## Testing Checklist

Check all that apply:


cezudas (approved):
Screenshot 2025-03-26 at 22 52 27
we knowingly introduce an eslint warning. but it should be ok for a quick fix

PR #316, opened 2025-03-27
Add jest.env.js for test environment vars

MarceloRGonc:
Fixes OPS-1577.


packages/cli/src/lib/commands/create-block.ts
```diff
@@ -141,11 +143,26 @@ const updateEslintFile = async (blockName: string) => {
   );
   await writePackageEslint(`packages/blocks/${blockName}`, eslintFile);
 };
+
+const updateJestConfigFile = async (blockName: string) => {
+  let jestConfig = await readJestConfig(`packages/blocks/${blockName}`);
+
+  jestConfig = `/* eslint-disable */\n${jestConfig}`;
```
rita-gorokhod, line 150: why are you adding the eslint disable?

packages/cli/src/lib/commands/create-block.ts
```diff
@@ -141,11 +143,26 @@ const updateEslintFile = async (blockName: string) => {
...
+
+  jestConfig = jestConfig.replace(
+    /preset:\s'..\/..\/..\/jest.preset.js',/,
+    (match) => `${match}\n  setupFiles: ['../../../jest.config'],`,
```
rita-gorokhod, line 154: is there a way to pass it to nx generation when we're creating the package?

MarceloRGonc: I didn't find a way. Even other files were already changing them after being generated.

rita-gorokhod (approved):

PR #318, opened 2025-03-27
Add validation for new Google Cloud connections

MarceloRGonc:
Part of OPS-1427

### DIFF
--- a/packages/openops/src/lib/google-cloud/auth.ts
+++ b/packages/openops/src/lib/google-cloud/auth.ts
@@ -34,20 +34,22 @@ export const googleCloudAuth = BlockAuth.CustomAuth({
 });
 
 async function runAuthCommand(keyObject: string): Promise<string> {
+  const gcpConfigDir = await getDefaultCloudSDKConfig();
+
   const envVars: Record<string, string> = {
     PATH: process.env['PATH'] || '',
     CLOUDSDK_CORE_DISABLE_PROMPTS: '1',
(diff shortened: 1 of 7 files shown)


bigfluffycookie (approved):
looks good, just needs the tests fixed

packages/openops/src/lib/google-cloud/auth.ts
```diff
@@ -13,4 +18,40 @@ export const googleCloudAuth = BlockAuth.CustomAuth({
     }),
   },
   required: !enableHostSession,
+  validate: async ({ auth }) => {
+    try {
+      await runAuthCommand(auth.keyFileContent);
```
rita-gorokhod, line 23: so if this doesn't work, it's just going to get stuck forever until it times out?

MarceloRGonc: This is just the login command. If the command takes a long time, yes. But it's the same for properties.

rita-gorokhod: The question is will the user eventually see an error, or is it going to get stuck for a minute+ and not show anything?
Will the user see a descriptive message that there are missing permissions?

MarceloRGonc: The user will see an error on the popup

MarceloRGonc: image

rita-gorokhod (approved):

PR #326, opened 2025-03-28
Add Ternary block integration

StephanQuilyx:
Fixes OPS-1517.

Ternary connection & sample action

New connection using the Ternary API
Fields:
- API key (JWT) generated in Ternary
- Tenant ID (1 tenant per connection)
- API URL (this can vary based on the Ternary environment

## Additional Notes
The supporting API call checks if the JWT (API KEY) is valid before accessing the URL.

## Testing Checklist

Check all that apply:

- [x ] I tested the feature thoroughly, including edge cases

- [x ] I verified all affected areas still work as expected


StephanQuilyx:
Note; please replace the Ternary image with one uploaded to the static section of the OpenOps website

packages/blocks/ternary/src/lib/common/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+}
+
+function validateJwt(token: string): boolean {
+  if (!token) {
+    return false;
+  }
+  try {
+    const decoded = jwtDecode(token);
+    if (decoded && decoded.exp && dayjs().isAfter(dayjs.unix(decoded.exp))) {
```
rita-gorokhod, line 35: Very minor: for readability, would be better to extract this to a var and be explicit in the name. Also note that the IF will evaluate to false if the decoding failed/there is no exp. Perhaps it should be something like:
```
var isValid = decoded && decoded.exp && dayjs().isBefore(dayjs.unix(decoded.exp));

return isValid;
```
(that said, is it a valid use case in ternary tokens that there is no exp?)

packages/blocks/ternary/src/lib/common/index.ts
```diff
@@ -0,0 +1,42 @@
+import {
+  AuthenticationType,
+  // HttpMessageBody,
```
rita-gorokhod, line 3: minor: unused

packages/blocks/ternary/src/index.ts
```diff
@@ -0,0 +1,17 @@
+import { createBlock } from '@openops/blocks-framework';
+// import { ternaryAuth } from '@openops/common';
```
rita-gorokhod, line 2: can remove

packages/blocks/ternary/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+import { getBudgets } from './lib/actions/get-budgets';
+import { ternaryCloudAuth } from './lib/common/auth';
+
+export const ternary = createBlock({
+  displayName: 'Ternary',
+  description: 'FinOps multi-cloud analytics platform.',
+  auth: ternaryCloudAuth,
+  minimumSupportedRelease: '0.20.0',
+  logoUrl:
+    'https://firebasestorage.googleapis.com/v0/b/quilops-712fd.firebasestorage.app/o/public%2Fternary-logo.png?alt=media&token=20820a50-f4ae-4385-942b-bb58cb7cb5cf',
```
rita-gorokhod, line 12: Uploaded to static.openops.com/blocks/ternary.png

tsconfig.base.json
```diff
@@ -17,100 +17,47 @@
     "skipDefaultLibCheck": true,
     "baseUrl": ".",
     "paths": {
```
rita-gorokhod, line 19: what happened here? some auto-linter change?

package.json
```diff
@@ -367,6 +367,7 @@
     "ts-node": "10.9.1",
     "turbowatch": "2.29.4",
     "typescript": "5.5.4",
+    "verdaccio": "^5.0.4",
```
rita-gorokhod, line 370: We never really used it, did you see any improvement? 
The blocks are not published to npm anyway, it's all loaded locally with npm link.

packages/blocks/ternary/src/index.ts
StephanQuilyx: Removed

packages/blocks/ternary/src/index.ts
StephanQuilyx: Updated link

packages/blocks/ternary/src/lib/common/index.ts
StephanQuilyx: Removed

packages/blocks/ternary/src/lib/common/index.ts
StephanQuilyx: I re-used this from 'allow-logged-in-user-only-guard.tsx'. Altered it for better readability. However the function expects a boolean in return so I had to work around that in the return.
Do you have a better solution?

package.json
StephanQuilyx: I never installed this, don't even know what it is/does

StephanQuilyx:
@rita-gorokhod I've updated the files

package.json
rita-gorokhod: Ok I reverted it, not sure where it came from

rita-gorokhod (approved):

PR #327, opened 2025-03-28
Ensure copy action is not visible on trigger step

cezudas:
Fixes OPS-1473.


rSnapkoOpenOps:
If I select a step with a left click and then right click on a trigger I see 'copy' action for a moment 
https://www.loom.com/share/5f97db10b5294e46b7d3e94b3bb9bac4

alexandrudanpop (approved):

PR #329, opened 2025-03-28
Restart engine when blocks change 

MarceloRGonc:
Fixes OPS-1304.

## Testing Checklist

Check all that apply:

- Screenshots
- Loom video
- Preview deployment link


packages/server/api/src/app/app.ts
```diff
@@ -159,13 +159,6 @@ export const setupApp = async (
     transports: ['websocket'],
   });
 
-  app.addHook('onClose', async () => {
-    logger.info('App shutting down');
-    await flowConsumer.close();
-    await systemJobsSchedule.close();
-    await webhookResponseWatcher.shutdown();
-  });
-
```
MarceloRGonc, line 168: Repeated code

maor-rozenfeld:
So.. why doesn't nx do it? Shouldn't nx know that the blocks are a dependency of the engine?

https://nx.dev/nx-api/js/executors/node

MarceloRGonc:
> So.. why doesn't nx do it? Shouldn't nx know that the blocks are a dependency of the engine?
> 
> https://nx.dev/nx-api/js/executors/node

Blocks are implicit dependencies in the engine and in the app. That's why we have to be listening for changes. There is probably an alternative to declaring the implicit dependencies in the project.json file, but we will have to add all the blocks. I haven't tested it. Is this alternative worth looking at?

rita-gorokhod:
> > So.. why doesn't nx do it? Shouldn't nx know that the blocks are a dependency of the engine?
> > https://nx.dev/nx-api/js/executors/node
> 
> Blocks are implicit dependencies in the engine and in the app. That's why we have to be listening for changes. There is probably an alternative to declaring the implicit dependencies in the project.json file, but we will have to add all the blocks. I haven't tested it. Is this alternative worth looking at?

As we discussed offline, "inputs" wildcard might help.
https://nx.dev/reference/inputs

packages/server/shared/src/lib/blocks-builder.ts
MarceloRGonc, line 1: Moved. And now only runs the "nx run-many -t build -p blocks-*" command. I added a lock to not run this command in parallel, between engine and server

maor-rozenfeld (approved):
Great work!

I tried it out and it's super fast :)

PR #330, opened 2025-03-28
Refactor copy paste for better browser support

alexandrudanpop:
Fixes OPS-1492

Support non-https, as the clipboard API is limited to https and localhost.
Just in time permissions request.


packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -103,17 +105,111 @@ export const InteractiveContextProvider = ({
...
+      } else {
+        copySelectedArea();
+      }
+    };
+
+    const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
+    let targetEl: HTMLElement | Document | null = null;
+
+    if (isSafari) {
+      document.addEventListener('copy', copyHandler);
```
alexandrudanpop, line 186: Safari copy handler does not work when attaching to the canavs element. Here we should probably check in the event itseft and not prevent default if copy is from outside canvas

packages/react-ui/src/app/features/builder/hooks/use-paste.ts
```diff
@@ -42,13 +59,21 @@ export const usePaste = () => {
...
+            isCopy: false,
+            itemsCount,
+          });
+        })
+        .catch(() => {
+          copyPasteToast({
+            success: false,
+            isCopy: false,
+            itemsCount,
+          });
```
rSnapkoOpenOps, line 75: `applyOperationAndPushToHistory` is wrapped in `try ... catch`, so catch on this level makes no sense

alexandrudanpop: good catch, no pun intended

packages/ui-components/src/components/flow-canvas/canvas-context.tsx
```diff
@@ -103,17 +104,163 @@ export const InteractiveContextProvider = ({
...
+            });
+          });
+      } else {
+        if (isSafari) {
+          fallbackCopy('-', null, 0);
+          return;
+        }
+
+        fallbackCopy(flowString, action, actionCount);
+      }
```
alexandrudanpop, line 176: for some reason fallback copy the copy doesn't work in safari the first time, so had to add another hack for it

rSnapkoOpenOps (approved):

cezudas (approved):

PR #333, opened 2025-03-30
Bind app service reverse proxy (nginx) to port 8080 and send nginx logs to stdout/stderr

### LINKED ISSUE #361: Support binding to non-privileged ports for compatibility with common Kubernetes SecurityContext settings
To support community members running the app with restrictive Kubernetes SecurityContext configurations (e.g., non-root users), the app service and internal Nginx should support binding to a non-privileged port (e.g., 8080 instead of 80).

Binding to privileged ports like 80 requires elevated permissions, which conflicts with best practices around running containers as non-root.

vicrem:
Fixes #361 .

## Additional Notes

To apply securityContext as non root user we need nginx to listen on ports >1024
Sending nginx logs to console will prevent bloating image with logs and take us one step closer to readonlyfilesystem

Side note, do we need a new layer for creating symbolic links or can we insert the commands somewhere else?

## Testing Checklist

Check all that apply:

- Screenshots
- Loom video
- Preview deployment link


vicrem:
A workaround but nothing I would recommend

# deployment.yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: app
spec:
  selector:
    matchLabels:
      app: app
  template:
    spec:
      securityContext:
        sysctls:
        - name: "net.ipv4.ip_unprivileged_port_start"  #<---
          value: "80"
        runAsNonRoot: true
        runAsGroup: 65534
        runAsUser: 65534
      containers:
        ...

maor-rozenfeld:
Thanks for the contribution!

This is a problematic change, as the gateway nginx is intended for standalone installations where the expected public URL should be 80 or 443 (once we add a TLS guide).

Adding a port as an environment variable would be preferable, but nginx is very challenging when it comes to using environment variables in its configuration files. What do you think?

vicrem:
Happy to help @maor-rozenfeld,

Maybe I'm missing something but I see no problem changing upstream (in this case App service) to listen on port 8080. 
The gateway (in this case Nginx) will respond to inbound traffic on port 80 or 443 and proxy traffic to App service on port 8080. Just like it proxy traffic to Analytics on port 8088, 

If the plan is to remove the gateway and redirect inbound traffic to App service itself then I agree with you, if not, then I see no problem.

Plz clarify 
//Victor

maor-rozenfeld:
> Happy to help @maor-rozenfeld,
> 
> Maybe I'm missing something but I see no problem changing backend (in this case App service) to listen on port 8080. The gateway (in this case Nginx) will respond to inbound traffic on port 80 or 443 and proxy traffic to App service on port 8080. Just like it proxy traffic to Analytics on port 8088,
> 
> If the plan is to remove the gateway and redirect inbound traffic to App service itself then I agree with you, if not, then I see no problem.
> 
> Plz clarify //Victor

Hi, sorry for my previous sleep deprived reply, I thought you were changing the gateway nginx. Yes, this PR makes more sense now :)

vicrem:
even the sun has its spots :)

Dockerfile
```diff
@@ -36,6 +36,9 @@ LABEL service=openops
 
 # Copy Nginx configuration template and static files
 COPY nginx.standard.conf /etc/nginx/nginx.conf
+# Send logs to console instead of writing to fileystem.
+RUN ln -sf /dev/stdout /var/log/nginx/access.log \
+  && ln -sf /dev/stderr /var/log/nginx/error.log
```
maor-rozenfeld, line 41:
```suggestion
# Send nginx logs to console instead of writing to fileystem.
RUN ln -sf /dev/stdout /var/log/nginx/access.log \
  && ln -sf /dev/stderr /var/log/nginx/error.log
```

I don't think the extra layer is a big deal here.

maor-rozenfeld:
This PR is mostly/only relevant in the scope of k8s, right? Could you please describe it a little more in details so I can create an issue to link to?

Thanks you.

vicrem:
Yes more or less.. 

SecurityContext in k8s defines security settings (like user, group IDs, capabilities etc.) for Pods/Containers to enforce least privilege. To allow non-root users/processes to bind to privileged ports (<1024), weakening security.

Configuration for App service:
Pod level:
    securityContext:
      runAsUser: 65534
      runAsGroup: 65534
      runAsNonRoot: true

Container level:
    securityContext:
      capabilities:
        drop:
        - all
      readOnlyRootFilesystem: true
      allowPrivilegeEscalation: false


//Victor

PR #336, opened 2025-03-31
Remove turbowatch from the dependencies

MarceloRGonc:
Part of OPS-1304.


bigfluffycookie (approved):
Looks good but please leave a comment on why this change was needed.

MarceloRGonc:
> Looks good but please leave a comment on why this change was needed.

Unused dependency

PR #340, opened 2025-03-31
Add support for date comparisons when inserting in tables

MarceloRGonc:
Fixes OPS-1377.

## Additional Notes

The date must be in the format 'YYYY-MM-DD' and can include the time if the table is set to do so.

image

In case of invalid format:
image

### DIFF
--- a/packages/openops/src/lib/openops-tables/rows.ts
+++ b/packages/openops/src/lib/openops-tables/rows.ts
@@ -153,7 +153,7 @@ export async function getRowByPrimaryKeyValue(
       {
         fieldName: primaryKeyFieldName,
         value: primaryKeyFieldValue,
-        type: getFilterTypesEnum(primaryKeyFieldType),
+        type: getEqualityFilterType(primaryKeyFieldType),
       },
     ],
   });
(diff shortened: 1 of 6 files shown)


packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -163,3 +164,11 @@ export async function getRowByPrimaryKeyValue(
 
   return rows[0];
 }
+
+function getFilterTypesEnum(primaryKeyFieldType: string): ViewFilterTypesEnum {
+  if (primaryKeyFieldType === 'date') {
+    return ViewFilterTypesEnum.date_equal;
+  }
```
MarceloRGonc, line 171: Only the date type is using a different equal.
https://github.com/openops-cloud/tables/blob/2f0418ddc210ba89385d2a73861794e2e04185fc/backend/src/baserow/config/settings/base.py#L556

packages/openops/test/openops-tables/rows.test.ts
```diff
@@ -296,54 +296,72 @@ describe('getRowByPrimaryKeyValue', () => {
...
-    ]);
-    createAxiosHeadersMock.mockReturnValue('some header');
-    await expect(
-      getRowByPrimaryKeyValue(
+  test.each([
+    ['text', 'equal'],
+    ['rating', 'equal'],
+    ['email', 'equal'],
+    ['date', 'date_equal'],
+  ])(
```
MarceloRGonc, line 304: Added some test cases. And the validation. The rest is indentation

bigfluffycookie (approved):

PR #342, opened 2025-04-01
Add protocol and hostname check to disable problematic "paste" context menu actions in insecure contexts

cezudas:
Fixes OPS-1491.

https://www.loom.com/share/e6488453b1a6409495c627b3533ff3f0

### DIFF
--- a/packages/react-ui/src/app/features/builder/flow-canvas/context-menu/utils.ts
+++ b/packages/react-ui/src/app/features/builder/flow-canvas/context-menu/utils.ts
@@ -1,1 +1,6 @@
-export const isProtocolHttps = window.location.protocol === 'https:';
+/**
+ * Whether the current page is served over HTTPS or if the hostname is "localhost".
+ */
+export const hasSecureClipboardAccess =
+  window.location.protocol === 'https:' ||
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/app/features/builder/flow-canvas/context-menu/utils.ts
```diff
@@ -0,0 +1 @@
+export const isProtocolHttp = window.location.protocol === 'http:';
```
cezudas, line 1: In our case, it's enough to evaluate this once when the module is imported, so it's okay to define it "constant like."

packages/react-ui/src/app/features/builder/flow-canvas/context-menu/utils.ts
```diff
@@ -0,0 +1 @@
+export const isProtocolHttps = window.location.protocol === 'https:';
```
alexandrudanpop, line 1: Paste should be enabled in localhost where it's a secure context:
https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#when_is_a_context_considered_secure

PR #346, opened 2025-04-01
Add GCP sample template to home page

rSnapkoOpenOps:
Fixes OPS-1428.

## Testing Checklist

Check all that apply:

image

### DIFF
--- a/packages/react-ui/src/app/features/home/components/home-get-started.tsx
+++ b/packages/react-ui/src/app/features/home/components/home-get-started.tsx
@@ -7,6 +7,7 @@ import {
 } from '@openops/components/ui';
 import { t } from 'i18next';
 import { BookOpenText } from 'lucide-react';
+import { useMemo } from 'react';
 
 type HomeGetStartedProps = {
   sampleTemplates: FlowTemplateMetadataWithIntegrations[];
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/app/features/home/components/home-get-started.tsx
```diff
@@ -30,27 +30,17 @@ const HomeGetStarted = ({
...
-                </>
-              )}
+            <div className="flex gap-2">
+              {sampleTemplates.map((template, index) => (
+                <SampleTemplateCard
+                  key={template.id}
+                  templateMetadata={template}
+                  onClick={() => onSampleTemplateClick(template)}
+                />
+              ))}
```
alexandrudanpop, line 43: will this work and render correctly if we have more than 3 templates? otherwise we should just take first 3 and render them, and probably log and error to console in case we have more than 3, just to not break the homepage in case more than 3 are setup by mistake in the DB.

rSnapkoOpenOps: Good catch! I added slicing to the array. But I don't think we need to log an error

alexandrudanpop (approved):

PR #350, opened 2025-04-01
Improve error message in the 'Get Price from Price catalog' action

MarceloRGonc:
Fixes OPS-1498

### DIFF
--- a/packages/server/api/src/app/aws/pricing-module.ts
+++ b/packages/server/api/src/app/aws/pricing-module.ts
@@ -5,6 +5,7 @@ import {
 } from '@fastify/type-provider-typebox';
 import { SupportedPricingRegion } from '@openops/common';
 import { FastifyRequest } from 'fastify';
+import { StatusCodes } from 'http-status-codes';
 import { GetPricingRequest } from './pricing-request';
 import { getPrice } from './pricing-service';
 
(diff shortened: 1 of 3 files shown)


packages/server/api/src/app/aws/pricing-module.ts
```diff
@@ -20,12 +21,24 @@ const PricingController: FastifyPluginCallbackTypebox = (
...
+    async (
+      request: FastifyRequest<{ Querystring: GetPricingRequest }>,
+      reply,
+    ) => {
+      try {
+        const result = await getPrice(
+          request.query.serviceCode,
+          JSON.parse(request.query.filters) as Filter[],
+          request.query.region as unknown as SupportedPricingRegion,
+        );
```
MarceloRGonc, line 33: We need to wait to catch the error.

packages/server/api/src/app/aws/pricing-service.ts
```diff
@@ -39,3 +40,21 @@ function createCacheKey(serviceCode: string, filters: Filter[]): string {
 
   return cacheKey;
 }
+
+function getCredentialValue(prop: AppSystemProp): string {
+  const value = system.getOrThrow(prop);
+
```
MarceloRGonc, line 46: If the variable is set as empty, we don't throw an error.

bigfluffycookie (approved):
Looks good but please add tests for the changes in pricing-service

PR #372, opened 2025-04-03
Return the exit code of the node process to exit the container 

MarceloRGonc:
Part of OPS-1516.

## Additional Notes
When the container crashes due to lack of memory, an error occurs but the container does not restart.

### DIFF
--- a/deploy/docker-compose/docker-compose.yml
+++ b/deploy/docker-compose/docker-compose.yml
@@ -33,7 +33,7 @@ services:
     image: public.ecr.aws/openops/openops-engine:${OPS_VERSION:-latest}
     restart: unless-stopped
     env_file: .env
-    command: ["cp -r /var/tmp-base/. /tmp/ && node main.js"]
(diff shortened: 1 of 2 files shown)


maor-rozenfeld:
This change seems redundant, I doubt it will change anything. The command exit code should already be returned.

I'm guessing the issue is an out-of-memory exception that kills the app but doesn't kill the node process.

MarceloRGonc:
> This change seems redundant, I doubt it will change anything. The command exit code should already be returned.
> 
> I'm guessing the issue is an out-of-memory exception that kills the app but doesn't kill the node process.

The node process runs out of memory and stops, but the container keeps running.

maor-rozenfeld:
> The node process runs out of memory and stops, but the container keeps running.

Did you manage to reproduce? This PR fixes the issue?

MarceloRGonc:
> > The node process runs out of memory and stops, but the container keeps running.
> 
> Did you manage to reproduce? This PR fixes the issue.

Yes. I tested this by simulating a release, and the container after this PR restarts.

PR #376, opened 2025-04-04
Add Snowflake block

cezudas:
Part of OPS-1482.
Demo
https://www.loom.com/share/6a51303d6da24014b18137729eaea18a?sid=2a33c615-5bf8-4390-aa36-6d4c955e93fa

P.S. The Demo contains actions that were removed last minute from the PR to keep the PR simple -> run-query and run-multiple-queries will be added in follow up PRs


package-lock.json
bigfluffycookie, line 1: what are these huge changes?

THIRD_PARTY_LICENSES.txt
bigfluffycookie, line 1: Why are there so many changes in here? (besides the snowflake license part)

packages/blocks/snowflake/test/insert-row.test.ts
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
bigfluffycookie, line 1: Thanks for adding a test!

package-lock.json
cezudas: this is just `npm i` output

THIRD_PARTY_LICENSES.txt
cezudas: package lock has many changes, that's the reason

bigfluffycookie (approved):

PR #392, opened 2025-04-07
Fix committing third party license changes for dependabot

maor-rozenfeld:
The automatic commit of third party licenses by dependabot stopped working since we moved to "pull request" trigger. This PR fixes that.

Part of OPS-195.


rita-gorokhod:
As we discussed, closing, or else it can be missed when reviewing

MarceloRGonc (approved):

PR #395, opened 2025-04-07
Add get recommendations action to google cloud block

bigfluffycookie:
Fixes OPS-1483

image

image

image

image


engine.Dockerfile
```diff
@@ -39,6 +39,8 @@ RUN <<-```
     curl -sSL https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-516.0.0-linux-x86_64.tar.gz -o /tmp/gcloud.tar.gz
     mkdir -p /opt && tar -C /opt -xf /tmp/gcloud.tar.gz
     /opt/google-cloud-sdk/install.sh --quiet
+    /opt/google-cloud-sdk/bin/gcloud components install beta --quiet
```
rita-gorokhod, line 42: why did you need this? vs the previous `/opt/google-cloud-sdk/install.sh` ?

bigfluffycookie: The recommenders list cli call needs to be done with goole cloud beta, and you have to install it as a component

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+      refreshers: ['filterBySelection'],
+      props: async ({ filterBySelection }) => {
+        if (!filterBySelection) {
+          return {} as any;
+        }
+        return filterBySelection;
+      },
+    }),
+    recommenders: getRecommendersDropdown(),
+    location: Property.ShortText({
```
bigfluffycookie, line 61: There is an API call to get this list i think but as we are only authenticated with the cli i dont think we can get it here

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+      label: 'Filter by Billing Account',
+      value: { billingAccount: dropdowns['billingAccount'] },
+    },
+    {
+      label: 'Filter by Organization ID',
+      value: { organization: dropdowns['organization'] },
+    },
+    { label: 'Filter by Project ID', value: { project: dropdowns['project'] } },
+    {
+      label: 'Filter by Folder ID',
```
bigfluffycookie, line 202: This has to be a text as the input for the cli call needs to contain either Org id, or folder id, so you can either get all folder within an orgs are below another folder

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
+import { createAction, Property } from '@openops/blocks-framework';
```
bigfluffycookie, line 1: Okay so the reason this looks do confusing is:

For the get recommendations call the user must provide either 

Project id, organization id, billing account id or folder id.
 
For each of those ( except folder id ) we do a call to google cloud to get the corresponding list. 

Once the user has selected that we need to get a list of all the recommenders, which also requires one of the above choices as input parameter.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+
+      let baseCommand = `gcloud recommender recommendations list --format=json ${getFilterByPropertyCommandParam(
+        filterByProperty,
+      )}`;
+
+      if (location) {
+        baseCommand += ` --location=${location}`;
+      }
+
+      const commands = recommenders.map(
```
bigfluffycookie, line 84: Small  problem, some of the recommenders throw an invalid argument error. I dont know why this is yet, I am only listing the recommenders that are valid for the project/org/billingAccount/folder so in theory it should work. It could be that it has trouble when the api isnt enabled? Or when there are no resources for the recommendations? anyways trying to figure it out

packages/blocks/google-cloud/test/index.test.ts
```diff
@@ -0,0 +1,17 @@
+import { googleCloud } from '../src/index';
```
bigfluffycookie, line 1: realized it was missing

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,304 @@
...
+        return {
+          disabled: true,
+          options: [],
+          placeholder: 'Please authenticate to see recommenders.',
+          error: undefined,
+        };
+      }
+      
+      let filterFlag = '';
+      if (!filterByProperty['billingAccount']) {
```
bigfluffycookie: So, recommenders list does not support billing account as a param, it does allow folder, org and project though

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: okay below is  a list of all the ones that are not working any why  -> 
```
google.accounts.security.SecurityKeyRecommender - https://cloud.google.com/recommender/docs/account-security-recommender - only available in console, not cli
google.bigquery.jobs.ErrorMitigationRecommender- invalid arg
google.cloudbilling.commitment.SpendBasedCommitmentRecommender - invalid arg
google.cloudplatform.productledgrowth.Recommender - invalid arg
google.composer.environment.Recommender - returns 500, internal error
google.di.productx.Recommender - invalid arg
```

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
MarceloRGonc: I think you can use the CLI to get a token and then make the API call, but I don't know if it's worth it.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,294 @@
...
+      handleCliError({
+        provider: 'Google Cloud',
+        command: currentCommand,
+        error,
+      });
+    }
+  },
+});
+
+async function getScopeOptions(auth: any, useHostSession: boolean) {
```
MarceloRGonc, line 105: Suggestion. I think we can still do better. If you want to try.

```
  const scopeConfigs = {
    billingAccount: {
      command: 'gcloud billing accounts list',
      extractLabelAndValue: (item: { name: string; displayName: string }) => ({
        label: item.displayName,
        value: item.name.split('/')[1],
      }),
      dropdown: createDropdown(
        'Billing Account',
        'The Google Cloud Platform billing account ID to use for this invocation.',
        true,
      ),
    },
    organization: {
      command: 'gcloud organizations list',
      extractLabelAndValue: (item: { name: string; displayName: string }) => ({
        label: item.displayName,
        value: item.name.split('/')[1],
      }),
      dropdown: createDropdown(
        'Organization ID',
        'The Google Cloud Platform organization ID to use for this invocation.',
        true,
      ),
    },
    project: {
      command: 'gcloud projects list',
      extractLabelAndValue: (item: { name: string; projectId: string }) => ({
        label: item.name,
        value: item.projectId,
      }),
      dropdown: createDropdown(
        'Project ID',
        'The Google Cloud Platform project ID.',
        true,
      ),
    },
  };
  ```

File: packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts (hunk 5 above)
Comment (MarceloRGonc) [reply packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts:84]: Those that return invalid argument? Do you know if there is something we are doing wrong in the request.

File: packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
Comment (MarceloRGonc) [inline packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts:1]: Maybe separate this file, at least the properties from the run block

File: packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts (hunk 2 above)
Comment (bigfluffycookie) [reply packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts:61]: I tried that, but it said its the wrong type of auth token, it needs to oauth2 or the project token

File: packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts (hunk 5 above)
Comment (bigfluffycookie) [reply packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts:84]: hm no thats the problem, im not sure what the issue is with those, I suspect it could have something to do with the fact that none of them are present in the official list here -> https://cloud.google.com/recommender/docs/recommenders

but I honestly do not know why it would return them then.

File: engine.Dockerfile [hunk 9]
```diff
@@ -39,6 +39,8 @@ RUN <<-```
...
+    /opt/google-cloud-sdk/bin/gcloud beta --help
```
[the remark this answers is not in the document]
bigfluffycookie: hm, mostly so that it will fail properly if the component wasnt installed, later on in the action it is very unclear what the problem is if this fails silently.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: yea i thought of it originally, but im not sure it provides much value. My concern is that in the future we will add actions that use the API, which will almost for sure have a dropdown for the scopes like here, so if we put this into a seperate file we will have the dropdown twice once for cli and once for api, I think it would be confusing

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,296 @@
...
+  googleCloudAuth,
+  handleCliError,
+  tryParseJson,
+} from '@openops/common';
+import { runCommand, runCommands } from '../google-cloud-cli';
+
+export const getRecommendationsAction = createAction({
+  auth: googleCloudAuth,
+  name: 'google_get_recommendations_cli',
+  description: 'Fetch the recommendations for the provided recommenders',
```
rita-gorokhod, line 13: nitpicky:
maybe "for the selected"
provided sounds like it's input from another step

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: @rita-gorokhod wdyt? Should i like manually remove those 5 from the list or?

engine.Dockerfile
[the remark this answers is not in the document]
MarceloRGonc: If the installation fails, doesn't the previous command fail? This one won't run

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+
+async function getScopeOptionProperty(
+  auth: any,
+  useHostSession: any,
+  filterSelection: any,
+) {
+  if (filterSelection === 'folder') {
+    return {
+      folder: Property.ShortText({
+        displayName: 'Folder ID',
```
rita-gorokhod, line 155: how will the user know this?
is there a way to fetch all folders or something? can we do by folder name?

bigfluffycookie: uegh my comment got lost, there is a way to get folders, but it requires the user to either provide orgId or folderId, 

I talked with marco about this and he agreed it makes more sense to just add it as a text, considering we have no way of having the user input the parameters for the request. 

They can find the folder ids in the console

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+
+        return await getScopeOptionProperty(
+          auth,
+          shouldUseHostCredentials,
+          filterBySelection,
+        );
+      },
+    }),
+    recommenders: getRecommendersDropdown(),
+    location: Property.ShortText({
```
rita-gorokhod, line 93: @bigfluffycookie is this like regions? is there a static list?

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+          error: `${error}`,
+        };
+      }
+    },
+  });
+}
+
+function getFilterByPropertyCommandParam(filterByProperty: any) {
+  if (filterByProperty?.['billingAccount']) {
+    return `--billing-account=${filterByProperty['billingAccount']}`;
```
MarceloRGonc, line 288: Can't this be the dropdown value?

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
```diff
@@ -0,0 +1,298 @@
...
+      }),
+    };
+  }
+
+  const scopeConfigs = {
+    billingAccount: {
+      command: 'gcloud billing accounts list',
+      extractLabelAndValue: (item: { name: string; displayName: string }) => ({
+        label: item.displayName,
+        value: item.name.split('/')[1],
```
MarceloRGonc, line 168: If the value is the filtering flag we do not need to validate it afterwards. --billing-account=""

packages/blocks/google-cloud/test/get-recommendations-cli-action.test.ts
```diff
@@ -0,0 +1,209 @@
...
+    expect(result).toMatchObject({
+      markdown: {
+        type: 'MARKDOWN',
+        description:
+          'Please select a filter from the dropdown to see the options.',
+      },
+    });
+  });
+
+  test('should return dynamic dropdown property for billingAccount', async () => {
```
rita-gorokhod, line 99: what about the others?
can maybe do test cases here to not repeat the test code

packages/blocks/google-cloud/test/get-recommendations-cli-action.test.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  test('should build correct command and parse responses for multiple recommenders', async () => {
+    const context = createContext({
+      useHostSession: { useHostSessionCheckbox: true },
+      filterByProperty: { organization: 'org-789' },
+      location: 'us-east1',
+      recommenders: ['recommender-1', 'recommender-2'],
+    });
+
+    gcloudCliMock.runCommands.mockResolvedValueOnce([
```
rita-gorokhod, line 160: try also a recommender that returns nothing, and a recommender that returns 2 instead if 1. To make sure that the flattening works properly.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: a

packages/blocks/google-cloud/test/get-recommendations-cli-action.test.ts
bigfluffycookie: thanks for spotting, i had seperate test cases but i guess i lost them during the rewrite ;(

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: wdym?

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: I dont understand, how are we validating it?

engine.Dockerfile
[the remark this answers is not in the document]
bigfluffycookie: Not necessarily, I have had it many times with cli installations that it didnt install properly but it continued on anyways, I just wanted to be safe but ig i can remove it if you think its better.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
rita-gorokhod: yes @bigfluffycookie let's remove it

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
MarceloRGonc: In your code, the dropdown value is a string that you will later append to the filter. Example:
- billingAccount: xpto
--billing-account=xpto

- project: xpto
--project=xpto

If the dropdown returns the filter this method is not needed.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
MarceloRGonc: Other comment.

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
bigfluffycookie: yea the method is still needed because i still need to access the property by name, eg -> filterByProperty?.['billingAccount'] has to stay regardless no? I guess it would make it a tiny bit better because i could to 

if (filterByProperty?.['billingAccount']) {
    return filterByProperty['billingAccount'];
}

but im not sure its super worth the change (unless as i said my initial assumption is wrong)

engine.Dockerfile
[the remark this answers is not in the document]
MarceloRGonc: In my opinion this doesn't make sense. I understand what you want to do, but we don't do it for other commands.

rita-gorokhod (approved):

packages/blocks/google-cloud/src/lib/actions/get-recommendations-cli-action.ts
MarceloRGonc: The key could be the same for all dropdowns since only one will appear.

PR #406, opened 2025-04-08
Add explanation to "Ignored" run status

rmn-snpk:
Fixes OPS-1533.

## Testing Checklist

Check all that apply:

image


packages/react-ui/src/app/features/flow-runs/lib/flow-run-utils.ts
```diff
@@ -115,6 +116,11 @@ export const flowRunUtils = {
         };
     }
   },
+  getStatusExplanation(status: FlowRunStatus): string | undefined {
+    if (status === FlowRunStatus.IGNORED) {
+      return t('Previous scheduled run is still in progress');
+    }
+  },
```
rSnapkoOpenOps, line 123: We can update this function and use switch case in future

PR #408, opened 2025-04-08
Install the clis in the codespaces container

bigfluffycookie:
Fixes OPS-1540


.devcontainer/Dockerfile
```diff
@@ -31,3 +31,22 @@ RUN wget https://github.com/minamijoyo/hcledit/releases/download/v0.2.15/hcledit
   tar -C /usr/bin -xf /tmp/hcledit_0.2.15_linux_amd64.tar.gz && \
   chmod +x /usr/bin/hcledit && \
   rm /tmp/hcledit_0.2.15_linux_amd64.tar.gz
+
+# Install AWS CLI
```
bigfluffycookie, line 35: its an easy fix and I think its completely worth it, its a pain having to install them for every new codespace.

MarceloRGonc (approved):

PR #409, opened 2025-04-08
Align Store block Scope options

rita-gorokhod:
Fixes OPS-1541

### DIFF
--- a/packages/blocks/store/src/lib/actions/common.ts
+++ b/packages/blocks/store/src/lib/actions/common.ts
@@ -37,19 +37,19 @@ export const common = {
     options: {
       options: [
         {
-          label: 'Project',
-          value: BlockStoreScope.PROJECT,
+          label: 'Within a single run of this workflow',
+          value: BlockStoreScope.RUN,
         },
         {
-          label: 'Flow',
+          label: 'Between all runs of this workflow',
           value: BlockStoreScope.FLOW,
         },
         {
-          label: 'Run',
-          value: BlockStoreScope.RUN,
+          label: 'Between all runs of all workflows',
+          value: BlockStoreScope.PROJECT,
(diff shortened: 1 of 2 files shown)


packages/blocks/store/src/lib/actions/common.ts
```diff
@@ -37,19 +37,19 @@ export const common = {
...
         {
-          label: 'Run',
-          value: BlockStoreScope.RUN,
+          label: 'Between all runs of all workflows',
+          value: BlockStoreScope.PROJECT,
         },
       ],
     },
-    defaultValue: BlockStoreScope.PROJECT,
+    defaultValue: BlockStoreScope.RUN,
```
rita-gorokhod, line 53: I don't think the default should've been for the whole workspace

packages/blocks/store/src/lib/actions/common.ts
```diff
@@ -37,19 +37,19 @@ export const common = {
     options: {
       options: [
         {
```
rita-gorokhod, line 39: Changed the order to match the default, it should go from most restrictive to less

packages/blocks/store/src/lib/actions/store-get-action.ts
```diff
@@ -28,28 +23,7 @@ export const storageGetAction = createAction({
...
-          },
-          {
-            label: 'Run',
-            value: BlockStoreScope.RUN,
-          },
-        ],
-      },
-      defaultValue: StoreScope.PROJECT,
-    }),
+    store_scope: common.store_scope,
```
rita-gorokhod, line 26: All other actions were already referencing this, only here was the copy-paste

bigfluffycookie (approved):

PR #419, opened 2025-04-08
Support changing the admin credentials after app initialization

maor-rozenfeld:
Today it's impossible to change the admin credentials after the initial setup, since the seed sequence only creates the admin user but doesn't update its password or email.

This PR allow users to change the admin email and password in their `.env` file and the changes will be applied when they start the application. The code checks if the admin user exist, and if it does, update its email and password but in the app and in tables (baserow).

The analytics container was also updated to behave the same way, with its default admin and poweruser passwords.

Fixes OPS-1387.


deploy/docker-compose/docker-compose.yml
```diff
@@ -44,14 +44,16 @@ services:
       - ${HOST_AZURE_CONFIG_DIR:-openops_azure_cli_data}:/tmp/azure
       - ${HOST_CLOUDSDK_CONFIG:-openops_gcloud_cli_data}:/tmp/gcloud
   openops-tables:
-    image: public.ecr.aws/openops/openops-tables:0.1.8
+    image: 586998899304.dkr.ecr.us-east-2.amazonaws.com/openops-tables:support-admin-env-credentials-latest
```
rita-gorokhod, line 47: I imagine this will change before merge?

packages/server/api/src/app/authentication/authentication-service/hooks/community-authentication-hooks.ts
```diff
@@ -31,15 +31,8 @@ export const communityAuthenticationServiceHooks: AuthenticationServiceHooks = {
     let organization = await organizationService.getOldestOrganization();
 
     const isAdminUser = user.email === adminEmail;
-    if (isAdminUser) {
-      if (organization) {
-        throw Error(
```
rita-gorokhod, line 36: why did you remove this check?

maor-rozenfeld: I will revert. Originally I allowed creating another admin user, but now the code always updates the existing one.

packages/server/api/src/app/authentication/authentication-service/index.ts
```diff
@@ -24,12 +29,18 @@ export const authenticationService = {
...
-      email: params.email,
-      password: params.password,
-      authenticate: true,
-    });
+    // The default superuser should already exist, since the tables container
+    // creates it on startup, and update its password if it was changed.
+    const isSuperuser =
+      params.email == system.getOrThrow(AppSystemProp.OPENOPS_ADMIN_EMAIL);
+    const { token, refresh_token } = isSuperuser
+      ? await authenticateUserInOpenOpsTables(params.email, params.password)
```
rita-gorokhod, line 37: I don't understand this logic.

This func will be called in 2 instances:
1. Either it's brand new installation and it's the first time we're creating a seed user at all.
2. We're registering with a non-admin user

For 1, why would we need authenticateUserInOpenOpsTables?

maor-rozenfeld: The admin user is unique, as its parallel in the tables container is created on the tables container boot, and not during the app user sign up.

maor-rozenfeld: The signUp includes a postSignUp hook that creates an organization and a tables workspace, and adds the admin to them. The original code used the signUp function - didn't see a reason to change that (will require refactoring of the postSignUp function).

MarceloRGonc (approved):

PR #420, opened 2025-04-09
Redirect users to correct page after login

alexandrudanpop:
Fixes OPC-78

## Additional Notes

- when a user just clicks log-out, he is always redirected to homepage
- when a user is logged out because his session expired, he is redirected after login to the page he was trying to visit

## Testing Checklist

Check all that apply:

Deployed to UX


packages/react-ui/src/app/lib/authentication-session.ts
```diff
@@ -73,17 +75,30 @@ export const authenticationSession = {
...
-      )
+      window.location.pathname === '/sign-in' ||
+      window.location.pathname === '/sign-up'
     ) {
-      window.location.reload();
+      return;
+    }
+
+    if (navigate) {
+      navigate('/sign-in');
```
alexandrudanpop, line 101: we don't need a hard reload when we log-out

rSnapkoOpenOps (approved):
lgtm

cezudas (approved):

PR #429, opened 2025-04-10
Retry failing 409 requests to tables service during app seeding phase

cezudas:
Fixes OPS-1311.


packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -101,6 +101,7 @@ export async function getRows(getRowsParams: GetRowsParams) {
     const getRowsResult = await makeOpenOpsTablesGet<{ results: any[] }[]>(
       url,
       authenticationHeader,
+      false,
```
cezudas, line 104: I would say it's unlikely this module (used by`openops-tables` block) will be extended/changed more often than the openops-tables seed logic in packages/server/api.

packages/openops/src/lib/openops-tables/requests-helpers.ts
```diff
@@ -1,42 +1,94 @@
...
   return new AxiosHeaders({
     'Content-Type': 'application/json',
     Authorization: `JWT ${token}`,
   });
 }
 
+const getAxiosInstance = (useAxiosRetry: boolean): AxiosInstance => {
+  return useAxiosRetry ? retryAxiosInstance : axios;
+};
+
```
cezudas, line 38: It would be convenient always to use axiosRetry with all the "tables" requests.
But it might be a bad idea to use axiosRetry when executing Tables actions.
@rita-gorokhod 
WDYT?

packages/openops/src/lib/openops-tables/requests-helpers.ts
```diff
@@ -1,42 +1,94 @@
...
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export async function makeOpenOpsTablesPost<T>(
   route: string,
   body: any,
   headers?: AxiosHeaders,
+  useAxiosRetry = true,
```
cezudas, line 44: Not ideal, yet it avoids a more significant refactor = big PR. 
It makes this assumption:
https://github.com/openops-cloud/openops/pull/429/files#diff-a9edf8d5509b4c4d58a0aeed27230fa9278754f16e14c77fe3dc14f864cb1d91R104

packages/openops/src/lib/openops-tables/rows.ts
cezudas, line 104: same remark as here: https://github.com/openops-cloud/openops/pull/429/files#diff-a9edf8d5509b4c4d58a0aeed27230fa9278754f16e14c77fe3dc14f864cb1d91R104

packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -114,6 +115,7 @@ export async function updateRow(updateRowParams: UpdateRowParams) {
       `api/database/rows/table/${updateRowParams.tableId}/${updateRowParams.rowId}/?user_field_names=true`,
       updateRowParams.fields,
       authenticationHeader,
+      false,
```
cezudas, line 118: same remark as here https://github.com/openops-cloud/openops/pull/429/files#diff-a9edf8d5509b4c4d58a0aeed27230fa9278754f16e14c77fe3dc14f864cb1d91R104

packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -125,6 +127,7 @@ export async function addRow(addRowParams: AddRowParams) {
       `api/database/rows/table/${addRowParams.tableId}/?user_field_names=true`,
       addRowParams.fields,
       authenticationHeader,
+      false,
```
cezudas, line 130: same remark as here https://github.com/openops-cloud/openops/pull/429/files#diff-a9edf8d5509b4c4d58a0aeed27230fa9278754f16e14c77fe3dc14f864cb1d91R104

packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -135,6 +138,7 @@ export async function deleteRow(deleteRowParams: DeleteRowParams) {
     return await makeOpenOpsTablesDelete(
       `api/database/rows/table/${deleteRowParams.tableId}/${deleteRowParams.rowId}/`,
       authenticationHeader,
+      false,
```
cezudas, line 141: same remark as here https://github.com/openops-cloud/openops/pull/429/files#diff-a9edf8d5509b4c4d58a0aeed27230fa9278754f16e14c77fe3dc14f864cb1d91R104

packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -7,10 +7,10 @@ import {
...
   createAxiosHeaders,
-  makeOpenOpsTablesDelete,
-  makeOpenOpsTablesGet,
-  makeOpenOpsTablesPatch,
-  makeOpenOpsTablesPost,
+  makeOpenOpsTablesDeleteWithoutRetry,
+  makeOpenOpsTablesGetWithoutRetry,
+  makeOpenOpsTablesPatchWithoutRetry,
+  makeOpenOpsTablesPostWithoutRetry,
 } from '../openops-tables/requests-helpers';
```
cezudas, line 14: it assumes the actions of table block should not use axios retry -> use retry only for the all tables requests fired during app seeding

packages/openops/src/lib/axios-wrapper.ts
```diff
@@ -15,8 +22,13 @@ export async function makeHttpRequest<T>(
       url,
     };
 
-    const response = await axios(config);
-
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let response: AxiosResponse<any, any>;
```
MarceloRGonc, line 26: let response: AxiosResponse<T, AxiosRequestConfig>;

packages/openops/src/lib/openops-tables/requests-helpers.ts
```diff
@@ -1,42 +1,94 @@
...
+  retries: 3,
+  retryDelay: (retryCount: number) => {
+    return retryCount * 10000;
+  },
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  retryCondition: (error: any) => {
+    return (
+      error?.response?.status &&
+      error?.response?.status === StatusCodes.CONFLICT
+    );
```
MarceloRGonc, line 24: This retry condition must be configurable. In the seed we are using this condition but we may want to use another one in other requests.

packages/openops/src/lib/openops-tables/requests-helpers.ts
```diff
@@ -1,28 +1,93 @@
...
 ): Promise<T> {
-  return makeOpenOpsTablesRequest<T>('POST', route, body, headers);
+  return makeOpenOpsTablesRequest<T>(
+    'POST',
+    route,
+    standardAxiosInstance,
+    body,
+    headers,
+  );
 }
```
MarceloRGonc, line 65: Do not create a new method. Add an optional parameter in the previous one:
```
retryConfigs?: {
  retries: number, // default 3
  retryCondition: fn: (error: any) => boolean,
  retryDelay: (retryCount: number) => number, // some default strategy
}```

File: packages/openops/src/lib/openops-tables/requests-helpers.ts [hunk 11]
```diff
@@ -1,28 +1,93 @@
...
+  AxiosResponse,
+  Method,
+} from 'axios';
+import axiosRetry from 'axios-retry';
+import { StatusCodes } from 'http-status-codes';
 import { makeHttpRequest } from '../axios-wrapper';
 
+const retryAxiosInstance = axios.create();
+
+axiosRetry(retryAxiosInstance, {
```
MarceloRGonc, line 14: Move this method to the axios-wrapper file.

packages/openops/src/lib/axios-wrapper.ts
```diff
@@ -1,11 +1,18 @@
...
+  AxiosResponse,
+  Method,
+} from 'axios';
 
 export async function makeHttpRequest<T>(
   method: Method,
   url: string,
   headers?: AxiosHeaders,
   body?: any,
+  axiosInstance?: AxiosInstance,
```
MarceloRGonc, line 15: Change this to something like:

```
retryConfigs?: {
  retries: number, // default 3
  retryCondition: fn: (error: any) => boolean,
  retryDelay: (retryCount: number) => number, // some default strategy
}
```

If defined create the axios instance with the retries.

packages/openops/src/lib/axios-wrapper.ts
```diff
@@ -1,22 +1,39 @@
 import { logger } from '@openops/server-shared';
 import axios, { AxiosError, AxiosHeaders, Method } from 'axios';
+import axiosRetry, { IAxiosRetryConfig } from 'axios-retry';
+
+const standardAxiosInstance = axios;
+
+let retryAxiosInstance: ReturnType<typeof axios.create> | null = null;
+
+function getRetryAxiosInstance(retryConfigs: IAxiosRetryConfig) {
```
cezudas, line 9: I wanted to have it as top level code (without any function)
```
const retryAxiosInstance = axios.create();
axiosRetry(retryAxiosInstance, { ... });
```

But I just couldn't get `axios-wrapper.test.ts` to pass because the could would run before your test file can mock axios.create(), so the create() call returns an incomplete mock, and axiosRetry() tries to access .interceptors on it → crash.

packages/openops/src/lib/openops-tables/auth-user.ts
```diff
@@ -44,7 +49,11 @@ export async function authenticateDefaultUserInOpenOpsTables(): Promise<AuthToke
     const email = system.getOrThrow(AppSystemProp.OPENOPS_ADMIN_EMAIL);
     const password = system.getOrThrow(AppSystemProp.OPENOPS_ADMIN_PASSWORD);
 
-    tokens = await authenticateUserInOpenOpsTables(email, password);
+    tokens = await authenticateUserInOpenOpsTables(
+      email,
+      password,
+      axiosTablesSeedRetryConfig,
+    );
```
MarceloRGonc, line 56: Why we need to retry?

packages/openops/src/lib/openops-tables/auth-user.ts
```diff
@@ -32,6 +36,7 @@ export async function authenticateUserInOpenOpsTables(
     'api/user/token-auth/',
     requestBody,
     headers,
+    axiosRetryConfig,
```
MarceloRGonc, line 39: Why we need to retry?

packages/openops/src/lib/openops-tables/requests-helpers.ts
```diff
@@ -9,34 +11,73 @@ export function createAxiosHeaders(token: string): AxiosHeaders {
   });
 }
 
+export const axiosTablesSeedRetryConfig: IAxiosRetryConfig = {
+  retries: 3,
+  retryDelay: (retryCount: number) => {
+    return retryCount * 10000;
```
MarceloRGonc, line 17: Isn't that a long time?
10s 20s 30s

packages/openops/src/lib/openops-tables/fields.ts
```diff
@@ -12,6 +13,7 @@ export async function getFields(
   const fields = await makeOpenOpsTablesGet<any[]>(
     `api/database/fields/table/${tableId}/?user_field_names=${userFieldNames}`,
     authenticationHeader,
+    axiosTablesSeedRetryConfig,
```
MarceloRGonc, line 16: Why we need to retry?

packages/openops/src/lib/openops-tables/applications-service.ts
```diff
@@ -13,6 +17,7 @@ export async function getDefaultDatabaseId(
     await makeOpenOpsTablesGet<Application>(
       `api/applications/`,
       authenticationHeader,
+      axiosTablesSeedRetryConfig,
```
MarceloRGonc, line 20: Why we need to retry?

cezudas: it's technically an API call to tables service and `getDefaultDatabaseId` is called also from `packages/server/api/src/app/database/seeds/openops-tables-rename-database.ts`

packages/openops/src/lib/openops-tables/auth-user.ts
cezudas: There is a `authenticateDefaultUserInOpenOpsTables` function in `packages/openops/src/lib/openops-tables/auth-user.ts`. Following function calls we have seed code using it:
packages/server/api/src/app/database/seeds/openops-opportunities-table-seed.ts
packages/server/api/src/app/database/seeds/openops-opportunities-table-seed.ts
packages/server/api/src/app/database/seeds/openops-delete-old-opportunities-table.ts

packages/openops/src/lib/axios-wrapper.ts
cezudas: Code-wise it would be simple to create an instance for every tables request but it has implications 🚫 
- Performance hit: Creating an instance allocates memory and processes configuration each time, which adds overhead in high-frequency or concurrent calls.
- No caching or connection reuse: Recreating instance means not reusing connection pools or interceptors.

I've also considered creating a Pool/Cache of Instances per Retry Config but IAxiosRetryConfig interface definition has functions and I can't think of a reliable way since both hashing and deep comparison will be unreliable.
WDYT? @MarceloRGonc

packages/openops/src/lib/openops-tables/fields.ts
cezudas: I added axiosRetryConfig?: IAxiosRetryConfig param to getFields function in the meantime. So the seed logic call can pass axiosTablesSeedRetryConfig and tables block code can skip retry

packages/openops/src/lib/openops-tables/requests-helpers.ts
cezudas: Alright, I've reduced it by a factor of 10: 1s 2s 3s

bigfluffycookie (approved):
Looks good, just wait until Marcelo replies to the open comment.

packages/openops/src/lib/axios-wrapper.ts
MarceloRGonc: For now, I think we should keep it simple. And create a new instance for every request with retries. Performance won't be affected much and we keep the method reusable. If we notice performance issues, we adopt another strategy, but it's not worth the effort at this point.

packages/openops/src/lib/openops-tables/requests-helpers.ts
```diff
@@ -9,34 +11,73 @@ export function createAxiosHeaders(token: string): AxiosHeaders {
   });
 }
 
+export const axiosTablesSeedRetryConfig: IAxiosRetryConfig = {
+  retries: 3,
+  retryDelay: (retryCount: number) => {
+    return retryCount * 1000;
```
MarceloRGonc, line 17: Add a debug log, to understand if we are doing too many retries

packages/openops/src/lib/axios-wrapper.ts
```diff
@@ -1,22 +1,36 @@
...
       method,
       url,
+      headers,
+      data: body,
     };
 
-    const response = await axios(config);
+    const axiosInstance = retryConfigs
+      ? getRetryAxiosInstance(retryConfigs)
+      : standardAxiosInstance;
```
MarceloRGonc, line 31:
```suggestion
      : axios;
```

packages/openops/src/lib/axios-wrapper.ts
```diff
@@ -1,22 +1,36 @@
 import { logger } from '@openops/server-shared';
 import axios, { AxiosError, AxiosHeaders, Method } from 'axios';
+import axiosRetry, { IAxiosRetryConfig } from 'axios-retry';
+
+const standardAxiosInstance = axios;
```
MarceloRGonc, line 5: You don't need this variable

MarceloRGonc (approved):

PR #432, opened 2025-04-10
Add try catch in sendJiraRequest to not return the whole error object.

bigfluffycookie:
Fixes OPC-57

image

### DIFF
--- a/packages/blocks/jira-cloud/src/lib/common/index.ts
+++ b/packages/blocks/jira-cloud/src/lib/common/index.ts
@@ -24,14 +24,14 @@ export async function sendJiraRequest(
     });
   } catch (e) {
     if (e instanceof HttpError) {
-      const errorBody = e.response?.body || {};
+      const errorBody = e.response?.body ?? {};
       const fullError = JSON.stringify(errorBody);
 
       if (
         fullError.includes('It is not on the appropriate screen, or unknown.')
       ) {
         throw Error(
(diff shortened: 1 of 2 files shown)


packages/blocks/jira-cloud/src/lib/common/index.ts
```diff
@@ -1,25 +1,45 @@
...
+    });
+  } catch (e) {
+    if (e instanceof HttpError) {
+      const errorBody = e.response?.body || {};
+      const fullError = JSON.stringify(errorBody);
+
+      if (
+        fullError.includes('It is not on the appropriate screen, or unknown.')
+      ) {
+        throw Error(
```
bigfluffycookie, line 33: the only issue is, as you can see in the screenshot we are basically throwing an error inside another error, idk how much that bothers you or not.

rita-gorokhod: So is this why the FE writes "Error 1: Jira, Error 2:" ? 
I think just don't add then `Jira Error:`

packages/blocks/jira-cloud/src/lib/common/index.ts
```diff
@@ -1,25 +1,45 @@
...
+          `Jira Error: The field you're trying to set (e.g., 'priority') is not configured on the project's create/edit screen. You need to add it in Jira's screen settings.\n\nOriginal error: ${fullError}`,
```
rita-gorokhod, line 34: I'd remove the e.g., it might confuse even more. 
"One or more of the field(s) you're trying to set is not configured on the project's...

rita-gorokhod:
@bigfluffycookie please address also Sonar comments

PR #433, opened 2025-04-10
Adjust block categories

rita-gorokhod:
Fixes OPS-1558

image

image
image
image

image
image
image

image
image


packages/blocks/sftp/src/index.ts
```diff
@@ -63,7 +63,7 @@ export const sftp = createBlock({
 
   minimumSupportedRelease: '0.7.1',
   logoUrl: 'https://static.openops.com/blocks/sftp.svg',
-  categories: [BlockCategory.NETWORK],
+  categories: [BlockCategory.CORE],
```
rSnapkoOpenOps, line 66: I'm not sure if it's the best choice. We filter by the 'CORE' category when calculating integrations. Since SFTP has a connection, I would consider it an integration. However, if the category is 'CORE,' it won't be considered as an integration

packages/blocks/github/src/index.ts
```diff
@@ -12,6 +12,6 @@ export const github = createBlock({
   logoUrl: 'https://static.openops.com/blocks/github.png',
   authors: [],
   actions: [getFileAction, createPullRequestAction, runWorkflowAction],
-  categories: [BlockCategory.IaC],
+  categories: [BlockCategory.DEVOPS],
```
rSnapkoOpenOps, line 15: I'm not sure if it's DevOps. Maybe better name would be 'Development'

packages/blocks/sftp/src/index.ts
rita-gorokhod: @rSnapkoOpenOps the category is not required, right? maybe I'll just leave it empty right now, like we do for code

packages/react-ui/src/app/features/builder/blocks-selector/block-tag-group.tsx
```diff
@@ -18,42 +18,34 @@ const tags: Record<
     title: 'All',
     color: 'blue',
   },
-  [BlockCategory.WORKFLOW]: {
-    title: 'Workflow',
+  [BlockCategory.FINOPS]: {
+    title: 'FinOps',
```
rSnapkoOpenOps, line 22: We also need to update translations file

packages/blocks/sftp/src/index.ts
rSnapkoOpenOps: Sounds good!

packages/blocks/github/src/index.ts
rita-gorokhod: Marco's choice :D

packages/react-ui/src/app/features/builder/blocks-selector/index.tsx
```diff
@@ -275,7 +275,7 @@ const BlockSelector = ({
     >
       <PopoverTrigger asChild={asChild}>{children}</PopoverTrigger>
       <PopoverContent
-        className="w-[600px] p-0 shadow-lg"
+        className="w-[634px] p-0 shadow-lg"
```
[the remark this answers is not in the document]
rita-gorokhod: true, maybe it should be dynamic with mix and max? wdyt?

packages/react-ui/src/app/features/builder/blocks-selector/index.tsx
```diff
@@ -245,17 +245,22 @@ const BlockSelector = ({
...
+          stepMetadata as BlockStepMetadata
+        ).categories?.includes(selectedTag);
+
         if (selectedTag === BlockCategory.WORKFLOW) {
-          return [
-            ActionType.LOOP_ON_ITEMS,
-            ActionType.SPLIT,
-            ActionType.BRANCH,
-          ].includes(stepMetadata.type as ActionType);
+          return (
```
rita-gorokhod, line 253: @rSnapkoOpenOps please review, it's for this:
image

rSnapkoOpenOps (approved):

PR #434, opened 2025-04-11
Add databricks block

rSnapkoOpenOps:
Fixes OPC-58.

## Additional Notes
https://www.loom.com/share/91bcb5c43d844a61abe2a0a215f4bb69?sid=456161e9-ea8e-4aaa-970a-09f9dfc0afc3

## Testing Checklist

Check all that apply:

image

image


packages/blocks/databricks/src/lib/actions/execute-sql-statement.ts
```diff
@@ -0,0 +1,109 @@
...
+      const status = statusResponse.status?.state;
+      if (status !== 'PENDING' && status !== 'RUNNING') {
+        return statusResponse;
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, RETRY_TIMEOUT));
+    }
+
+    return submission;
+  },
```
rSnapkoOpenOps, line 108: Databricks has a max timeout of 50 seconds for executing a statement. If that’s not enough, the user just gets a result with a "pending" status and no data. That can be a problem if the execution takes longer. So I added custom polling with a longer timeout to handle that

packages/blocks/databricks/src/lib/actions/execute-sql-statement.ts
```diff
@@ -0,0 +1,109 @@
...
+import { makeHttpRequest } from '@openops/common';
+import { AxiosHeaders } from 'axios';
+import { databricksAuth } from '../common/auth';
+import { getDatabricksToken } from '../common/get-databricks-token';
+import { DatabricksSqlExecutionResult } from '../common/sql-execution-result';
+import { warehouseId } from '../common/warehouse-id';
+import { workspaceDeploymentName } from '../common/workspace-deployment-name';
+
+const MAX_QUERY_TIMEOUT_SECONDS = 50;
+const RETRY_TIMEOUT = 15000;
```
bigfluffycookie, line 11: Maybe add "_MILLISECONDS" to the end of it, I find it a bit confusing that the top one specifies the time unit and this one doesnt

packages/blocks/databricks/src/lib/common/warehouse-id.ts
```diff
@@ -0,0 +1,50 @@
...
+    const authValue = auth as BlockPropValueSchema<typeof databricksAuth>;
+    const accessToken = await getDatabricksToken(authValue);
+
+    const workspaceListUrl = `https://${workspaceDeploymentName}.cloud.databricks.com/api/2.0/sql/warehouses`;
+
+    const headers = new AxiosHeaders({
+      Authorization: `Bearer ${accessToken}`,
+    });
+
+    const { warehouses } = await makeHttpRequest<{ warehouses: any[] }>(
```
bigfluffycookie, line 34: What if this fails? Maybe add a try catch with an error dropdown return

packages/blocks/databricks/src/lib/common/workspace-deployment-name.ts
```diff
@@ -0,0 +1,48 @@
...
+    const authValue = auth as BlockPropValueSchema<typeof databricksAuth>;
+    const accessToken = await getDatabricksToken(authValue);
+
+    const workspaceListUrl = `https://accounts.cloud.databricks.com/api/2.0/accounts/${authValue.accountId}/workspaces`;
+
+    const headers = new AxiosHeaders({
+      Authorization: `Bearer ${accessToken}`,
+    });
+
+    const workspaces = await makeHttpRequest<any[]>(
```
bigfluffycookie, line 32: same here, i would add a try catch just in case

bigfluffycookie (changes requested):

packages/blocks/databricks/src/lib/common/warehouse-id.ts
rSnapkoOpenOps: Done

packages/blocks/databricks/src/lib/common/warehouse-id.ts
```diff
@@ -0,0 +1,58 @@
...
+        label: warehouse.name,
+        value: warehouse.id,
+      }));
+
+      return {
+        disabled: false,
+        options: options,
+      };
+    } catch {
+      return {
```
bigfluffycookie, line 51: Sorry could have been more specific, there is an error property now for the dropdowns, i would catch the error and then add it to the error, theres an example in get-price-action.ts

packages/blocks/databricks/test/workspace-deployment-name.test.ts
```diff
@@ -0,0 +1,92 @@
...
+    expect(result).toEqual({
+      disabled: false,
+      options: [
+        { label: 'Workspace One', value: 'workspace-1' },
+        { label: 'Workspace Two', value: 'workspace-2' },
+      ],
+    });
+  });
+
+  it('should throw if makeHttpRequest fails', async () => {
```
bigfluffycookie, line 77: "Should return error and placeholder if makeHttpRequest fails"

bigfluffycookie (approved):
Looks good, just some last minor fixes

packages/blocks/databricks/src/lib/common/warehouse-id.ts
rSnapkoOpenOps: image

PR #437, opened 2025-04-14
Move get price logic into common

bigfluffycookie:
Part of OPS-972

This moves the logic of getPrice into the common package, with this setting the price env variable wont be necessary. There will be a second pr removing redundant and unused code.


packages/openops/src/lib/aws/pricing.ts
```diff
@@ -106,3 +107,29 @@ async function makeRequest(
...
+
+export async function getPriceListWithCache(
+  credentials: any,
+  serviceCode: string,
+  filters: Filter[],
+  region: SupportedPricingRegion,
+): Promise<unknown> {
+  const cacheKey = createCacheKey(serviceCode, filters);
+
+  return await cacheWrapper.getOrAdd(cacheKey, getPriceListFromAws, [
```
rita-gorokhod, line 119: nice, and this uses the redis cache, right? @MarceloRGonc

rita-gorokhod (approved):

PR #438, opened 2025-04-14
Remove unused and redundant pricing code.

bigfluffycookie:
Fixes OPS-972


rita-gorokhod (approved):

packages/openops/src/lib/aws/pricing.ts
```diff
@@ -106,3 +107,29 @@ async function makeRequest(
...
+
+  return await cacheWrapper.getOrAdd(cacheKey, getPriceListFromAws, [
+    credentials,
+    region,
+    serviceCode,
+    filters,
+  ]);
+}
+
+function createCacheKey(serviceCode: string, filters: Filter[]): string {
```
MarceloRGonc, line 127: Should the key be associated with the connection? An invalid connection will fetch the data if it is already cached.

packages/openops/src/lib/aws/pricing.ts
```diff
@@ -62,7 +63,7 @@ export async function getAttributeValues(
   return attributeValues;
 }
 
-export async function getPriceList(
+export async function getPriceListFromAws(
```
MarceloRGonc, line 66: Do we still need export?

packages/openops/src/lib/aws/pricing.ts
bigfluffycookie: hm, I mean I didnt change the behavior, so its always worked like this until now, but I guess I can open another pr for this specifically

MarceloRGonc: Before we only had one connection. Now we can have multiple. @rita-gorokhod WDYT? It's not a problem, we'll have the data. It's just a matter of an invalid connection gaining access. Another point, I don't know if it's possible for different accounts to give different results.

packages/openops/src/lib/aws/pricing.ts
MarceloRGonc: I'll help you with that. But I only realized now that it was the result of another PR.

PR #445, opened 2025-04-15
Add input with autocomplete component

rSnapkoOpenOps:
Part of OPS-1578.

## Testing Checklist

Check all that apply:

image

### DIFF
--- a/packages/ui-components/src/ui/autocomplete-input.tsx
+++ b/packages/ui-components/src/ui/autocomplete-input.tsx
@@ -1,6 +1,12 @@
 import { t } from 'i18next';
 import { Check, ChevronsUpDown } from 'lucide-react';
-import { ChangeEvent, forwardRef, useCallback, useState } from 'react';
+import {
+  ChangeEvent,
+  forwardRef,
+  useCallback,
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/ui/autocomplete-input.tsx
```diff
@@ -0,0 +1,132 @@
...
+      value = '',
+      disabled,
+      className,
+    },
+    ref,
+  ) => {
+    const [open, setOpen] = useState(false);
+    const [inputValue, setInputValue] = useState(value);
+    const [selectValue, setSelectValue] = useState(value);
+
```
cezudas, line 43: If you expect the value prop to be dynamic after the initial render, you might introduce a useEffect to update inputValue and selectValue whenever the value prop changes:

```
useEffect(() => {
  setInputValue(value);
  setSelectValue(value);
}, [value]);
```

rSnapkoOpenOps: Nice catch, fixed

cezudas (approved):

PR #456, opened 2025-04-16
Add databricks "run job" action

rSnapkoOpenOps:
Fixes OPS-1600.

## Testing Checklist

Check all that apply:

https://www.loom.com/share/fd023c9125794d9fb57498c3140971fb?sid=3e7d2ba3-f353-4d6f-97d4-3ad6640cc2a7


packages/blocks/databricks/src/lib/common/job-id.ts
```diff
@@ -0,0 +1,66 @@
...
+        value: job.job_id,
+      }));
+
+      return {
+        disabled: false,
+        options: options,
+      };
+    } catch (error: any) {
+      let errorMessage;
+      try {
```
bigfluffycookie, line 52: Why are you doing this? I'd just return error.message, the double try catch isnt nice to read

bigfluffycookie (approved):
Looks good, just a minor comment

packages/blocks/databricks/src/lib/common/job-id.ts
rSnapkoOpenOps: Their error structure may be different, so I'm not sure weather `JSON.parse(error.message)?.message` throws an error or not

bigfluffycookie: yea which is why I think its probably okay to just return "erros.message" no risk of anything throwing

rSnapkoOpenOps: The `error.message` is `{
  "error_code": "
REQUEST_LIMIT_EXCEEDED",
  "message": "Request is rejected due to throttling."
}`. Which can be too long and not user friendly. So I suggest to extract the error message

packages/blocks/databricks/src/lib/actions/run-job.ts
```diff
@@ -0,0 +1,86 @@
...
+      required: false,
+      description: 'Optional parameter values to bind job.',
+    }),
+    timeout: Property.Number({
+      displayName: 'Task timeout',
+      description:
+        'Maximum number of seconds to wait for a task to complete before timing out.',
+      required: true,
+      validators: [Validators.number, Validators.minValue(0)],
+      defaultValue: 50,
```
MarceloRGonc, line 30: MAX_QUERY_TIMEOUT_SECONDS?

packages/blocks/databricks/src/lib/common/job-id.ts
bigfluffycookie: I get your point, but the error message is not meant to be super userfriendly, that's what we use the placeholder for IMO. I also think the error_code could be useful, that way a user can search the docs themselves rather than having to try and find it my error message . On top of that its just not super consistent with out other code, in all instances I could find we do catch(error){
  error: error,
}

packages/blocks/databricks/src/lib/common/wait-for-task-completion.ts
```diff
@@ -0,0 +1,44 @@
...
+      state !== 'BLOCKED' &&
+      state !== 'QUEUED'
+    ) {
+      return output;
+    }
+
+    await new Promise((resolve) =>
+      setTimeout(resolve, RETRY_TIMEOUT_MILLISECONDS),
+    );
+  }
```
MarceloRGonc, line 41: The method 'makeHttpRequest' accepts a retry config. That will do the retries for you.

packages/blocks/databricks/src/lib/actions/run-job.ts
```diff
@@ -0,0 +1,86 @@
...
+      number_in_job: number;
+    }>('POST', runUrl, headers, {
+      job_id: jobId,
+      notebook_params: parameters,
+    });
+
+    const runDetailsUrl = `https://${workspaceDeploymentName}.cloud.databricks.com/api/2.2/jobs/runs/get?run_id=${runId}&include_resolved_values=true`;
+    const runDetails = await makeHttpRequest<{
+      tasks: { run_id: number; task_key: string }[];
+    }>('GET', runDetailsUrl, headers);
```
MarceloRGonc, line 56: I think you should move this to a service. We have two actions with this type of requests. Also a question why does executeSqlStatement use version 2.0 and runJob uses 2.2?

rSnapkoOpenOps: That's based on their docs

https://docs.databricks.com/api/workspace/statementexecution/executestatement
https://docs.databricks.com/api/workspace/jobs/list

packages/blocks/databricks/src/lib/actions/run-job.ts
rSnapkoOpenOps: No, the values are independent. 

`MAX_QUERY_TIMEOUT_SECONDS` - limitation of the api for some queries 
 `defaultValue: 50` approximate time for execution for one action. It's kinda placeholder

packages/blocks/databricks/src/lib/common/job-id.ts
rita-gorokhod: I agree @rSnapkoOpenOps , since this is an API wrapper, it's best to display the full error so that the user could google it.

packages/blocks/databricks/src/lib/actions/run-job.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const taskOutputs = [];
+
+    for (const task of runDetails.tasks) {
+      try {
+        const output = await waitForTaskCompletion({
```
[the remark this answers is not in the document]
rSnapkoOpenOps: You can set timeout to 0 if you don't want to wait

packages/blocks/databricks/src/lib/common/job-id.ts
rSnapkoOpenOps: done

MarceloRGonc (approved):

PR #473, opened 2025-04-21
Add createLanguageModel to AI providers

MarceloRGonc:
Part of OPS-1588


packages/openops/src/lib/ai/providers/amazon-bedrock.ts
MarceloRGonc, line 1: This provider does not work with API Key. It is a connection like the ones we have in the AWS block.

AWS_ACCESS_KEY_ID=YOUR_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY=YOUR_SECRET_ACCESS_KEY
AWS_REGION=YOUR_REGION

packages/openops/src/lib/ai/providers/lmnt.ts
MarceloRGonc, line 1: I don't think we are able to support this provider, which is a text-to-speech.

bigfluffycookie: Ah youre right, then you can remove it.

packages/openops/src/lib/ai/providers/amazon-bedrock.ts
bigfluffycookie: hmmm maybe for now we leave it out then, or we could let the user provide it via the model settings but thats not great

bigfluffycookie (approved):

PR #478, opened 2025-04-23
Add get config and list configs

bigfluffycookie:
Fixes OPS-1576

Adds three new endpoints, one to list all AI configs, one to get an AI config by ID and one to get the currently active config


packages/server/api/src/app/ai/config/ai-config.service.ts
```diff
@@ -45,4 +49,42 @@ export const aiConfigService = {
...
+      ...config,
+      apiKey: AiApiKeyRedactionMessage,
+    }));
+
+    return redactedConfigs;
+  },
+
+  async get(params: { projectId: string; id: string }): Promise<AiConfig> {
+    const { projectId, id } = params;
+    const config = await repo().findOneByOrFail({
```
MarceloRGonc, line 68: This throws an exception. I think you want to return null so findOne

packages/server/api/src/app/ai/config/ai-config.service.ts
```diff
@@ -45,4 +64,54 @@ export const aiConfigService = {
...
   },
+
+  async list(projectId: string): Promise<AiConfigRedacted[]> {
+    const configs = await repo().findBy({
+      projectId,
+    });
+
+    const redactedConfigs = configs.map((config) => ({
+      ...config,
+      apiKey: AiApiKeyRedactionMessage,
```
MarceloRGonc, line 75: Suggestion: create a method to do the redact

PR #483, opened 2025-04-23
Extend MD component to add inject code variation

alexandrudanpop:
Fixes OPS-1616

## Testing Checklist

Check all that apply:

See Chromatic


packages/ui-components/src/components/custom/markdown.tsx
```diff
@@ -42,7 +44,54 @@ const Container = ({
...
+}: {
+  content: string;
+  codeVariation?: CodeVariations;
+}) => {
+  const textareaRef = useRef<HTMLTextAreaElement>(null);
+
+  const isInjectVariation =
+    codeVariation === MarkdownCodeVariations.WithCopyAndInject;
+
+  useEffect(() => {
```
alexandrudanpop, line 59: we need the useEffect to adjust the textarea based on scrollHeight (for when LLM output will return multi-line commands)

packages/ui-components/src/components/custom/types.ts
```diff
@@ -0,0 +1,7 @@
+export enum MarkdownCodeVariations {
```
cezudas, line 1: Looks great!

Maybe a topic for R&D weekly - not critical, not new: we have over 100 enums in the codebase and I think in 9/10 cases we don't actualy use a feature that can't be achieved with `as const` objects + derived union types so some unecesarry "enum" runtime code always ends up in the final bundle.
Enum code snippet transpiled to JS:
Screenshot 2025-04-24 at 11 08 23
 `as const` objects + derived union code snippet transpiled to JS:
 
Screenshot 2025-04-24 at 11 08 45

cezudas (changes requested):

packages/ui-components/src/components/custom/types.ts
alexandrudanpop: the only reason I choose to use an `enum` here, otherwise I'm not really a fan of them, is having it easier to get a union of the values, see `CodeVariations` below

cezudas: It's just as easy
Screenshot 2025-04-24 at 11 25 32

alexandrudanpop: IMO it's just not so readable to get this union:
```ts
export type CodeVariations =
  (typeof MarkdownCodeVariations)[keyof typeof MarkdownCodeVariations];
```

cezudas: ¯\_(ツ)_/¯  it's limited to a single module, and it's enough to hover over the type in a IDE/Github codespace. That's how I tend to review PRs anyway.
From the outside (reviewing another module) it's just a plain import and it doesn't change a thing.

cezudas (approved):

PR #484, opened 2025-04-23
Wire up clickable Generate with AI CTA

cezudas:
Fixes OPS-1596.

Demo:
https://www.loom.com/share/d6780882e35045fbb94ee108e71a17d2
The version demo-ed in Loom is missing one feature, which is to make the AIChat window aware of the property "it was opened from".  This is needed later on for the "inject command" functionality.
The screenshot below shows the latest code rendering the `property.displayName`
Screenshot 2025-04-23 at 18 44 04


packages/ui-components/src/components/ai-chat-container/ai-chat-container.tsx
```diff
@@ -82,8 +84,11 @@ const AiChatContainer = ({
       >
         <ScrollArea className="transition-all h-full w-full">
           <div className="py-8 flex flex-col h-full">
-            <div className="justify-center dark:text-primary text-base font-bold leading-[25px] flex-1 px-6">
-              <span>Welcome to OpenOps AI Chat!</span>
+            <div className="flex-1 px-6">
+              <div className="justify-center dark:text-primary text-base font-bold leading-[25px]">
+                <span>Welcome to OpenOps AI Chat!</span>
```
alexandrudanpop, line 89: need to wrap with translate function

packages/react-ui/src/app/features/builder/ai-chat/conversation.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const { isPending, data } = useQuery({
+    queryKey: ['openChat', flowVersion.flowId, blockName, stepName],
+    queryFn: async () => {
+      if (!stepDetails) {
+        throw new Error('Step not found');
+      }
+      return aiChatApi.open(flowVersion.flowId, blockName, stepName);
+    },
+    enabled: !!stepDetails && !!stepDetails.settings.blockName,
+    staleTime: Infinity,
```
alexandrudanpop, line 30: I think we need to call the endpoint every time a step settings is opened for an action to get the conversation history. I can imagine at some point someone could try to use it across different tabs, or when logged in to different browsers.

packages/react-ui/src/app/features/builder/ai-chat/conversation.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const stepDetails = flowHelper.getStep(flowVersion, stepName);
+  const blockName = stepDetails?.settings?.blockName;
+
+  const { isPending, data } = useQuery({
+    queryKey: ['openChat', flowVersion.flowId, blockName, stepName],
+    queryFn: async () => {
+      if (!stepDetails) {
+        throw new Error('Step not found');
+      }
+      return aiChatApi.open(flowVersion.flowId, blockName, stepName);
```
alexandrudanpop, line 27: Maybe not related to the FE implementation, but I think we should have the userId/org combination here as well, not to show messages from another user. WDYT @MarceloRGonc ?

alexandrudanpop: I mean we should not modify it here, as we have that info in the BE session

packages/react-ui/src/app/features/builder/ai-chat/conversation.tsx
cezudas: alright, I removed `staleTime: Infinity,` and pushed
https://github.com/openops-cloud/openops/pull/484/commits/3957e0d0a52398f655a9eaf94c18b2fe547f0f33

packages/react-ui/src/app/features/builder/ai-chat/conversation.tsx
cezudas: Alright, @MarceloRGonc and @alexandrudanpop DM'ed each other and aligned on this -> not in the scope of this PR

alexandrudanpop (approved):

PR #489, opened 2025-04-24
Add env var to control execFile max buffer

bigfluffycookie:
Fixes OPS-1618


maor-rozenfeld (commented):
What's the default value? Shouldn't we set it explicitly?

deploy/docker-compose/.env.defaults
```diff
@@ -16,7 +16,7 @@ OPS_TRIGGER_DEFAULT_POLL_INTERVAL=1
 OPS_WEBHOOK_TIMEOUT_SECONDS=30
 OPS_BLOCKS_SOURCE='FILE'
 OPS_BLOCKS_SYNC_MODE=NONE
-
+OPS_EXECFILE_MAX_BUFFERSIZE_MB=
```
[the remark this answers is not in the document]
bigfluffycookie: I thought we have to set it here, do we not? theres lots of unused env vars here so i figured it must be in this file aswell

bigfluffycookie:
> What's the default value? Shouldn't we set it explicitly?

I think its around 1MB but it doesnt actually say, we can set it if we want but i just wasnt sure how much, what about 2 maybe?

deploy/docker-compose/.env.defaults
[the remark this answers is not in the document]
maor-rozenfeld: We don't _have_ to add it, that's for sure, because it will just stay empty if we don't. It looks like an internal, edge-case specific var that won't be relevant to most users. So I think we don't need it here.

packages/server/shared/src/lib/system/system-prop.ts
```diff
@@ -95,6 +95,8 @@ export enum AppSystemProp {
   CODE_BLOCK_MEMORY_LIMIT_IN_MB = 'CODE_BLOCK_MEMORY_LIMIT_IN_MB',
 
   CANDU_CLIENT_TOKEN = 'CANDU_CLIENT_TOKEN',
+
+  EXECFILE_MAX_BUFFERSIZE_MB = 'EXECFILE_MAX_BUFFERSIZE_MB',
```
maor-rozenfeld, line 99: The naming is a bit all over the place, `EXEC_FILE_MAX_BUFFER_SIZE_MB`

bigfluffycookie: Yea the german grammar was strong in that one

maor-rozenfeld: Ah ireallylike germangrammer it's soeasytounderstandish

maor-rozenfeld (approved):

PR #490, opened 2025-04-24
Add AWS block AI prompt

MarceloRGonc:
Fixes OPS-1583.


packages/server/api/src/app/ai/chat/prompts.service.ts
```diff
@@ -1,31 +1,32 @@
...
+  } catch (error) {
+    logger.error('', error);
+    return '';
   }
 };
 
-const awsCliPrompt = `
-You are a cloud infrastructure assistant skilled in AWS.
-Your job is to help users generate correct and efficient CLI commands for any of these cloud providers.
+async function loadFile(filename: string): Promise<string> {
```
MarceloRGonc, line 26: For now, we are reading the files from the same deployment. But I think the idea is to get it directly from the repository.

PR #502, opened 2025-04-25
Replace upsert config with save config

bigfluffycookie:
Fixes OPS-1631

### DIFF
--- a/packages/server/api/src/app/ai/config/ai-config.service.ts
+++ b/packages/server/api/src/app/ai/config/ai-config.service.ts
@@ -47,7 +47,7 @@ export const aiConfigService = {
 
     const aiConfig: Partial<AiConfig> = {
       ...request,
-      id: existing?.id ?? openOpsId(),
+      id: request?.id ?? openOpsId(),
       projectId,
       created: existing?.created ?? new Date().toISOString(),
       updated: new Date().toISOString(),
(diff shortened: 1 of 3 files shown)


packages/server/api/src/app/ai/config/ai-config.service.ts
```diff
@@ -26,43 +26,31 @@ function redactApiKey(config: AiConfig): AiConfigRedacted {
 }
 
 export const aiConfigService = {
-  async upsert(params: {
+  async save(params: {
```
MarceloRGonc, line 29: I think upsert makes more sense. You are either inserting or updating.

MarceloRGonc (approved):

packages/server/api/src/app/ai/config/ai-config.service.ts
bigfluffycookie: yes but right now save makes more sense, it will override the existing one without having to give it conditions.

MarceloRGonc: The creation date will always be the first.

bigfluffycookie: Yes, it was the same before. Unless you delete the entity it stays the same.

PR #513, opened 2025-04-28
Add providerSettings to createLanguageModel

bigfluffycookie:
Fixes OPS-1633


packages/openops/src/lib/ai/providers.ts
```diff
@@ -72,11 +74,14 @@ export const getAiProviderLanguageModel = async (aiConfig: {
   providerSettings?: Record<string, unknown> | null;
 }): Promise<LanguageModelV1> => {
   const aiProvider = getAiProvider(aiConfig.provider);
+  const { baseUrl: _ignored, ...otherProviderSettings } =
```
bigfluffycookie, line 77: this is a bit ugly

rita-gorokhod: Why are you taking out the baseUrl?
can't we use the provider settings for everything, including baseUrl?

MarceloRGonc: The sanitize is only for not passing empty values to the createMethod.
Rename sanitizeBaseUrl -> sanitizeProviderSettings. Apply it to all the values in providerSettings Record.  Pass the result to the createLanguageModel method.

packages/openops/src/lib/ai/providers/anthropic.ts
```diff
@@ -19,10 +19,12 @@ function createLanguageModel(params: {
   apiKey: string;
   model: string;
   baseUrl?: string;
```
MarceloRGonc, line 21: You can remove this parameter and use only the spread of providerSettings

packages/openops/src/lib/ai/providers.ts
```diff
@@ -72,11 +73,14 @@ export const getAiProviderLanguageModel = async (aiConfig: {
...
+    aiConfig.providerSettings ?? {};
 
   return aiProvider.createLanguageModel({
     apiKey: aiConfig.apiKey,
     model: aiConfig.model,
     baseUrl: sanitizeBaseUrl(aiConfig.providerSettings),
+    ...otherProviderSettings,
```
MarceloRGonc, line 83: Don't need the spread you can pass the Record.

packages/openops/src/lib/ai/providers/anthropic.ts
bigfluffycookie: leaving it as we discussed that we will add it as its own field in the frontend anyways so we will need it again once thats implemented

packages/openops/src/lib/ai/providers.ts
```diff
@@ -72,11 +72,19 @@ export const getAiProviderLanguageModel = async (aiConfig: {
   providerSettings?: Record<string, unknown> | null;
 }): Promise<LanguageModelV1> => {
   const aiProvider = getAiProvider(aiConfig.provider);
-
+  const sanitizedSettings = sanitizeProviderSettings(aiConfig.providerSettings);
+  const mappedProviderSettings: Record<string, unknown> = {
```
bigfluffycookie, line 76: we could also just rename it in the frontend, to be "baseURL" instead i dont mind either way

rita-gorokhod: that would be much cleaner imo

packages/openops/src/lib/ai/providers.ts
```diff
@@ -137,3 +129,23 @@ const invalidConfigError = (
...
 };
+
+function sanitizeProviderSettings(
+  settings: Record<string, unknown> | null | undefined,
+): Record<string, unknown> {
+  if (!settings) return {};
+
+  const sanitized: Record<string, unknown> = {};
+
+  for (const [key, value] of Object.entries(settings)) {
```
rita-gorokhod, line 140: why is this relevant?
is there an error if we pass settings that are null/undefined/empty?

bigfluffycookie: hmm not sure, I've inherited it from @MarceloRGonc

packages/openops/src/lib/ai/providers/openai-compatible.ts
```diff
@@ -5,12 +5,16 @@ import { AiProvider } from '../providers';
...
 }): LanguageModelV1 {
+  if (!params.providerSettings?.['baseURL']) {
+    throw new Error('baseURL is required for OpenAI-compatible providers');
+  }
   return createOpenAICompatible({
     name: 'open-ai-compatible-provider',
     apiKey: params.apiKey,
-    baseURL: params.baseUrl,
+    ...params.providerSettings,
+    baseURL: params.providerSettings['baseURL'] as string,
```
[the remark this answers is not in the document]
bigfluffycookie: I changed it a bit now, but yes i need to specifically pass it or else typescript complains

MarceloRGonc (approved):

rita-gorokhod (approved):

PR #519, opened 2025-04-29
Enable AI via env variable

alexandrudanpop:
Fixes OPS-1648

Use variable:
`OPS_SHOW_AI_SETTINGS=true`

### DIFF
--- a/packages/server/api/src/app/flags/flag.service.ts
+++ b/packages/server/api/src/app/flags/flag.service.ts
@@ -6,7 +6,7 @@ import {
   system,
   webhookSecretsUtils,
 } from '@openops/server-shared';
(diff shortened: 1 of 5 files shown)


packages/server/api/src/app/flags/flag.service.ts
```diff
@@ -290,7 +290,7 @@ export const flagService = {
       },
       {
         id: FlagId.SHOW_AI_SETTINGS,
-        value: false,
+        value: system.get<string>(AppSystemProp.SHOW_AI_SETTINGS),
```
MarceloRGonc, line 293: Isn't it better to use getBoolean?

alexandrudanpop: fixed

PR #525, opened 2025-04-29
Get decrypted api key to validate ai provider

bigfluffycookie:
Fixes OPS-1653

### DIFF
--- a/packages/server/api/src/app/ai/config/ai-config.controller.ts
+++ b/packages/server/api/src/app/ai/config/ai-config.controller.ts
@@ -6,6 +6,7 @@ import { validateAiProviderConfig } from '@openops/common';
 import { AiConfig, PrincipalType, SaveAiConfigRequest } from '@openops/shared';
 import { StatusCodes } from 'http-status-codes';
 import { entitiesMustBeOwnedByCurrentProject } from '../../authentication/authorization';
+import { encryptUtils } from '../../helper/encryption';
 import { aiConfigService } from './ai-config.service';
 
 export const aiConfigController: FastifyPluginAsyncTypebox = async (app) => {
@@ -15,7 +16,26 @@ export const aiConfigController: FastifyPluginAsyncTypebox = async (app) => {
     '/',
     SaveAiConfigOptions,
     async (request, reply): Promise<AiConfig> => {
-      const { valid, error } = await validateAiProviderConfig(request.body);
+      let existingApiKey = request.body.apiKey;
+      if (request.body.id) {
+        const exisiting = await aiConfigService.get(
+          {
+            projectId: request.principal.projectId,
+            id: request.body.id,
+          },
+          false,
+        );
+
+        existingApiKey = exisiting
+          ? encryptUtils.decryptString(JSON.parse(exisiting.apiKey))
+          : existingApiKey;
+      }
+
+      const { valid, error } = await validateAiProviderConfig({
+        ...request.body,
+        apiKey: existingApiKey,
+      });
+
       if (!valid) {
         return reply.status(StatusCodes.BAD_REQUEST).send(error);
       }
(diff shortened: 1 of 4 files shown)


packages/server/api/src/app/ai/config/ai-config.controller.ts
```diff
@@ -15,7 +16,26 @@ export const aiConfigController: FastifyPluginAsyncTypebox = async (app) => {
...
     async (request, reply): Promise<AiConfig> => {
-      const { valid, error } = await validateAiProviderConfig(request.body);
+      let existingApiKey = request.body.apiKey;
+      if (request.body.id) {
+        const exisiting = await aiConfigService.get(
+          {
+            projectId: request.principal.projectId,
+            id: request.body.id,
+          },
+          false,
```
rita-gorokhod, line 26: what is this false?

bigfluffycookie: default the get method returns **REDACTED** with this flag it returns me the actual apiKey

rita-gorokhod: I think it's very unclear, it would be better to have a separate method, i.e. getWithApiKey

packages/server/api/src/app/ai/config/ai-config.controller.ts
```diff
@@ -15,7 +16,26 @@ export const aiConfigController: FastifyPluginAsyncTypebox = async (app) => {
     '/',
     SaveAiConfigOptions,
     async (request, reply): Promise<AiConfig> => {
-      const { valid, error } = await validateAiProviderConfig(request.body);
+      let existingApiKey = request.body.apiKey;
+      if (request.body.id) {
+        const exisiting = await aiConfigService.get(
```
rita-gorokhod, line 21: typo, existingConfig

packages/server/api/src/app/ai/config/ai-config.controller.ts
```diff
@@ -15,7 +16,26 @@ export const aiConfigController: FastifyPluginAsyncTypebox = async (app) => {
...
+          {
+            projectId: request.principal.projectId,
+            id: request.body.id,
+          },
+          false,
+        );
+
+        existingApiKey = exisiting
+          ? encryptUtils.decryptString(JSON.parse(exisiting.apiKey))
```
rita-gorokhod, line 30: why do we need a JSON parse?

packages/server/api/src/app/ai/config/ai-config.controller.ts
bigfluffycookie: Best i cant offer is adding it to params, so you have to write { getRedacted: false }

packages/server/api/src/app/ai/config/ai-config.controller.ts
bigfluffycookie: because its an encrypted object, not a string

packages/server/api/src/app/ai/config/ai-config.controller.ts
bigfluffycookie: nvm i can not offer that, but if you look at the function signature its clear. If you want we can add a cleanup ticket for it either way i dont think its relevant fixing it right now.

bigfluffycookie: getActiveConfig gets the currently enabled config. get config gets the config based on ID.

packages/server/api/src/app/ai/config/ai-config.controller.ts
rita-gorokhod: I don't understand, the method is called decryptString.
Isn't it `encryptUtils.decryptString(exisiting.apiKey)`?

rita-gorokhod: ew as we discussed, it's "decryptToAString"

PR #527, opened 2025-04-29
Improve messages style

alexandrudanpop:
Fixes OPS-1615

As textarea was quite problematic for this usecase, I think it's better to use content editable div in this case.

Please check Chromatic


rSnapkoOpenOps:
Can we change the scroll bar? 
image

alexandrudanpop:
> Can we change the scroll bar? image

This PR just improves the messages style, not the textarea for the prompt. But if you think we need it, can you add a ticket?

rSnapkoOpenOps (approved):

PR #530, opened 2025-04-30
Add ability to read prompts from github

MarceloRGonc:
Fixes OPS-1663.

## Additional Notes
- By default, we read the files from the main branch.
- If INTERNAL_PROMPTS_LOCATION is set to empty, we read the prompts from local files
- We can use the INTERNAL_PROMPTS_LOCATION to read the prompts from a different branch


packages/server/api/src/app/ai/chat/prompts.service.ts
```diff
@@ -1,32 +1,36 @@
...
-        return await loadFile('gcp-cli.txt');
-      default:
-        return '';
-    }
-  } catch (error) {
-    logger.error('', error);
-    return '';
+  switch (context.blockName) {
+    case '@openops/block-aws':
+      return loadFile('aws-cli.txt');
```
rita-gorokhod, line 9: why did you remove the await?

packages/server/shared/src/lib/system/system.ts
```diff
@@ -83,6 +83,8 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
   [SharedSystemProp.INTERNAL_OAUTH_PROXY_URL]: 'https://oauth.openops.com',
   [AppSystemProp.CODE_BLOCK_MEMORY_LIMIT_IN_MB]: '128',
   [SharedSystemProp.INTERNAL_PARALLEL_LOOP_ITERATIONS_LIMIT]: '1',
+  [AppSystemProp.INTERNAL_PROMPTS_LOCATION]:
```
rita-gorokhod, line 86: what about local?

packages/server/shared/src/lib/system/system-prop.ts
```diff
@@ -99,6 +99,8 @@ export enum AppSystemProp {
   EXEC_FILE_MAX_BUFFER_SIZE_MB = 'EXEC_FILE_MAX_BUFFER_SIZE_MB',
 
   SHOW_AI_SETTINGS = 'SHOW_AI_SETTINGS',
+
+  INTERNAL_PROMPTS_LOCATION = 'INTERNAL_PROMPTS_LOCATION',
```
rita-gorokhod, line 103: Minor, I think it should be AI_PROMPTS_LOCATION

MarceloRGonc: You mean: INTERNAL_AI_PROMPTS_LOCATION or AI_PROMPTS_LOCATION. We are using INTERNAL for environment variables that we don't want the user to change

packages/server/shared/src/lib/system/system.ts
MarceloRGonc: Updated the description

packages/server/api/src/app/ai/chat/prompts.service.ts
```diff
@@ -1,32 +1,58 @@
...
+
+async function loadFromCloud(
+  promptsLocation: string,
+  filename: string,
+): Promise<string> {
+  const slash = promptsLocation.endsWith('/') ? '' : '/';
+  const promptFile = `${promptsLocation}${slash}${filename}`;
+
+  const response = await fetch(promptFile);
+  if (!response.ok) {
```
bigfluffycookie, line 49: this will throw if response is undefined

packages/server/api/test/unit/ai/prompts.service.test.ts
```diff
@@ -0,0 +1,183 @@
...
+}));
+
+import { getSystemPrompt } from '../../../src/app/ai/chat/prompts.service';
+
+describe('getSystemPrompt', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should load aws cli block prompt from cloud', async () => {
```
bigfluffycookie, line 29: these tests are all identical except for "cloud" / "local", maybe add a common function?

packages/server/shared/src/lib/system/system-prop.ts
MarceloRGonc: Changed to: AI_PROMPTS_LOCATION

rita-gorokhod (approved):

PR #538, opened 2025-05-01
Fix test output for split inside loop

rSnapkoOpenOps:
Fixes OPS-1670.

## Testing Checklist

Check all that apply:

- Screenshots
- Loom video
- Preview deployment link

### DIFF
--- a/packages/react-ui/src/app/features/flow-runs/lib/flow-run-utils.ts
+++ b/packages/react-ui/src/app/features/flow-runs/lib/flow-run-utils.ts
@@ -222,6 +222,20 @@ function findStepParents(
       return [step, ...pathFromFalseBranch];
     }
   }
+  if (step.type === ActionType.SPLIT) {
+    const pathToAllBranches = step.branches
+      .map((branch) => {
+        return branch.nextAction
+          ? findStepParents(stepName, branch.nextAction)
+          : undefined;
+      })
+      .flat()
+      .filter(Boolean) as Action[];
+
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/app/features/flow-runs/lib/flow-run-utils.ts
```diff
@@ -222,6 +222,19 @@ function findStepParents(
       return [step, ...pathFromFalseBranch];
     }
   }
+  if (step.type === ActionType.SPLIT) {
```
alexandrudanpop, line 225: I think we don't have any tests in this area, @rSnapkoOpenOps do you think we should add some initial tests that also cover the SPLIT for `extractStepOutput`?

rSnapkoOpenOps: Done

packages/react-ui/src/app/features/flow-runs/lib/tests/flow-run-utils.test.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    const result = flowRunUtils.extractStepOutput(
+      'stepInBranch2',
+      {},
+      output as any,
+      trigger as any,
+    );
+    expect(result).toEqual(undefined);
+  });
+});
```
alexandrudanpop, line 191: can you add 2 cases where branch and split also return output?

alexandrudanpop (approved):

PR #539, opened 2025-05-01
Align the github issue templates and links

maor-rozenfeld:
Fixes OPS-1676.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,7 +1,7 @@
 ---
 name: "Report a Bug"
 about: "Report a bug in the OpenOps platform"
-labels: ["bug"]
+labels: ["Bug"]
 assignees: ""
 ---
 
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -1,34 +1,33 @@
 ---
-name: "General Issue"
-about: "Describe your issue here"
-title: "[Issue]: "
-labels: ["needs-triage"]
+name: "Report a Bug"
+about: "Report a bug in the OpenOps platform or its documentation"
```
alexandrudanpop, line 3: Not sure if documentation should be here at all, but if we keep it here we could change this to:

```suggestion
about: "Report a bug in the OpenOps platform or a problem with the documentation"
```

maor-rozenfeld: Yea you're right I'll remove it

.github/ISSUE_TEMPLATE/bug.md
```diff
@@ -1,34 +1,33 @@
...
+labels: ["bug"]
```
alexandrudanpop, line 4: Labels should be uppercase like in Linear

alexandrudanpop (approved):

PR #540, opened 2025-05-02
Select duplicated step

alexandrudanpop:
Fixes OPC-65

Will add a second [PR](https://github.com/openops-cloud/openops/pull/542/files) where I add some tests (as doing this requires refactoring out the builder state and update flow version function)

https://www.loom.com/share/24dcd6151b3e416fb1d9373516862d11

### DIFF
--- a/packages/react-ui/src/app/features/builder/builder-hooks.ts
+++ b/packages/react-ui/src/app/features/builder/builder-hooks.ts
@@ -597,6 +597,17 @@ const applyMidpanelAction = (state: BuilderState, action: MidpanelAction) => {
   };
 };
 
+async function deleteChatRequest(flowVersion: FlowVersion, stepName: string) {
+  try {
+    const stepDetails = flowHelper.getStep(flowVersion, stepName);
+    const blockName = stepDetails?.settings?.blockName;
+    const chat = await aiChatApi.open(flowVersion.flowId, blockName, stepName);
+    await aiChatApi.delete(chat.chatId);
+  } catch (err) {
+    console.error(err);
+  }
+}
+
 const updateFlowVersion = (
   state: BuilderState,
   operation: FlowOperationRequest,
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/app/features/builder/update-flow-version.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  if (operation.type === FlowOperationType.DUPLICATE_ACTION) {
+    set({
+      selectedStep: flowHelper.getStep(
+        newFlowVersion,
+        operation.request.stepName,
+      )?.nextAction?.name,
+    });
+  }
```
alexandrudanpop, line 43: this logic was added

packages/react-ui/src/app/features/builder/builder-hooks.ts
```diff
@@ -597,6 +597,17 @@ const applyMidpanelAction = (state: BuilderState, action: MidpanelAction) => {
   };
 };
 
+async function deleteChatRequest(flowVersion: FlowVersion, stepName: string) {
```
alexandrudanpop, line 600: I extracted this function, I think it's cleaner for it to be outside `updateFlowVersion`

packages/react-ui/src/app/features/builder/builder-hooks.ts
```diff
@@ -597,6 +597,17 @@ const applyMidpanelAction = (state: BuilderState, action: MidpanelAction) => {
...
+  try {
+    const stepDetails = flowHelper.getStep(flowVersion, stepName);
+    const blockName = stepDetails?.settings?.blockName;
+    const chat = await aiChatApi.open(flowVersion.flowId, blockName, stepName);
+    await aiChatApi.delete(chat.chatId);
+  } catch (err) {
+    console.error(err);
+  }
```
rSnapkoOpenOps, line 608: Maybe we should add a toast to catch block?

packages/react-ui/src/app/features/builder/builder-hooks.ts
```diff
@@ -619,19 +630,13 @@ const updateFlowVersion = (
...
-      const chat = await aiChatApi.open(
-        newFlowVersion.flowId,
-        blockName,
-        stepName,
-      );
-      await aiChatApi.delete(chat.chatId);
-    } catch (err) {
-      console.error(err);
-    }
+  if (operation.type === FlowOperationType.DUPLICATE_ACTION) {
```
rita-gorokhod, line 633: @alexandrudanpop did you see my comment on the ticket, that it might be better to do this on move, rather on duplicate?

packages/react-ui/src/app/features/builder/builder-hooks.ts
alexandrudanpop: I don't think so, this is an internal cleanup we are doing for the chat API, the user should not be concerned, and it will hit the catch quite often, because there are many steps a user could delete that don't have an AI chat session open

packages/react-ui/src/app/features/builder/builder-hooks.ts
alexandrudanpop: will do a separate PR for drag as discussed and will keep the new behavior for duplicate

PR #541, opened 2025-05-02
Add logic to redact connection secrets and add new endpoint to get connection with id

bigfluffycookie:
Fixes OPS-1674


packages/server/api/src/app/app-connection/app-connection-utils.ts
```diff
@@ -1,11 +1,84 @@
...
+): AppConnectionWithoutSensitiveData | RedactedAppConnection {
+  const value: Record<string, any> | undefined = connection.value;
+  if (!auth || !value) {
+    return removeSensitiveData(connection);
+  }
+
+  const redacted = { ...value };
+
+  switch (auth.type) {
+    case PropertyType.SECRET_TEXT: {
```
bigfluffycookie, line 30: In this case we could technically just use "removeSensitiveData", as the only thing in the properties will be the secret text but idk if its better to stay consistent or not

packages/server/api/src/app/app-connection/app-connection-utils.ts
```diff
@@ -1,11 +1,90 @@
...
+        typeof value.client_id === 'string' &&
+        typeof value.redirect_url === 'string'
+      ) {
+        return {
+          ...connection,
+          value: {
+            type: PropertyType.OAUTH2,
+            client_id: value.client_id,
+            client_secret: REDACTED_MESSAGE,
+            redirect_url: value.redirect_url,
```
bigfluffycookie, line 78: so, for oauth 2 theres two options:

1) the user connects entierly through the pop up window
2) the user uses their own client id and password

Now, im not sure we need to return redirect_url, i know somewhere we take it from an env variable anyways but idk if the frontend or the backend does that

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -136,3 +163,21 @@ const DeleteAppConnectionRequest = {
     },
   },
 };
+
+const GetAppConnectionRequest = {
+  config: {
+    allowedPrincipals: [PrincipalType.USER, PrincipalType.SERVICE],
```
MarceloRGonc, line 169: Remove PrincipalType.SERVICE

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -136,3 +163,21 @@ const DeleteAppConnectionRequest = {
...
+  },
+  schema: {
+    tags: ['app-connections'],
+    security: [SERVICE_KEY_SECURITY_OPENAPI],
+    description: 'Get an app connection',
+    params: Type.Object({
+      id: OpenOpsId,
+    }),
+    response: {
+      [StatusCodes.NO_CONTENT]: Type.Never(),
```
MarceloRGonc, line 180: [StatusCodes.OK]: AppConnectionWithoutSensitiveData | RedactedAppConnection,

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -136,3 +163,21 @@ const DeleteAppConnectionRequest = {
...
 };
+
+const GetAppConnectionRequest = {
+  config: {
+    allowedPrincipals: [PrincipalType.USER, PrincipalType.SERVICE],
+    permission: Permission.READ_APP_CONNECTION,
+  },
+  schema: {
+    tags: ['app-connections'],
+    security: [SERVICE_KEY_SECURITY_OPENAPI],
```
MarceloRGonc, line 174: Don't need this

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -58,6 +61,30 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
       return appConnectionsWithoutSensitiveData;
     },
   );
+  app.get(
+    '/:id',
+    GetAppConnectionRequest,
+    async (
+      request,
+    ): Promise<AppConnectionWithoutSensitiveData | RedactedAppConnection> => {
```
MarceloRGonc, line 69: The AppConnectionWithoutSensitiveData does not contain the "value" but in some cases of RedactedAppConnection you return the "value" in REDACTED in some properties. Maybe we should also return "value" REDACTED in the other cases

packages/server/api/src/app/app-connection/app-connection.controller.ts
bigfluffycookie: Why should this request be different from the others? all of them have prinicpaltype service and the security thing

MarceloRGonc: We don't use PrincipalType.Service.

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -136,3 +157,28 @@ const DeleteAppConnectionRequest = {
...
+    permission: Permission.READ_APP_CONNECTION,
+  },
+  schema: {
+    tags: ['app-connections'],
+    description: 'Get an app connection',
+    params: Type.Object({
+      id: OpenOpsId,
+    }),
+    response: {
+      [StatusCodes.OK]: Type.Intersect([
```
bigfluffycookie, line 173: dont ask me any questions about this

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -58,6 +60,30 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
...
+    });
+
+    const block = await blockMetadataService.get({
+      name: connection.blockName,
+      projectId: request.principal.projectId,
+      version: undefined,
+    });
+
+    if (!block) {
+      throw new Error(`Block metadata not found for ${connection.blockName}`);
```
MarceloRGonc, line 76: Return bad request or Not found.

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -58,6 +60,30 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
       return appConnectionsWithoutSensitiveData;
     },
   );
+  app.get('/:id', GetAppConnectionRequest, async (request): Promise<any> => {
+    const connection = await appConnectionService.getOneOrThrow({
```
MarceloRGonc, line 64: Return not found.

bigfluffycookie: getOneOrThrow already returns entity not found

MarceloRGonc: The method throws an NotFound exception. Is that the response we have in this endpoint?

bigfluffycookie: Im not sure I understand what you mean. When you give it an id and it cannot find the connection it returns this: {
    "code": "ENTITY_NOT_FOUND",
    "params": {
        "entityType": "AppConnection",
        "entityId": "<id>"
    }
}

MarceloRGonc: In that case the response it's a 404 with that body? Or it's a general 400?

bigfluffycookie: yes its a 404 with that body

MarceloRGonc (approved):

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -58,6 +61,36 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
...
+      });
+
+      const block = await blockMetadataService.get({
+        name: connection.blockName,
+        projectId: request.principal.projectId,
+        version: undefined,
+      });
+
+      if (!block) {
+        logger.error(`Block metadata not found for ${connection.blockName}`);
```
MarceloRGonc, line 80: We don't need to log

PR #542, opened 2025-05-02
Extract builder types and add tests for update-flow-version

alexandrudanpop:
Fixes OPS-1680

### DIFF
--- a/packages/react-ui/src/app/features/builder/builder-types.ts
+++ b/packages/react-ui/src/app/features/builder/builder-types.ts
@@ -1,0 +1,106 @@
+import { BlockProperty } from '@openops/blocks-framework';
+import { AiChatContainerSizeState } from '@openops/components/ui';
+import {
+  Flow,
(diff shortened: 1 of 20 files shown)


packages/react-ui/src/app/features/builder/builder-types.ts
```diff
@@ -0,0 +1,106 @@
...
+import { AiChatContainerSizeState } from '@openops/components/ui';
+import {
+  Flow,
+  FlowOperationRequest,
+  FlowRun,
+  FlowVersion,
+} from '@openops/shared';
+import { DataSelectorSizeState } from './data-selector/data-selector-size-togglers';
+
+export enum LeftSideBarType {
```
alexandrudanpop, line 11: all types moved from builder-hooks & updated other files to reference this. We get circular dependency issue on build if we don't extract out the types

cezudas (approved):

PR #544, opened 2025-05-02
Fix loop results when an iteration fails

MarceloRGonc:
Fixes OPS-1671


packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -188,9 +189,7 @@ async function waitForIterationsToFinishOrPause(
     const { verdict, verdictResponse } = iterationContext;
 
     if (verdict === ExecutionVerdict.FAILED) {
-      return iterationContext.setCurrentPath(
```
rita-gorokhod, line 191: I think there was a reason to do this, no?

rita-gorokhod (approved):

PR #552, opened 2025-05-06
Add service to save step test output

MarceloRGonc:
Fixes OPS-1686


packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    return flowStepTestOutputRepo().save(stepOutput);
+  },
+};
+
+type SaveParams = {
+  outputId?: OpenOpsId | undefined;
+  // TODO: remove optional
+  stepId?: OpenOpsId | undefined;
+  flowVersionId: OpenOpsId;
```
bigfluffycookie, line 44: this should be type FlowVersionId for consistency

bigfluffycookie (commented):
Are you gonna add the actual save endpoint in another pr?

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -0,0 +1,46 @@
...
+  output: unknown;
```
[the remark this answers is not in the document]
MarceloRGonc: We don't know the output type of a block

packages/shared/src/lib/flows/triggers/trigger.ts
```diff
@@ -10,6 +10,7 @@ export enum TriggerType {
 }
 
 const commonProps = {
+  id: Type.Optional(Type.String({})),
```
MarceloRGonc, line 13: It will be mandatory in the future

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -19,6 +19,7 @@ export enum RiskLevel {
 }
 
 const commonActionProps = {
+  id: Type.Optional(Type.String({})),
```
MarceloRGonc, line 22: It will be mandatory in the future

packages/shared/src/lib/flows/triggers/trigger.ts
bigfluffycookie: why Type.String({}) instead of  Type.String() ?

bigfluffycookie (approved):

PR #553, opened 2025-05-06
Add list endpoint to step test output service

bigfluffycookie:
Fixes OPS-1685


packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -110,30 +111,38 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
...
 
-  fastify.get('/:id/test-output', GetFlowTestOutputRequestOptions, (request) =>
-    request.query.stepIds.reduce<Record<string, unknown>>((acc, stepId) => {
-      acc[stepId] = {};
-      return acc;
-    }, {}),
-  );
-};
+  fastify.get(
+    '/:flowVersionId/test-output',
```
rita-gorokhod, line 115: in terms of rest, :id should be for the main id

for everything else we should use query params.
so it's `/` with filters ?flowVersionId=&stepIds=

bigfluffycookie: Disagree, in this same file :flowVersionId is used in post, I think it would be confusing to change it for only this one.

MarceloRGonc: @rita-gorokhod If we change the endpoint to receive the flowVersionId as a parameter, I think we should change all the endpoint.

Right now, it is like this:
- /flow-version/:flowVersionId/test-output?stepIds=

To put the flowVersionId as a parameter, I think we should change it to something like:
-  /flow-step-test-output?flowVersionId=&stepIds=

packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -110,30 +111,38 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
...
+        }),
+        tags: ['flow-step-test-output'],
+        description: 'Gets the test output for a flow version',
+        security: [SERVICE_KEY_SECURITY_OPENAPI],
+        querystring: Type.Object({
+          stepIds: Type.Array(Type.String()),
+        }),
+      },
+    },
+    async (request): Promise<Record<OpenOpsId, unknown>> => {
```
MarceloRGonc, line 132: We need to return the lastUpdated date.

```
Record<OpenOpsId, {"output": unknown, "lastTestDate": string}>
```

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -45,6 +45,46 @@ export const flowStepTestOutputService = {
...
   },
+
+  async list(params: ListParams): Promise<FlowStepTestOutput[]> {
+    const results: FlowStepTestOutput[] = [];
+
+    for (const stepId of params.stepIds) {
+      const flowStepTestOutput = await flowStepTestOutputRepo().findOneBy({
+        flowVersionId: params.flowVersionId,
+        stepId,
+      });
```
MarceloRGonc, line 56: I think we can use only one query to get all the outputs, please try or something similar

```
const flowStepTestOutputs = await flowStepTestOutputRepo().findBy({
  flowVersionId: params.flowVersionId,
  stepId: In(stepIds),
});
```

MarceloRGonc (commented):
Please check with the UI how to answer if a stepId does not have any output yet.

packages/server/api/src/app/flows/flow/flow-version.controller.ts
cezudas: yes, this is needed for the FE.

packages/server/api/src/app/flows/flow/flow-version.controller.ts
rita-gorokhod: `/flow-version/:flowVersionId`
this makes sense, because it's the id of the relevant resource you just requested.
It's very different from /flow-step-test-output/:flowVersionId, which arguably shouldn't have been in this controller in the first place, because it's a different entity.

rita-gorokhod: Ignore, I thought it's `/flow-step-test-output/:flowVersionId`

packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -110,30 +111,38 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
...
-    }),
-  },
+      const outputs = await flowStepTestOutputService.list({
+        stepIds,
+        flowVersionId,
+      });
+
+      return Object.fromEntries(
+        outputs.map(({ stepId, output }) => [stepId, output]),
+      );
```
alexandrudanpop, line 143: We also need the lastTestDate (I guess the `lastUpdated` timestamp from the DB).

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
bigfluffycookie: oh thats cool, didnt know

packages/server/api/src/app/flows/flow/flow-version.controller.ts
bigfluffycookie: done

packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -129,17 +129,26 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
...
       });
-
       return Object.fromEntries(
-        outputs.map(({ stepId, output }) => [stepId, output]),
+        flowStepTestOutputs.map((flowStepTestOutput) => [
+          flowStepTestOutput.stepId as OpenOpsId,
+          {
+            output: flowStepTestOutput.output as unknown,
+            lastTestDate: flowStepTestOutput.updated as string,
+          },
```
cezudas, line 150: Is using type assertion needed here?
`output` and `lastTestDate` are already inferred as strings.
Screenshot 2025-05-08 at 10 30 51

bigfluffycookie: yea youre right, when i added it and didnt update the return type yet it was complaining but its fine now without hah

MarceloRGonc (approved):

PR #558, opened 2025-05-07
Copy over output data from old step output data entries

bigfluffycookie:
Fixes OPS-1689


packages/server/api/test/integration/ce/flows/flow-step-test-output.test.ts
```diff
@@ -22,69 +22,58 @@ afterAll(async () => {
 });
 
 describe('Flow Step Test output', () => {
```
bigfluffycookie, line 24: i extracted common logic so that the individual test cases wouldnt be so long haha

rita-gorokhod (approved):

MarceloRGonc (approved):

PR #559, opened 2025-05-07
Add stepIDs to the existing flow versions

MarceloRGonc:
Fixes OPS-1687.

### DIFF
--- a/packages/server/api/src/app/app.ts
+++ b/packages/server/api/src/app/app.ts
@@ -20,7 +20,6 @@ import {
   FlowImportTemplate,
   FlowRun,
   Folder,
-  isNil,
   Project,
   spreadIfDefined,
   UserInvitation,
@@ -49,8 +48,6 @@ import { flowModule } from './flows/flow.module';
 import { formModule } from './flows/flow/form/form.module';
 import { folderModule } from './flows/folder/folder.module';
 import { triggerEventModule } from './flows/trigger-events/trigger-event.module';
-import { encryptUtils } from './helper/encryption';
-import { jwtUtils } from './helper/jwt-utils';
 import { systemJobsSchedule } from './helper/system-jobs';
 import { organizationModule } from './organization/organization.module';
 import { projectModule } from './project/project-module';
@@ -225,8 +222,6 @@ export const setupApp = async (
     },
   );
 
-  await validateEnvPropsOnStartup();
-
   await app.register(projectModule);
   await app.register(communityBlocksModule);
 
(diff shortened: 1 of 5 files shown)


packages/server/api/src/app/app.ts
```diff
@@ -255,40 +250,6 @@ export const setupApp = async (
...
-  if (isNil(jwtSecret)) {
-    throw new Error(
-      JSON.stringify({
-        message:
-          'OPS_JWT_SECRET is undefined, please define it in the environment variables',
-      }),
-    );
-  }
-};
-
```
MarceloRGonc, line 291: Moved to be able to use it in the migrations

packages/server/api/src/main.ts
```diff
@@ -48,9 +53,47 @@ function setupTimeZone(): void {
...
+  if (isNil(jwtSecret)) {
+    throw new Error(
+      JSON.stringify({
+        message:
+          'OPS_JWT_SECRET is undefined, please define it in the environment variables',
+      }),
+    );
+  }
+}
+
```
MarceloRGonc, line 90: Moved

rita-gorokhod: why do you need it? how is it related to the migration?

packages/server/api/src/app/database/migrations/1746618034190-AddStepIdToExistingFlowVersions.ts
```diff
@@ -0,0 +1,114 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { fileCompressor, logger } from '@openops/server-shared';
+import { FileCompression, openOpsId } from '@openops/shared';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { encryptUtils } from '../../helper/encryption';
+
+export class AddStepIdToExistingFlowVersions1746454781866
```
bigfluffycookie, line 7: I would rename it, its confusing, at first I thought we were updating table 'flow_version' based on the name, maybe 'addStepIdToExistingFlowVersionsOfTestOutputTable' or something along those lines

packages/server/api/src/main.ts
bigfluffycookie: +1, i dont see how youre using it, do you mean it needs to be initialized at an earlier point?

MarceloRGonc: We are encrypted data. But in the migrations, we don't have the value initialized, yet. So, I had to move it to an earlier state

packages/server/api/src/app/database/migrations/1746618034190-AddStepIdToExistingFlowVersions.ts
MarceloRGonc: Renamed

bigfluffycookie (approved):

PR #560, opened 2025-05-07
Add stepId for the existing templates

MarceloRGonc:
Part of OPS-1687.

### DIFF
--- a/packages/server/api/src/app/database/migrations/1746618034190-AddStepIdToExistingFlowVersions.ts
+++ b/packages/server/api/src/app/database/migrations/1746618034190-AddStepIdToExistingFlowVersions.ts
@@ -1,0 +1,116 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { fileCompressor, logger } from '@openops/server-shared';
+import { FileCompression, openOpsId } from '@openops/shared';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+import { encryptUtils } from '../../helper/encryption';
+
+export class AddStepIdToExistingFlowVersions1746454781866
+  implements MigrationInterface
+{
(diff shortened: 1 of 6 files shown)


packages/server/api/src/app/database/migrations/1746627735717-AddStepIdToExistingTemplates.ts
```diff
@@ -0,0 +1,61 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { logger } from '@openops/server-shared';
+import { openOpsId } from '@openops/shared';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddStepIdToExistingTemplates1746627735717
```
rita-gorokhod, line 6: why do we need this as part of the migration?
the FE should generate the ids when using a template

packages/server/api/src/app/database/migrations/1746618034190-AddStepIdToExistingFlowVersions.ts
```diff
@@ -0,0 +1,116 @@
...
+    const jsonData = record.trigger ?? record.template;
+
+    const { updatedJson, extractedData } = await processJsonObject(jsonData);
+
+    await queryRunner.query(
+      `UPDATE "${tableName}" SET "trigger" = $1 WHERE "id" = $2`,
+      [updatedJson, record.id],
+    );
+
+    logger.info(`LENGTH ${extractedData.length}`);
```
bigfluffycookie, line 42: remove

rita-gorokhod (changes requested):
@MarceloRGonc as we discussed offline, we don't need to do it. The templates need to be compatible with older versions that don't have this property and they're versioned. We can create new versions of templates with it but I don't see the need, the FE will generate the ids. I think we can close this.

PR #564, opened 2025-05-08
Use new test output data API to fetch data for the Test Flow Widget

cezudas:
Fixes OPS-1710.


packages/react-ui/src/app/features/builder/test-step/step-test-output-hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        const stepTestOutput = await flowsApi.getStepTestOutput(
+          flowVersionId,
+          stepId,
+        );
+
+        if (isNil(stepTestOutput) || isEmpty(stepTestOutput)) {
+          return resolveFallbackData();
+        }
```
alexandrudanpop, line 39: will need to cleanup this fallback data logic once the feature is implemented

packages/react-ui/src/app/features/builder/test-step/step-test-output-hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+    });
+  },
+  useStepTestOutputFormData(
+    flowVersionId: string,
+    form: UseFormReturn<Action> | UseFormReturn<Trigger>,
+  ) {
+    const fallbackData = useWatch({
+      name: 'settings.inputUiInfo.currentSelectedData',
+      control: form.control as any,
+    }) as unknown;
```
alexandrudanpop, line 52: Can't we just pass a `getFallbackData` to the hook so we don't have the form dependency here?

packages/react-ui/src/app/features/builder/test-step/step-test-output-hooks.ts
cezudas: Agree. Do you suggest adding a TODO now to remove the "fallback" logic later?

alexandrudanpop (approved):

PR #568, opened 2025-05-08
Add strict to cli prompts

bigfluffycookie:
Fixes OPS-1714

### DIFF
--- a/ai-prompts/aws-cli.txt
+++ b/ai-prompts/aws-cli.txt
@@ -4,10 +4,12 @@ Your only task is to convert user requests into exact CLI commands that will run
 
 Guidelines:
 
+STRICT Never do the following things: 
+  - You may not use pipes (|), redirects (>), or interact with the file system. All CLI commands must be able to execute in isolation.
+
 Output format
   - Return CLI commands as formatted code Markdown (example ```bash command ```), add explanations only if the command is complex, but keep it short and clear.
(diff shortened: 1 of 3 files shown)


rita-gorokhod (approved):

ai-prompts/azure-cli.txt
```diff
@@ -33,6 +35,7 @@ Best practices
 
 Placeholders
   - Always use descriptive placeholders like `<project-id>`, `<region>`, `<vm-name>`, etc.
+  - If using placeholders, state: `Note: This command uses placeholders which are signaled with <>`
```
bigfluffycookie, line 38: I thought about using 'PLACEHOLDER' but i really dont think its better than using <>, infact I think its more confusing, I think it makes more sense to stick with common practice and instead opt to explain it. WDYT? @rita-gorokhod

ai-prompts/aws-cli.txt
```diff
@@ -35,6 +37,7 @@ Best practices
 
 Placeholders
```
rita-gorokhod, line 38: https://github.com/openops-cloud/openops/pull/571/files#diff-07be703f304e825716966bb3626f9eb4b48aad8b3f2984fe6acf269e28aef569R61

as I wrote on slack, let's use PARAM_NAME instead of <param_name>, in caps lock and without <>

rita-gorokhod (commented):
See comment about the placeholders

ai-prompts/aws-cli.txt
bigfluffycookie: ok

PR #588, opened 2025-05-14
Add patch connection endpoint

bigfluffycookie:
Fixes OPS-1706


packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
```diff
@@ -94,6 +98,66 @@ export const appConnectionService = {
...
+    await repo().update(connection.id, connection);
+
+    const updatedConnection = await repo().findOneByOrFail({
+      name: request.name,
+      projectId,
+    });
+
+    sendConnectionUpdatedEvent(params.userId, projectId, request.blockName);
+
+    return decryptConnection(updatedConnection);
```
rita-gorokhod, line 158: did you mean to encrypt here? why return the decrypted?

bigfluffycookie: Hmm no, im doing the same as in the "save" function, I guess we are just returning the saved / updated connection as is, the encryption is only important for the db. I suppose it is a bit confusing here bc i decrypt, encrypt and then decrypt again

packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
```diff
@@ -94,6 +98,66 @@ export const appConnectionService = {
...
+    });
+
+    const block = await blockMetadataService.get({
+      name: request.blockName,
+      projectId,
+      version: undefined,
+    });
+
+    if (!block) {
+      throw new Error(`Block metadata not found for ${request.blockName}`);
```
MarceloRGonc, line 116: This should originate a Bad Request response

packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
MarceloRGonc: Shouldn't we return the values redacted? I think you are redacting the values in the controller.

However, I also don't think you need to decrypt; you already have the values decrypted.

packages/server/api/src/app/app-connection/app-connection-service/app-connection-service.ts
```diff
@@ -331,61 +391,6 @@ function decryptConnection(
   return connection;
 }
 
-const engineValidateAuth = async (
```
bigfluffycookie, line 334: moved it in a seperate file, it was impossible to test else.

MarceloRGonc (approved):

PR #589, opened 2025-05-14
Support Slack actions via URLs

rita-gorokhod:
Fixes OPS-1727


packages/server/api/src/app/slack/slack-interaction-module.ts
```diff
@@ -115,7 +123,7 @@ async function evaluateUserInteraction(payload: any, reply: FastifyReply) {
     const userId = payload.user.id;
     const responseUrl = payload.response_url;
     const ephemeralText =
-      'Slack interactions are only available when running the entire workflow.';
+      'Test succeeded. Slack interactions are disabled in test mode and are only available when running the entire workflow.';
```
rita-gorokhod, line 126: image

packages/server/api/src/app/slack/slack-interaction-module.ts
```diff
@@ -90,6 +90,14 @@ const slackInteractionController: FastifyPluginCallbackTypebox = (
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any, @typescript-eslint/explicit-function-return-type
 async function evaluateUserInteraction(payload: any, reply: FastifyReply) {
+  const interactionsEnabled: boolean =
```
rita-gorokhod, line 93: I will open a PR to do the same on the lambda

packages/blocks/slack/src/lib/actions/request-action-message.ts
```diff
@@ -106,6 +104,27 @@ const sendMessageAskingForAction = async (
...
 
+  const useSlackInteractions =
+    system.getBoolean(SharedSystemProp.USE_SLACK_INTERACTIONS) || false;
+
+  if (!useSlackInteractions) {
+    const baseUrl = await networkUtls.getPublicUrl();
+
+    actions.forEach((action: SlackActionDefinition) => {
+      action.url = context.run.isTest
+        ? 'https://static.openops.com/test_slack_interactions.txt'
```
rita-gorokhod, line 115: image

packages/blocks/slack/src/lib/common/message-interactions.ts
```diff
@@ -26,13 +28,17 @@ export function removeActionBlocks(blocks: any): any[] {
...
+    : `Clicked on '${button}'`;
+
   const modifiedBlocks = [
     { type: 'divider' },
     {
       type: 'section',
       text: {
         type: 'mrkdwn',
-        text: `*:white_check_mark: Action received, user @${user} clicked on '${button}'*`,
+        text: `*:white_check_mark: Action received: ${whoClickedText}*`,
```
rita-gorokhod, line 41: image

packages/blocks/slack/src/lib/actions/request-action-message.ts
```diff
@@ -106,6 +104,27 @@ const sendMessageAskingForAction = async (
     conversationId,
   );
 
+  const disableSlackInteractions =
+    system.getBoolean(SharedSystemProp.DISABLE_SLACK_INTERACTIONS) || false;
```
rita-gorokhod, line 108: for those that already have an old installation, they won't have it in their .env file, so by default I'm keeping the old behavior

maor-rozenfeld: A very confusing variable name.

We now have two slack actions mode - interactions and URL. So we can just call this variable SLACK_ACTIONS_MODE which can be either URL or interactions.

If it must stay as a boolean, it should be `ENABLE_SLACK_INTERACTION_MODE` or `ENABLE_SLACK_ACTIONS_BY_URL`, definitely not a "DISABLE" variable.

deploy/docker-compose/.env.defaults
```diff
@@ -112,3 +112,4 @@ HOST_CLOUDSDK_CONFIG=
 # ---------------------------------------------------------
 
 OPS_CODE_BLOCK_MEMORY_LIMIT_IN_MB=256
+DISABLE_SLACK_INTERACTIONS=false
```
maor-rozenfeld, line 115: You forgot the OPS prefix. It can join the slack connection variable a few lines above.

PR #613, opened 2025-05-15
Add feature flag to conditionally load Superset and Tables mcp tools

cezudas:
Part of OPS-1760.

Local testing procedure: I logged `loadTablesAndSupersetMcpTools` to the console to ensure the system prop loads correctly:
1. Without `.env` file changes it prints `false` -> defaults to `false`
2. With `OPS_LOAD_TABLES_AND_SUPERSET_MCP_TOOLS=true` in `.env` it prints `true`


packages/server/api/src/app/ai/mcp/mcp-tools.ts
```diff
@@ -9,10 +9,14 @@ export const getMCPTools = async (): Promise<ToolSet> => {
   const docsTools = await safeGetTools('docs', getDocsTools);
```
rita-gorokhod, line 9: you shouldn't make the call if the flag is disabled

rita-gorokhod (approved):

PR #621, opened 2025-05-15
Use a commit hash for superset-mcp

MarceloRGonc:
Part of OPS-1741.


rita-gorokhod (approved):
lgtm but I'd call it superset mcp, to avoid mixing up with superset

PR #630, opened 2025-05-16
Add strategy to summarize ai message history

MarceloRGonc:
Fixes OPS-1787.

## Additional notes:
We are only summarizing after an error.

### We are summarizing the chat in the following situations:
- In selectRelevantTools when there is an error related tokens, we ask for a summary and we retry the request
- On the MCP controller
  - In the event of a general error, we request a summary in the background. And we append to the error message 'Please try again'
  - On an error streaming to the user if the error is related to tokens, we ask for a summary, and we retry the request
  - Upon finishing streaming, if the finish reason is 'length', we request a summary. And we append to the message 'The message was truncated because the maximum tokens for the context window was reached. Please try again.'
  
### Summarize strategy
- We ask the LLM to do the summary.
- If the last message is from the user, we remove it, and we append the message to the end
- If the request to the LLM fails, we truncate the chat history based on user interactions. I added an environment variable


packages/openops/src/lib/ai/providers/openai.ts
```diff
@@ -50,6 +50,7 @@ function createLanguageModel(params: {
 }): LanguageModelV1 {
   return createOpenAI({
     apiKey: params.apiKey,
+    compatibility: 'strict',
```
MarceloRGonc, line 53: It's needed for OpenAI to return the token usage.

packages/server/shared/src/lib/system/system.ts
```diff
@@ -88,6 +88,9 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
   [AppSystemProp.SUPERSET_MCP_SERVER_PATH]: '/root/.mcp/superset',
   [AppSystemProp.DOCS_MCP_SERVER_PATH]: '/root/.mcp/docs.openops.com',
   [AppSystemProp.LOAD_TABLES_AND_SUPERSET_MCP_TOOLS]: 'false',
+  [AppSystemProp.MAX_LLM_CALLS_WITHOUT_INTERACTION]: '10',
+  [AppSystemProp.MAX_TOKENS_IN_LLM_HISTORY]: '15000',
+  [AppSystemProp.MAX_MESSAGES_IN_LLM_HISTORY]: '15',
```
MarceloRGonc, line 93: We need to decide on these values

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export async function summarizeMessages(
+  languageModel: LanguageModel,
+  aiConfig: AiConfig,
+  messages: CoreMessage[],
+  totalTokens: number,
+): Promise<CoreMessage[]> {
+  const maxTokens = getHistoryMaxTokens(aiConfig);
+  let debugMessage = `Token count ${totalTokens}, which exceeds the configured limit of ${maxTokens}. Summarizing messages.`;
```
bigfluffycookie, line 47: why is the logger.info(debugMessage) so far away from this logic? is there a reason it should only be logged after the if else?

packages/server/api/src/app/ai/chat/ai-chat.service.ts
```diff
@@ -87,6 +95,73 @@ export const saveChatHistory = async (
...
+    timeout: 10000,
+  });
+
+  try {
+    const existingMessages = await getChatHistory(chatId);
+
+    existingMessages.push(...messages);
+
+    await saveChatHistory(chatId, existingMessages);
+  } finally {
```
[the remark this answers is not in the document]
MarceloRGonc: We don't need it. I only have the try/finally to do the release of the lock. If we get an error we can throw it.

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -255,63 +263,3 @@ function endStreamWithErrorMessage(
     `d:{"finishReason":"stop","usage":{"promptTokens":null,"completionTokens":null}}\n`,
   );
 }
-
-function removeToolMessages(messages: CoreMessage[]): CoreMessage[] {
```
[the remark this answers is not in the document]
MarceloRGonc: Now, we save all the messages and resume the history when needed. So we save all the history for the summary to have all the context.

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
MarceloRGonc: I only want to log this message if we are summarising the history. The message varies according to the different strategies.

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  if (isNaN(totalTokens)) {
+    debugMessage = `Message count is ${messages.length}, which exceeds the configured limit of ${historyMaxMessages}. Summarizing messages.`;
+    logger.debug(
+      `The model is not providing token usage. Checking if the number of messages exceeds ${historyMaxMessages}.`,
+    );
+
+    if (messages.length < historyMaxMessages) {
```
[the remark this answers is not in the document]
MarceloRGonc: Can't merge them. The only thing I can do is to split this in two if...else blocks.

bigfluffycookie (approved):

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  return system.getNumberOrThrow(AppSystemProp.MAX_TOKENS_FOR_HISTORY_SUMMARY);
+}
+
+export function shouldTryToSummarize(
+  message: string,
+  attemptIndex: number,
+): boolean {
+  return message.includes('tokens') && attemptIndex < 2;
```
MarceloRGonc, line 19: Not sure if the max attempts should be an environment variable

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,140 @@
...
+        aiConfig,
+        attemptIndex,
+      );
+    }
+
+    throw error;
+  }
+}
+
+function truncateByTheNumberOfUserInteractions(
```
MarceloRGonc, line 116: If we can't summarize using the normal flow, we will summarize based on user interactions. I added an environment variable with a default value of 2 interactions.

packages/server/api/src/app/ai/chat/tools.service.ts
```diff
@@ -64,7 +72,28 @@ export async function selectRelevantTools({
...
+
+    logger.debug('Retry the call to selectRelevantTools');
+    return selectRelevantTools({
+      chatId,
+      messages: newHistory,
+      languageModel,
+      aiConfig,
+      tools,
+      attemptIndex,
+    });
```
MarceloRGonc, line 97: I added this here because it can happen when requesting tools. But I think that to request the tools to use we should only use the user's last message.

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -101,18 +122,23 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
...
 
-    const messages = await getChatHistory(chatId);
-    messages.push({
+    const newMessage: CoreMessage = {
       role: 'user',
       content: request.body.message,
-    });
+    };
+
+    await appendMessagesToChatHistory(chatId, [newMessage]);
```
rita-gorokhod, line 130: why is this called twice?

MarceloRGonc: It's two different histories. One is the full history with all the messages to be render in the UI. The other is the one that we summarise.

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -132,38 +158,50 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
       isTablesLoaded,
     });
 
+    const streamMessagesParams: StreamMessagesParams = {
```
rita-gorokhod: extract everything please from this point and the helper functions streamMessages. it's very difficult to review, especially with the error handling. There shouldn't be this much logic in a controller.

packages/server/shared/src/lib/system/system.ts
```diff
@@ -88,6 +88,9 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
   [AppSystemProp.SUPERSET_MCP_SERVER_PATH]: '/root/.mcp/superset',
   [AppSystemProp.DOCS_MCP_SERVER_PATH]: '/root/.mcp/docs.openops.com',
   [AppSystemProp.LOAD_EXPERIMENTAL_MCP_TOOLS]: 'false',
+  [AppSystemProp.MAX_LLM_CALLS_WITHOUT_INTERACTION]: '10',
+  [AppSystemProp.MAX_TOKENS_FOR_HISTORY_SUMMARY]: '2000',
```
rita-gorokhod, line 95: How does it work with our huge prompt and small models?

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,140 @@
...
+export async function summarizeChatHistoryContext(
+  languageModel: LanguageModel,
+  aiConfig: AiConfig,
+  chatId: string,
+): Promise<CoreMessage[]> {
+  return appendMessagesToChatHistoryContext(
+    chatId,
+    [],
+    async (existingMessages) => {
+      let reAdd = false;
```
rita-gorokhod: I think removing and readding is hard to read.

```
async function summarizeChatHistory(messages) {
  const lastMessage = messages[messages.length - 1];
  const isLastMessageFromUser = lastMessage && lastMessage.role === 'user' && messages.length > 1;
  const messagesForSummary = isLastMessageFromUser ? messages.slice(0, -1) : messages;

  const summarizedHistory = await requestToGenerateSummary(languageModel, messagesForSummary, aiConfig);
  return isLastMessageFromUser ? [...summarizedHistory, lastMessage] : summarizedHistory;
}
```
wdyt?

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,140 @@
...
+      'Focus only on the most important information, decisions, and context. ' +
+      'Omit pleasantries, redundant information, and unnecessary details. ' +
+      'Your summary should be as brief as possible while preserving all critical information.';
+
+    const { text } = await generateText({
+      messages,
+      model: languageModel,
+      system: systemPrompt,
+      ...aiConfig.modelSettings,
+      maxTokens: getHistoryMaxTokens(aiConfig),
```
rita-gorokhod: how does the maxTokens work in this case?

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    return [
+      {
+        role: 'system',
+        content: `The following is a summary of the previous conversation: ${text}`,
+      },
+    ];
+  } catch (error) {
+    if (attemptIndex < 1 && APICallError.isInstance(error)) {
```
rita-gorokhod, line 93: explain this logic please, why do we need it?

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    },
+  );
+}
+
+async function requestToGenerateSummary(
+  languageModel: LanguageModel,
+  messages: CoreMessage[],
+  aiConfig: AiConfig,
+  attemptIndex = 0,
```
rita-gorokhod: I think the recursion here is confusing with the attemptIndex . 
Why not instead call an inner func? 
this method is supposed to be called only twice, no? for full input, and for last 2 interactions -- if those fail we don't try anymore

packages/server/api/src/app/ai/chat/tools.service.ts
```diff
@@ -64,7 +72,28 @@ export async function selectRelevantTools({
         .slice(0, MAX_SELECTED_TOOLS),
     );
   } catch (error) {
-    logger.error('Error selecting tools', error);
-    return;
+    const errorMessage = error instanceof Error ? error.message : String(error);
+
+    attemptIndex = attemptIndex + 1;
+    if (!shouldTryToSummarize(errorMessage, attemptIndex)) {
```
rita-gorokhod, line 78: I don't think this should be here, this gives you the tools selection, it should already receive a summarized data if summarization was needed.
All the error handling and summarization should be done on the caller level.

packages/server/api/src/app/ai/chat/tools.service.ts
rita-gorokhod: I think that the last message could maybe be partial, so we might need more than just 1

packages/server/shared/src/lib/system/system.ts
MarceloRGonc: I don't think there are any models with less than ~4000 tokens. So, I'm defining half of that as default. But the prompt is not included in the summary.

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
MarceloRGonc: WDYM? The maxTokens is the maximum number of tokens to generate. So it will be the value from the env var (2000 by default) or the value defined in the model settings.

packages/server/api/src/app/ai/chat/tools.service.ts
MarceloRGonc: We are calling the LLM to select the tools and we are using the entire history. It may be necessary to summarize since we are only summarizing on error.

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
MarceloRGonc: I think these retries (2) are enough. However, I'm not sure if you want to configure it. Can I always assume 2 retries?

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
rita-gorokhod: yes but does it mean that the history will be max 2000 tokens? what if it needs more?

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
rita-gorokhod: even if it's configurable, you can do it without recursion.

packages/server/api/src/app/ai/chat/tools.service.ts
rita-gorokhod: Sure but then you can do it in the caller.
Call tools with everything --> failed on max? --> summarize --> call tools with summarized

otherwise this logic is spread in multiple places

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
MarceloRGonc: If it needs more, the behaviour is the same as if you reach the maximum tokens of a context window, it will be truncated.

packages/server/api/src/app/ai/chat/ai-message-history-summarizer.ts
MarceloRGonc: I removed the attemptIndex. However, it now only allows two retries without configuration.
- Full input
- Last 2 interactions

I don't know how you want the two retries to be configurable

packages/server/api/src/app/ai/chat/tools.service.ts
MarceloRGonc: What do you mean by caller? This is called in the'ai-mcp-chat.controller.ts'. The other shouldTryToSummarize is called in the 'ai-stream-handler.ts'

Please don't look at the old code, as I have already made some changes, as you requested.

PR #633, opened 2025-05-19
Use only list MCP tools for table

alexandrudanpop:
Fixes OPS-1769


ai-prompts/mcp-tables.txt
```diff
@@ -7,10 +7,11 @@ If needed, handle invoking different OpenOps MCP (model context protocol) tools
...
 When invoking the OpenOps Tables (Baserow) MCP, you must fully understand the Baserow REST API and its core functionality. Specifically:
-- Understand how to create add/update/delete records, query existing data.
+- Understand how to query existing data.
 - **Strict** Do not manage table schema (e.g., adding, removing and changing fields).
 - Always ensure:
   - Table IDs, row IDs, and field IDs are correctly obtained through API calls and not assumed.
   - Names, field types, or constraints match Baserow's schema or configuration exactly.
   - Token-based authentication is used for all API interactions.
-- Format payloads appropriately for each endpoint, adhering to Baserow's API requirements (e.g., the types and structure of fields and rows).
+  - If asked to provide specific information, when using the list_rows_table tools, provide the appropriete search filter.
```
rita-gorokhod, line 16: Minor typo: appropriate

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -104,8 +104,18 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
     });
 
     const tools = await getMCPTools();
-    const isAnalyticsLoaded = Object.keys(tools).includes('superset');
-    const isTablesLoaded = Object.keys(tools).includes('table');
+    logger.debug(
```
rita-gorokhod, line 107: We can remove the debug before merging, since this is called every time it's very spammy

rita-gorokhod: Same for the other debug logs here

ai-prompts/mcp-tables.txt
alexandrudanpop: done

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
alexandrudanpop: done

packages/server/api/src/app/ai/mcp/tables-tools.ts
```diff
@@ -26,5 +36,70 @@ export async function getTablesTools(): Promise<ToolSet> {
...
+
+  // List Tables Tool
+  const listTablesTool = tools['list_tables'];
+  if (listTablesTool) {
+    toolSet.listTables = listTablesTool;
+  }
+
+  // List Rows Tool
+  if (registry.list) {
+    toolSet.listRowsTable = tool({
```
rita-gorokhod, line 72: @alexandrudanpop I'm a bit confused about this.
Why do we need this?
Why not return the existing list of tools, as-is, and just filter out anything that doesn't start with "list"? 
I'm concerned about this modification, if we pull from baserow and something changes, this will break.

alexandrudanpop: if we have a customer if a lot of tables, and keep adding our own MCP tools, we will eventually reach the tools limit
```
3:"Invalid 'tools': array too long. Expected an array with maximum length 128, but got an array with length 179 instead

```

alexandrudanpop: for exampel OpenAI have 128 tool limitation https://platform.openai.com/docs/assistants/deep-dive/creating-assistants#creating-assistants

alexandrudanpop: related to baserow pulls, we need to test if there are API changes, I hope they would version breaking changes

rita-gorokhod: > if we have a customer if a lot of tables, and keep adding our own MCP tools, we will eventually reach the tools limit

It's a valid concern but I think we can live with this for now, it would happen only if the user has more than 100 tables. And:

> related to baserow pulls, we need to test if there are API changes, I hope they would version breaking changes

We can't rely on this, so it's best not to touch their tools. Or if we do, then this modification needs to be on the tables repo.

So I propose to not do this here anyway, and once we add more mcp tools, we should do this on tables repo (or hopefully baserow will do it by then)

alexandrudanpop: ok, as discussed, we are not returning only the list tools

rita-gorokhod (approved):

PR #639, opened 2025-05-19
Add fallback to reading ai prompts from local

rita-gorokhod:
Fixes OPS-1802


.github/workflows/release.yml
```diff
@@ -82,9 +82,12 @@ jobs:
       RELEASE_DC_FILENAME: openops-dc-${{ needs.get-version.outputs.version }}.zip
     steps:
       - uses: actions/checkout@v4
```
rita-gorokhod, line 84: @maor-rozenfeld please review, I want the files to be available at this location in the installation:
https://github.com/openops-cloud/openops/blob/c093d7cd6048d54fb5db0ae033a1e9b585eae0a7/packages/server/api/src/app/ai/chat/prompts.service.ts#L63

maor-rozenfeld: You need to add a volume to docker-compose

bigfluffycookie (approved):

PR #645, opened 2025-05-20
Add One-Time notifications table to seed

MarceloRGonc:
Part of OPS-1812.


rita-gorokhod (changes requested):
As discussed offline, I don't think we should add new tables to the seed. It will confuse the user why to use this and not the opportunities table.

PR #647, opened 2025-05-20
Fix end workflow action failure

bigfluffycookie:
Fixes OPS-1791


packages/server/api/src/app/flows/step-run/step-run-service.ts
```diff
@@ -48,7 +48,7 @@ export const stepRunService = {
       projectId,
     });
 
-    if (step.id) {
+    if (step.id && result.output) {
       await flowStepTestOutputService.save({
```
rita-gorokhod, line 52: will it be a problem that we don't save it for this step? what is going to happen when you run a workflow that has it, and try to view it in runs pane?

bigfluffycookie: ok so apparently the run workflow panel takes the output from flow-run, not test-output, now for the output of test step it will just be a {} if it cannot find it in the DB

packages/blocks/end-flow/src/lib/actions/end-workflow.ts
```diff
@@ -14,6 +14,6 @@ export const endFlowAction = createAction({
     },
   },
   async run(ctx) {
-    return ctx.run.stop({ response: {} });
+    return ctx.run.stop({ response: {} }) ?? true;
```
bigfluffycookie, line 17: @rita-gorokhod like this?

rita-gorokhod: @bigfluffycookie seems like .stop is a void func, so you can do
```
ctx.run.stop({ response: {} });
return true;
```

packages/server/api/src/app/flows/step-run/step-run-service.ts
```diff
@@ -52,7 +52,7 @@ export const stepRunService = {
...
         stepId: step.id,
         flowVersionId: flowVersion.id,
-        output: result.output,
+        output: isNil(result.output) ? {} : result.output,
```
rita-gorokhod, line 55: if we don't need this I'd revert, it's a risky area

rita-gorokhod (approved):
LGTM see comments

PR #649, opened 2025-05-20
Fix inclusion of prompt for tables mcp tools

MarceloRGonc:
Part of OPS-1811

### DIFF
--- a/packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
+++ b/packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
@@ -112,7 +112,7 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
       key.includes('superset'),
     );
     const isTablesLoaded = Object.keys(tools).some((key) =>
-      key.includes('Table'),
+      key.includes('table'),
     );
 
     const systemPrompt = await getMcpSystemPrompt({


packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -112,7 +112,7 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
       key.includes('superset'),
     );
     const isTablesLoaded = Object.keys(tools).some((key) =>
-      key.includes('Table'),
+      key.includes('table'),
```
rita-gorokhod, line 115: @alexandrudanpop what were the values that you saw?

MarceloRGonc: For me it's always '...table...' but maybe we can do lowercase to be safe.

alexandrudanpop: yes, now we receive them as `table` but I think we should modify the keys so it doesn't conflict with superset. I think this issue was because of my previous change where I was wrapping each into one single tool and changed the name to table.
image

alexandrudanpop (approved):

rita-gorokhod (approved):

PR #652, opened 2025-05-20
Allow modifying a connection

cezudas:
Fixes OPS-1563.

Demo: https://www.loom.com/share/5f9556cd85474df5b3a94b78307b34da
I've [triggered](https://github.com/openops-cloud/devops/actions/runs/15145178692) a deployment to UX, UX env can be used for a quick test.

Rita edit: it's on Test now


packages/react-ui/src/app/features/connections/components/create-edit-connection-dialog-content.tsx
```diff
@@ -127,17 +130,33 @@ const CreateEditConnectionDialogContent = ({
...
+        if (
+          !isNil(existingConnection) &&
+          editConnection?.id !== existingConnection.id
+        ) {
           throw new ConnectionNameAlreadyExists();
         }
       }
-      return appConnectionsApi.upsert(formValues);
+      if (editConnection) {
+        return appConnectionsApi.patch(formValues);
```
cezudas, line 141: for edit scenario (including Reconnect from the connection picker) using patch because the BE will read the already persisted secrets when the FE passes value `**REDACTED**` . This is precisely how they come from the getConnectionById request, so both scenarios are covered:
1) if users changes the already saved secrets BE will use the new values to update in DB
2) FE passes `**REDACTED**`  for every secret that wasn't changed.

alexandrudanpop:
I can't edit a slack connection, the save button is disabled:
image

alexandrudanpop:
We should not allow renaming connections as it will break workflows (similar to reconnect behavior)
image

packages/react-ui/src/app/features/connections/components/connection-table.tsx
```diff
@@ -79,27 +84,62 @@ const DeleteConnectionColumn = ({
...
+            }}
+          >
+            <DeleteConnectionDialog
+              connectionName={row.name}
+              mutationFn={() =>
+                appConnectionsApi.delete(row.id).then((data) => {
+                  setRefresh((prev) => !prev);
+                  return data;
+                })
+              }
```
alexandrudanpop, line 122: would be good to extract the functions out of the JSX

packages/react-ui/src/app/features/connections/components/edit-connection-dialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+}: Readonly<EditConnectionDialogProps>) {
+  const { data: editConnection } = appConnectionsHooks.useConnection({
+    id,
+  });
+
+  const { blocks } = blocksHooks.useBlocks({});
+
+  const block = blocks?.find(
+    (block) => block.name === editConnection?.blockName,
+  );
```
alexandrudanpop, line 24: nitpick, could wrap with useMemo

packages/react-ui/src/app/features/connections/lib/connections-utils.ts
```diff
@@ -96,58 +101,84 @@ export const buildConnectionSchema = (
...
-          props: {},
-          code: '',
-        },
+        value: existingConnection
+          ? ({
+              ...existingConnection.value,
+              scope: block.auth?.scope.join(' '),
+              client_id: '',
+              props: {},
+              code: '',
```
alexandrudanpop, line 166: not sure if this is correct, we should pick the existing values from the existingConnection

cezudas: For some reason regular "Slack" connection doesn't contain those. Anywho I changed the code so if they are on the connection they will be used.
Screenshot 2025-05-21 at 13 32 01

packages/react-ui/src/app/features/connections/components/create-edit-connection-dialog-content.tsx
```diff
@@ -63,16 +64,18 @@ class ConnectionNameAlreadyExists extends Error {
 
 export type CreateEditConnectionDialogContentProps = {
   block: BlockMetadataModelSummary | BlockMetadataModel;
-  onConnectionCreated: (name: string) => void;
-  reconnectConnection: AppConnectionWithoutSensitiveData | null;
+  onConnectionSaved: (name: string) => void;
+  editConnection: AppConnection | null;
```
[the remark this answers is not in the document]
cezudas: done

packages/react-ui/src/app/features/connections/components/connection-table.tsx
cezudas: done

cezudas:
> I think we should keep the 3-dot context-menu style consistent (including icons) Ex WF page: image
> 
> Connections: image
> 
> FYI @rita-gorokhod

will be a follow ticket

packages/react-ui/src/app/features/connections/components/edit-connection-dialog.tsx
cezudas: I would like to do it in a follow-up ticket, as I need to re-test after every minor refactor

packages/react-ui/src/app/features/connections/components/create-edit-connection-dialog-content.tsx
```diff
@@ -208,7 +238,8 @@ const CreateEditConnectionDialogContent = ({
                   <FormLabel htmlFor="name">{t('Connection Name')}</FormLabel>
                   <FormControl>
                     <Input
-                      disabled={!isNil(reconnectConnection)}
+                      // TODO: OPS-XXX - When renaming a connection, it breaks all workflows that used it.
```
alexandrudanpop, line 241: you can now delete this todo

alexandrudanpop (approved):

PR #656, opened 2025-05-21
Fix trigger results not being added to test step output

bigfluffycookie:
Fixes OPS-1747


packages/server/api/src/app/flows/trigger-events/trigger-event.service.ts
```diff
@@ -45,13 +46,23 @@ export const triggerEventService = {
...
       sourceName,
       payload: sanitizeObjectForPostgresql(payload),
-    });
+    };
+
+    if (flow.version.trigger.id) {
+      await flowStepTestOutputService.save({
+        stepId: flow.version.trigger.id,
+        flowVersionId: flow.version.id,
+        output: triggerEventToSave.payload,
```
MarceloRGonc, line 61: Have you been able to check if this payload always contains data? For example, a webhook trigger with no parameters.

bigfluffycookie: im not sure what you mean with "a webhook trigger with no parameters", but ive tried with all three triggers we have and there is no issue

MarceloRGonc (approved):

PR #676, opened 2025-05-23
Move in memory flags to new file

MarceloRGonc:
Part of OPS-1711.


packages/server/api/src/app/flags/flag.service.ts
```diff
@@ -31,270 +23,10 @@ export const flagService = {
   },
   async getAll(): Promise<Flag[]> {
     const flags = await flagRepo().find({});
-    const now = new Date().toISOString();
```
MarceloRGonc, line 34: Moved

packages/server/api/src/app/flags/in-memory-flags.service.ts
```diff
@@ -0,0 +1,297 @@
...
+      created,
+      updated,
+    },
+    {
+      id: FlagId.USE_NEW_EXTERNAL_TESTDATA,
+      value: false,
+      created,
+      updated,
+    },
+  ];
```
MarceloRGonc, line 289: Same values as before

rita-gorokhod (approved):

PR #683, opened 2025-05-23
Scroll to the end of the AI chat on first opening 

rSnapkoOpenOps:
Fixes OPS-1778.

## Testing Checklist

Check all that apply:

### DIFF
--- a/packages/ui-components/src/components/ai-chat-container/ai-assistant-chat-container.tsx
+++ b/packages/ui-components/src/components/ai-chat-container/ai-assistant-chat-container.tsx
@@ -1,7 +1,7 @@
 import { UseChatHelpers } from '@ai-sdk/react';
 import { t } from 'i18next';
 import { Bot, Send as SendIcon } from 'lucide-react';
-import { ReactNode, useRef } from 'react';
+import { ReactNode, useEffect, useRef } from 'react';
 import TextareaAutosize from 'react-textarea-autosize';
 import { cn } from '../../lib/cn';
 import { Button } from '../../ui/button';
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/ai-chat-container/step-settings-ai-chat-container.tsx
```diff
@@ -40,9 +40,29 @@ const StepSettingsAiChatContainer = ({
...
+      scrollViewportRef.current &&
+      !isEmpty &&
+      showAiChat &&
+      !hasScrolledOnce.current
+    ) {
+      scrollViewportRef.current.scrollTop =
+        scrollViewportRef.current.scrollHeight;
+      hasScrolledOnce.current = true;
+    }
+  }, [isEmpty, showAiChat, stepName]);
```
rSnapkoOpenOps, line 65: similar to AI chat, but also scrolling if step name is changed

packages/ui-components/src/components/ai-chat-container/ai-assistant-chat-container.tsx
```diff
@@ -45,6 +45,23 @@ const AiAssistantChatContainer = ({
...
+        !isEmpty &&
+        showAiChat &&
+        !hasScrolledOnce.current
+      ) {
+        scrollViewportRef.current.scrollTop =
+          scrollViewportRef.current.scrollHeight;
+        hasScrolledOnce.current = true;
+      }
+    });
+  }, [isEmpty, showAiChat]);
```
rSnapkoOpenOps, line 64: It's wrapped in setTimeout as not always `scrollTop` is applied. Example: navigating from builder to menu

cezudas: have you tried requestAnimationFrame instead of setTimeout(fn, 0)? 🤔 It should be semantically closer to what you're trying ot do - ensure `scrollTop` is applied.

cezudas (commented):
lgtm, I left a single question for clarification

cezudas (approved):
Not thrilled about the arbitrary setTimeout, but it works.

packages/ui-components/src/lib/constants.ts
```diff
@@ -35,3 +35,5 @@ export const PRIMITIVE_STEP_METADATA = {
 };
 
 export const COPY_PASTE_TOAST_DURATION = 2000;
+
+export const AiChatScrollDelay = 200;
```
alexandrudanpop, line 39: please use the same naming convention for constants

alexandrudanpop (approved):

PR #687, opened 2025-05-26
Use currentSelectedData based on the feature flag on server/engine

MarceloRGonc:
Fixes OPS-1711

## Additional Notes
- We need to pass all the step outputs to the engine to execute a test block because of the references to other blocks.
- We still have an issue if all the step outputs of a workflow are greater than the request size limit

### DIFF
--- a/packages/server/api/src/app/flow-template/cloud-template.controller.ts
+++ b/packages/server/api/src/app/flow-template/cloud-template.controller.ts
@@ -64,6 +64,7 @@ export const cloudTemplateController: FastifyPluginAsyncTypebox = async (
           domains: Type.Optional(Type.Array(Type.String())),
           blocks: Type.Optional(Type.Array(Type.String())),
           version: Type.Optional(Type.String()),
+          categories: Type.Optional(Type.Array(Type.String())),
         }),
       },
     },
--- a/packages/server/api/src/app/flow-template/flow-template.controller.ts
+++ b/packages/server/api/src/app/flow-template/flow-template.controller.ts
@@ -28,6 +28,7 @@ export const flowTemplateController: FastifyPluginAsyncTypebox = async (
           domains: Type.Optional(Type.Array(Type.String())),
           blocks: Type.Optional(Type.Array(Type.String())),
           version: Type.Optional(Type.String()),
+          categories: Type.Optional(Type.Array(Type.String())),
         }),
       },
     },
@@ -41,6 +42,7 @@ export const flowTemplateController: FastifyPluginAsyncTypebox = async (
         projectId: request.principal.projectId,
         organizationId: request.principal.organization.id,
         version: request.query.version,
+        categories: request.query.categories,
       });
     },
   );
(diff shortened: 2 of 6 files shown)


packages/engine/src/api-handler.ts
```diff
@@ -67,7 +68,7 @@ export const start = async (): Promise<void> => {
     setStopHandlers(app);
 
     await blocksBuilder();
-
+    await encryptionKeyInitializer();
```
MarceloRGonc, line 71: We need to initialise the encryption key on the engine.

packages/engine/src/lambda-handler.ts
```diff
@@ -21,6 +26,8 @@ export async function lambdaHandler(
 
   const data = await parseJson<EngineRequest>(event.body);
 
+  await encryptionKeyInitializer();
```
MarceloRGonc, line 29: We need to initialise the encryption key on the engine.

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -65,14 +65,21 @@ export const flowStepTestOutputService = {
     );
   },
 
-  async list(params: ListParams): Promise<FlowStepTestOutput[]> {
+  async listDecrypted(params: ListParams): Promise<FlowStepTestOutput[]> {
```
MarceloRGonc, line 68: Rename, because now we have a method that returns the values encrypted

packages/engine/src/api-handler.ts
MarceloRGonc: No. It's just to read an environment variable

packages/engine/src/lib/handler/context/test-execution-context.ts
```diff
@@ -38,6 +42,15 @@ export const testExecutionContext = {
         continue;
       }
 
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      let currentOutput: any;
+      if (stepTestOutputs && step.id && stepTestOutputs?.[step.id]) {
```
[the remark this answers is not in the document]
MarceloRGonc: The stepId is a string

packages/server/api/src/app/blocks/base-block-module.ts
```diff
@@ -150,6 +156,21 @@ const baseBlocksController: FastifyPluginAsyncTypebox = async (app) => {
     const engineToken = await accessTokenManager.generateEngineToken({
       projectId,
     });
+
+    let stepTestOutputs: Record<OpenOpsId, string> | undefined = undefined;
```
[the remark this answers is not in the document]
MarceloRGonc: Line 182, if the feature flag is disabled, we send undefined.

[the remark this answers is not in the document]
bigfluffycookie: ah true, sorry im blind

PR #688, opened 2025-05-26
Move encrypt/decrypt logic from flow step test output service

MarceloRGonc:
Part of OPS-1711.

## Additional Notes
- We need to share the logic between the API and the Engine so I'm moving the encrypt/decrypt logic to a new service in the server-shared package


packages/server/shared/src/lib/blocks/step-output-transformer.ts
```diff
@@ -0,0 +1,33 @@
+import { FileCompression } from '@openops/shared';
+import { fileCompressor } from '../file-compressor';
+import { encryptUtils } from '../security/encryption';
+
+export const stepOutputTransformer = {
```
rita-gorokhod, line 5: @MarceloRGonc is it really shared?
isn't the engine only encrypting, and the server only decrypting?

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -18,13 +17,9 @@ export const flowStepTestOutputService = {
...
     output,
   }: SaveParams): Promise<FlowStepTestOutput> {
-    const encryptOutput = encryptUtils.encryptObject(output);
-    const binaryOutput = Buffer.from(JSON.stringify(encryptOutput));
-
-    const compressedOutput = await fileCompressor.compress({
-      data: binaryOutput,
-      compression: FileCompression.GZIP,
-    });
+    const compressedOutput = await stepOutputTransformer.encryptAndCompress(
```
rita-gorokhod, line 20: I'm not sure we need the stepOutputTransformer prefix, it doesn't mention anything step specific in the implementation.
it can be just 2 util funcs, encryptAndCompress and decompressAndDecrypt

MarceloRGonc: Done.

rita-gorokhod (approved):

PR #695, opened 2025-05-27
Fix get a value from Storage block when testing a step

MarceloRGonc:
Fixes OPS-1845.

## Additional Notes
- When we test a block, the flowRunId changes each time the button is clicked. We use the flowRunId to build the key for the storage service when using the RUN scope.
- Add a static value if we are testing a block.

### DIFF
--- a/packages/blocks/store/src/lib/actions/common.ts
+++ b/packages/blocks/store/src/lib/actions/common.ts
@@ -15,15 +15,19 @@ export function getScopeAndKey(params: Params): {
       return { scope: StoreScope.PROJECT, key: params.key };
     case BlockStoreScope.FLOW:
       return { scope: StoreScope.FLOW, key: params.key };
-    case BlockStoreScope.RUN:
+    case BlockStoreScope.RUN: {
+      const runId = params.isTest ? 'test-run' : params.runId;
+
       return {
         scope: StoreScope.FLOW,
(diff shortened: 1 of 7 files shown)


packages/blocks/store/src/lib/actions/common.ts
```diff
@@ -15,15 +15,19 @@ export function getScopeAndKey(params: Params): {
       return { scope: StoreScope.PROJECT, key: params.key };
     case BlockStoreScope.FLOW:
       return { scope: StoreScope.FLOW, key: params.key };
-    case BlockStoreScope.RUN:
+    case BlockStoreScope.RUN: {
+      const runId = params.isTest ? 'test-run' : params.runId;
+
       return {
         scope: StoreScope.FLOW,
```
rita-gorokhod, line 22: is this a mistake that we're scoping it to StoreScope.FLOW ?

rita-gorokhod (approved):
LGTM see comment

packages/blocks/store/src/lib/actions/common.ts
MarceloRGonc: Discussed offline. It will stay like this.

bigfluffycookie: *what* was discussed offline haha, what was the conclusion?

PR #697, opened 2025-05-27
Merge Settings About and General into a single page

cezudas:
Fixes OPS-1848.

### DIFF
--- a/packages/server/api/src/app/project/project-controller.ts
+++ b/packages/server/api/src/app/project/project-controller.ts
@@ -1,16 +1,5 @@
-import {
-  FastifyPluginCallbackTypebox,
-  Type,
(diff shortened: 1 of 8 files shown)


packages/server/api/src/app/project/project-controller.ts
```diff
@@ -39,31 +28,3 @@ export const userProjectController: FastifyPluginCallbackTypebox = (
   });
   done();
 };
-
-export const projectController: FastifyPluginCallbackTypebox = (
```
cezudas, line 43: Removing the BE code was an extension/stretch of the original ticket scope.
I can easily revert it, it's contained in this git commit: 664915092500b258e22ec20fbfd7437d3c72d0ec

PR #698, opened 2025-05-27
Add migration to add categories in flow templates

alexandrudanpop:
Part of OPS-1701

As discussed with Marco, we don't need nested categories. The structure will be:
```
Finops Capabilities
   Categories 
      Services 
```


packages/server/api/src/app/database/migrations/1748356358052-AddTemplateCategories.ts
```diff
@@ -0,0 +1,18 @@
...
+            ALTER TABLE "flow_template"
+            ADD "categories" jsonb
+        `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            ALTER TABLE "flow_template" DROP COLUMN "categories"
+        `);
+  }
```
MarceloRGonc, line 17: We are not using it.

```suggestion
  public async down(queryRunner: QueryRunner): Promise<void> {
    throw new Error('Rollback not implemented');
  }
```

packages/server/api/src/app/database/migrations/1748356358052-AddTemplateCategories.ts
```diff
@@ -0,0 +1,18 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddTemplateCategories1748356358052 implements MigrationInterface {
+  name = 'AddTemplateCategories1748356358052';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            ALTER TABLE "flow_template"
+            ADD "categories" jsonb
```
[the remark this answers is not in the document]
alexandrudanpop: It's the same thing, columns are null by default:
https://www.postgresql.org/docs/current/ddl-default.html#DDL-DEFAULT

packages/server/api/src/app/database/migrations/1748356358052-AddTemplateCategories.ts
alexandrudanpop: would not hurt to have it if someone changes to an older version, but I'll stay consistent then.

MarceloRGonc (approved):

PR #706, opened 2025-05-28
Remove projectId from request payloads on various BE API endpoints

cezudas:
Fixes OPS-1861.

On the controllers of these API endpoints the projectId is fetched from request's principal.

### DIFF
--- a/packages/react-ui/src/app/features/flows/lib/flows-hooks.ts
+++ b/packages/react-ui/src/app/features/flows/lib/flows-hooks.ts
@@ -21,7 +21,6 @@ export type FlowsSearchState = {
 async function fetchFlows(name: string, limit: number, signal: AbortSignal) {
   return flowsApi.list(
     {
-      projectId: authenticationSession.getProjectId()!,
       limit: limit,
       name: name,
       cursor: undefined,
@@ -33,14 +32,11 @@ async function fetchFlows(name: string, limit: number, signal: AbortSignal) {
 }
 
 export const flowsHooks = {
-  useFlows: (request: Omit<ListFlowsRequest, 'projectId'>) => {
+  useFlows: (request: ListFlowsRequest) => {
     return useQuery({
       queryKey: [QueryKeys.flows, authenticationSession.getProjectId()],
       queryFn: async () => {
-        return await flowsApi.list({
-          ...request,
-          projectId: authenticationSession.getProjectId()!,
-        });
+        return await flowsApi.list(request);
       },
       staleTime: 5 * 1000,
     });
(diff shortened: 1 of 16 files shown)


packages/server/api/src/app/flows/flow-run/flow-run-controller.ts
```diff
@@ -30,12 +30,10 @@ export const flowRunController: FastifyPluginCallbackTypebox = (
   done,
 ): void => {
   app.get('/', ListRequest, async (request) => {
-    // TODO project Id will be required after May 2024, this no longer needs to be optional
     const projectId =
-      request.query.projectId ??
-      (request.principal.type === PrincipalType.SERVICE
+      request.principal.type === PrincipalType.SERVICE
```
cezudas, line 34: Okay, here I took the liberty to remove projectId from the request. 
From my observation the FE was basically sending the same projectId at all times:

`authenticationSession.getProjectId()`

packages/shared/src/lib/flows/folders/folder-requests.ts
```diff
@@ -22,17 +20,6 @@ export const DeleteFolderRequest = Type.Object({
 
 export type DeleteFlowRequest = Static<typeof DeleteFolderRequest>;
 
-export const ListFolderRequest = Type.Object({
```
cezudas, line 25: no longer used

packages/react-ui/src/app/features/flows/lib/flows-hooks.ts
```diff
@@ -33,14 +31,11 @@ async function fetchFlows(name: string, limit: number, signal: AbortSignal) {
 }
 
 export const flowsHooks = {
-  useFlows: (request: Omit<ListFlowsRequest, 'projectId'>) => {
+  useFlows: (request: ListFlowsRequest) => {
     return useQuery({
-      queryKey: [QueryKeys.flows, authenticationSession.getProjectId()],
+      queryKey: [QueryKeys.flows],
```
alexandrudanpop, line 36: I'm not sure if this query key is correct, since we also use staleTime do you think we could have issues when we use search or pagination?

alexandrudanpop:
@cezudas do you think we would need this parameter in the future for service requests? For example, in cases where I want to list flows or folders from a specific project?

cezudas:
> @cezudas, do you think we would need this parameter in the future for service requests? For example, in cases where I want to list flows or folders from a specific project? Or would this be in the auth token?

Hi @alexandrudanpop 
The question in our thread is general and not specific to one file/block, isn't it?

In this PR, I identified BE endpoints where `projectId` is already being fetched from the request principal. At the same time, `projectId` was redundantly required in the payload, although never used by the BE. 
There is one exception—namely, the list flow-runs endpoint. I can easily revert that change, although I don't think we have a real use case for it right now.

packages/server/api/src/app/flows/flow-run/flow-run-controller.ts
alexandrudanpop: i think here it should just be 
`const projectId = request.principal.projectId;`

PR #709, opened 2025-05-29
Fix duplicate action ids creation

alexandrudanpop:
Fixes OPS-1873

Deployed on UX


packages/react-ui/src/app/features/flows/lib/flows-api.ts
```diff
@@ -158,7 +158,7 @@ export const flowsApi = {
       .get<Record<string, StepOutputWithData>>(
         `/v1/flow-versions/${flowVersionId}/test-output?stepIds=${stepId}`,
       )
-      .then((response) => response[stepId]);
+      .then((response) => response[stepId] || {});
```
rSnapkoOpenOps, line 161:
```suggestion
      .then((response) => response[stepId] ?? {});
```

Copilot always asks for that

rSnapkoOpenOps (approved):

PR #712, opened 2025-05-30
Add default value for Tables Get Records action

rita-gorokhod:
Fixes OPS-1875
image

### DIFF
--- a/packages/blocks/openops-tables/src/actions/get-records-action.ts
+++ b/packages/blocks/openops-tables/src/actions/get-records-action.ts
@@ -23,11 +23,18 @@ export const getRecordsAction = createAction({
       displayName: 'Filter type',
       required: false,
       options: {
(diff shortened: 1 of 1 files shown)


packages/blocks/openops-tables/src/actions/get-records-action.ts
```diff
@@ -23,11 +23,18 @@ export const getRecordsAction = createAction({
       displayName: 'Filter type',
       required: false,
       options: {
-        options: Object.entries(FilterType).map(([key, value]) => ({
-          label: value,
-          value: key,
-        })),
+        options: [
+          {
+            label: 'Match all filters',
```
rita-gorokhod, line 28: only label change so it won't break flows

bigfluffycookie (approved):

PR #716, opened 2025-05-30
Add custom API action to GitHub block

rita-gorokhod:
Fixes OPS-1899
Fixes OPS-1900

image


packages/blocks/common/src/lib/helpers/index.ts
```diff
@@ -82,11 +82,11 @@ export function createCustomApiCallAction({
         displayName: 'Headers',
         description:
           'Authorization headers are injected automatically from your connection.',
-        required: true,
+        required: false,
```
rita-gorokhod, line 85: Fixes https://linear.app/openops/issue/OPS-1900/allow-empty-headers-and-query-params-in-custom-api-actions

PR #721, opened 2025-06-02
Allow AWS connections to use attached roles (EC2, ECS)

maor-rozenfeld:
Fixes OPS-1695.


packages/server/shared/src/lib/blocks/file-blocks-utils.ts
```diff
@@ -130,7 +130,7 @@ async function loadBlockFromFolder(
     const indexPath = join(folderPath, 'src', 'index.js') + suffix;
     const stats = await fs.stat(packageJsonDir); // Get file stats
 
-    const cacheKey = `${blockName}-${blockVersion}-${stats.mtime.getTime()}`;
+    const cacheKey = `${blockName}-${blockVersion}-${stats.mtime.getTime()}-${system.calculateConfigurationHash()}`;
```
MarceloRGonc, line 134: Just curious. What was happening?

packages/openops/src/lib/aws/auth.ts
```diff
@@ -202,18 +212,22 @@ export const amazonAuth = BlockAuth.CustomAuth({
...
-    obj !== null &&
-    typeof obj === 'object' &&
-    Object.prototype.hasOwnProperty.call(obj, 'accessKeyId') &&
-    typeof obj.accessKeyId === 'string' &&
-    Object.prototype.hasOwnProperty.call(obj, 'secretAccessKey') &&
-    typeof obj.secretAccessKey === 'string'
-  );
-}
+    if (!auth.accessKeyId && !isImplicitRoleEnabled) {
+      return { valid: false, error: 'Access Key ID is required' };
```
MarceloRGonc, line 230: Is this possible to happen?

packages/blocks/aws/src/lib/actions/cli/aws-cli.ts
```diff
@@ -1,10 +1,19 @@
...
+  if (
+    credentials.accessKeyId === undefined &&
+    !system.getBoolean(SharedSystemProp.AWS_ENABLE_IMPLICIT_ROLE)
+  ) {
+    throw new Error(
+      'AWS credentials are required, please provide accessKeyId and secretAccessKey',
+    );
+  }
   const envVars = {
     AWS_ACCESS_KEY_ID: credentials.accessKeyId,
```
MarceloRGonc, line 18: Have you tested this in an environment? If AWS_ENABLE_IMPLICIT_ROLE is active and accessKeyId is undefined. Does it work as expected?

maor-rozenfeld: yes, it does. That's the CLI default behavior, just like the SDK.

packages/openops/src/lib/aws/auth.ts
maor-rozenfeld: If you create the connection via an API call and not the UI.

packages/server/shared/src/lib/blocks/file-blocks-utils.ts
maor-rozenfeld: Initially changing the env var didn't change the "require" state of the connection attributes, because the blocks metadata is cached. Had to add the env vars to the cache key. Will add a comment.

packages/blocks/aws/src/lib/actions/cli/aws-cli.ts
```diff
@@ -1,10 +1,19 @@
 import { runCliCommand } from '@openops/common';
+import { SharedSystemProp, system } from '@openops/server-shared';
 
 export async function runCommand(
   command: string,
   region: string,
   credentials: any,
 ): Promise<string> {
+  if (
+    credentials.accessKeyId === undefined &&
```
rita-gorokhod: might be better to do !accessKeyId to deal with empty string too and null.
Also add tests

packages/openops/src/lib/aws/auth.ts
```diff
@@ -166,17 +171,22 @@ export const amazonAuth = BlockAuth.CustomAuth({
   authProviderDisplayName: 'AWS',
   authProviderLogoUrl: `https://static.openops.com/blocks/aws.png`,
   props: {
```
rita-gorokhod, line 184: let's add a markdown to explain the new behavior and link to the docs, same as in gcp

packages/openops/src/lib/aws/auth.ts
```diff
@@ -166,17 +171,22 @@ export const amazonAuth = BlockAuth.CustomAuth({
...
+    defaultRegion: Property.ShortText({
+      displayName: 'Default Region',
       required: true,
+      defaultValue: 'us-east-1',
+    }),
+    accessKeyId: Property.SecretText({
+      displayName:
+        'Access Key ID' +
+        (isImplicitRoleEnabled
+          ? ' (leave blank to use the role assigned to the hosting environment)'
```
rita-gorokhod: maybe if you add markdown, then this can be just (optional), and add secretKey as well

PR #722, opened 2025-06-02
Add connection providers

MarceloRGonc:
Fixes OPS-1877.


packages/openops/src/lib/connection-providers/index.ts
```diff
@@ -0,0 +1,15 @@
+import { ConnectionProvider, Provider } from './providers';
+import { providerMap } from './providers-map';
+
+function getAll(): Provider[] {
+  return Object.values(providerMap);
+}
+
+function getOne(id: ConnectionProvider): Provider {
```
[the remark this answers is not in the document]
MarceloRGonc: The providerMap can't be out of sync.

packages/openops/src/lib/connection-providers/providers-map.ts
```diff
@@ -0,0 +1,76 @@
+import { ConnectionProvider, Provider } from './providers';
+
+const baseUrl = 'https://static.openops.com';
+
+export const providerMap: Record<ConnectionProvider, Provider> = {
+  [ConnectionProvider.AWS]: {
+    id: ConnectionProvider.AWS,
+    displayName: 'AWS',
+    logoUrl: `${baseUrl}/blocks/aws.png`,
+  },
```
alexandrudanpop: @MarceloRGonc for migration purposes I think we should add a field `supportedBlockTypes`, then in the migration after you adding the column, we can look at this data structure to update the provider and finally set the column as non-nullable.

MarceloRGonc: I think it makes more sense to create that mapping in the migration. Because if I add it here later, this list will appear in the block metadata. I think we'll only need this list during the migration, right? Or am I missing something?

https://linear.app/openops/issue/OPS-1903/add-migration-script-to-fill-the-provider-column-in-the-connections

MarceloRGonc: Discussed offline. It will be done in the migration

PR #723, opened 2025-06-02
Filter out tools for openops mcp

bigfluffycookie:
Fixes OPS-1871


packages/server/api/src/app/ai/mcp/openops-tools.ts
```diff
@@ -8,9 +8,63 @@ import {
...
+    summary?: string;
+    description?: string;
+    operationId?: string;
+    parameters?: unknown[];
+    requestBody?: unknown;
+    responses?: Record<string, unknown>;
+  };
+};
+
+function filterOpenApiSchema(schema: OpenAPI.Document): OpenAPI.Document {
```
MarceloRGonc, line 42: Add this inside of the method. Save the value in a global variable and return the JSON.stringify.

```
const openApiSchema = app.swagger() as OpenAPI.Document;
```
  
```
function getOpenApiSchema(): string {
```

This way we will do this once per deployment and not for every call to the AI.

PR #724, opened 2025-06-02
Improve initial visibility of failed step when opening run details

cezudas:
Fixes OPS-1894.
Upon loading the run details, if a run fails, the indices for all iterations will be set so that the failed step is rendered on the screen.
See Loom demo below: https://www.loom.com/share/fdce54571eb44cc2aa6fb9e560c4ebc2

### DIFF
--- a/packages/react-ui/src/app/features/builder/run-details/index.tsx
+++ b/packages/react-ui/src/app/features/builder/run-details/index.tsx
@@ -88,12 +88,10 @@ const FlowRunDetails = React.memo(() => {
 
   const message = getMessage(run, rententionDays);
 
-  let failedStepInfo = null;
-  if (run) {
-    failedStepInfo = flowRunUtils.findFailedStep(run);
-  }
-
   useEffectOnce(() => {
+    if (!run?.steps) return;
+    const failedStepInfo = flowRunUtils.findFailedStep(run);
+
     if (failedStepInfo && selectedStep !== failedStepInfo.stepName) {
       selectStepByName(failedStepInfo.stepName);
       document


packages/react-ui/src/app/features/flow-runs/lib/flow-run-utils.ts
```diff
@@ -155,20 +161,29 @@ function findLoopsState(
...
-        flowHelper.isChildOf(
-          step as LoopOnItemsAction | BranchAction | SplitAction,
-          failedStep,
+    (res: Record<string, number>, step: LoopOnItemsAction) => {
+      const loopName = step.name;
+      if (failedStepInfo) {
+        const isFailedStepChildOfLoop = flowHelper.isChildOf(
+          step,
+          failedStepInfo.stepName,
         );
```
alexandrudanpop: can you add tests that it also works for branch and split? also in the case where the failed step is the parent?

cezudas: Added [test](https://github.com/openops-cloud/openops/blob/b9d0256529b560dfd4fc1dcb94e68392d3f423a7/packages/react-ui/src/app/features/flow-runs/lib/tests/flow-run-utils.test.ts#L474) for "in the case where the failed step is the parent?"

cezudas: > can you add tests that it also works for branch and split

The variable rename in my initial patch was a bit misleading, I pushed one more [commit](https://github.com/openops-cloud/openops/pull/724/commits/b9d0256529b560dfd4fc1dcb94e68392d3f423a7).

> can you add tests that it also works for branch and split?
In our case the "parent" is LoopOnItemsAction, I added some basic test coverage for `isChildOf`, including [deeply](LoopOnItemsAction) nested failed step.

alexandrudanpop:
@cezudas if I manually run the flow the failed step is selected, but if I look at it from runs, the scroll works fine and shows the step, but the step is not selected:
image

alexandrudanpop (approved):

PR #727, opened 2025-06-02
Fix large test output displaying

rSnapkoOpenOps:
Fixes OPS-1895.

## Testing Checklist

Check all that apply:

By default all fields are collapsed. Strings are truncated:  
image
image
image

### DIFF
--- a/packages/react-ui/src/app/common/components/json-viewer.tsx
+++ b/packages/react-ui/src/app/common/components/json-viewer.tsx
@@ -173,42 +173,40 @@ const JsonViewer = React.memo(({ json, title }: JsonViewerProps) => {
         </div>
       </div>
 
-      {
+      {isNil(json) ? (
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/app/common/components/json-viewer.tsx
```diff
@@ -173,42 +173,40 @@ const JsonViewer = React.memo(({ json, title }: JsonViewerProps) => {
...
+              theme={viewerTheme}
+              enableClipboard={false}
+              groupArraysAfterLength={20}
+              displayDataTypes={false}
+              name={false}
+              quotesOnKeys={false}
+              src={json}
+              collapsed={1}
+              collapseStringsAfterLength={50}
+            />
```
rmn-snpk, line 206: I just removed the fragment. the only real change is adding  

`collapsed={1}`
 `collapseStringsAfterLength={50}`

PR #735, opened 2025-06-03
Add new endpoint for app connection metadata

bigfluffycookie:
Fixes OPS-1883

```
  "Databricks": {
    "authProviderKey": "Databricks",
    "authProviderDisplayName": "Databricks",
    "authProviderLogoUrl": "https://static.openops.com/blocks/databricks.png",
    "required": true,
    "description": "\n1. Go to the Databricks...\n",
    "props": {
      "accountId": {
        "displayName": "Account ID",
        "required": true,
        "type": "SHORT_TEXT",
        "defaultValidators": [
          {
            "type": "STRING"
          }
        ]
      }
      ...
    },
    "type": "CUSTOM_AUTH",
    "displayName": "Connection",
    "supportedBlocks": [
      "@openops/block-databricks"
    ]
  }
  ```

### CODE REVIEW

File: packages/server/api/src/app/app-connection/app-connection.controller.ts [hunk 1]
```diff
@@ -140,6 +142,29 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
...
+        release: await flagService.getCurrentRelease(),
+        includeHidden: false,
+        edition: system.getEdition(),
+      });
+
+      const authMetadata: Record<string, any> = {};
+
+      for (const block of blocks) {
+        if (block.auth) {
+          authMetadata[block.name] = block.auth;
```
alexandrudanpop: IMO it would be better to aggregate it by service provider not block name, with block name we will have multiple auth metadata of the same type

alexandrudanpop: I think it should be:
```json
{
    "AWS": {
        "supportedBlocks": [
            "block/aws",
            "block/aws-athena"
        ],
        "auth": {}
    }
}
```

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -140,6 +143,39 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
...
+              id: provider.id,
+              displayName: provider.displayName,
+              logoUrl: provider.logoUrl,
+              supportedBlocks: [],
+              props: block.auth ?? {},
+            };
+          }
+          providerMetadata[provider.id].supportedBlocks.push(block.name);
+        }
+      }
```
MarceloRGonc: Add this logic here: packages/server/api/src/app/app-connection/connection-providers-resolver.ts

packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -140,6 +143,39 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
...
+    GetConnectionMetadataRequest,
+    async (request): Promise<Record<string, any>> => {
+      const blocks = await blockMetadataService.list({
+        projectId: request.principal.projectId,
+        release: await flagService.getCurrentRelease(),
+        includeHidden: false,
+        edition: system.getEdition(),
+      });
+
+      const providerMetadata: Record<string, any> = {};
```
MarceloRGonc:
```suggestion
      const providerMetadata: Record<Provider, <create a type>> = {};
```

MarceloRGonc (approved):

PR #744, opened 2025-06-04
Redo

MarceloRGonc:
Fixes OPS-1877.

## Additional Notes
- Simplify architectur, remove ENUM and list of authProviders


bigfluffycookie (approved):
Its fine, but it will break my outstanding pr, i'll see if i manage to fix it tomorrow

PR #747, opened 2025-06-05
Fetch connections by auth provider in step settings

alexandrudanpop:
Fixes OPS-1884

## Additional Notes

image


packages/react-ui/src/app/features/connections/lib/app-connections-hooks.ts
```diff
@@ -46,4 +46,11 @@ export const appConnectionsHooks = {
         }, {}),
     });
   },
+  useConnectionsMetadata: () => {
+    return useQuery({
+      queryKey: [QueryKeys.appConnections, 'metadata'],
+      queryFn: () => appConnectionsApi.getMetadata(),
+      staleTime: Infinity,
+    });
+  },
```
alexandrudanpop, line 55: just some boilerplate, we don't use it yet

PR #774, opened 2025-06-10
Add Umbrella custom API action

rita-gorokhod:
Fixes OPS-1913

image
image

### DIFF
--- a/packages/react-ui/public/locales/en/translation.json
+++ b/packages/react-ui/public/locales/en/translation.json
@@ -496,6 +496,7 @@
   "Accept": "Accept",
   "Download File": "Download File",
   "Apply": "Apply",
+  "Paste sample data here": "Paste sample data here",
   "File is not available after execution.": "File is not available after execution.",
   "New chat": "New chat",
   "No workflows created yet": "No workflows created yet",
--- a/packages/react-ui/src/app/features/builder/data-selector/data-selector-utils.ts
+++ b/packages/react-ui/src/app/features/builder/data-selector/data-selector-utils.ts
@@ -294,95 +294,6 @@ const getAllStepsMentionsFromCurrentSelectedData: (
   });
 };
 
-const isPlainObject = (value: unknown): value is Record<string, unknown> => {
-  return (
-    typeof value === 'object' &&
-    value !== null &&
-    !Array.isArray(value) &&
-    !(value instanceof Map) &&
-    !(value instanceof Date)
-  );
-};
-
-const isAtomicValue = (value: unknown): boolean => {
-  return (
-    value === null ||
-    value === undefined ||
-    typeof value !== 'object' ||
-    Array.isArray(value) ||
-    value instanceof Map ||
-    value instanceof Date
-  );
-};
-
-const isEmptyArray = (value: unknown): boolean => {
-  return Array.isArray(value) && value.length === 0;
-};
-
-type MergedOutput = {
-  data: unknown;
-  usedSampleData: boolean;
-};
-
-/**
- * Merges sample data with test output, with sample data having priority.
- * If both are objects, sample data properties will override test output properties.
- * If test output is not an object, sample data will be used if it exists.
- * Handles nested objects by recursively merging them.
- * Arrays, Maps, and Dates are treated as atomic values - sample data takes precedence.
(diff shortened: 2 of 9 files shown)


packages/blocks/anodot/src/lib/get-user-accounts-action.ts
```diff
@@ -0,0 +1,33 @@
+import { createAction } from '@openops/blocks-framework';
+import { logger } from '@openops/server-shared';
+import { anadotAuth } from './anodot-auth-property';
+import { authenticateUserWithAnodot } from './common/auth';
+import { getAnodotUsers } from './common/users';
+
+export const getUserAccountsAction = createAction({
```
rita-gorokhod, line 7: I added this so I could do a loop over all accounts and then execute a custom API call

packages/blocks/anodot/src/lib/custom-anodot-api-action.ts
```diff
@@ -0,0 +1,133 @@
...
+import { createAction, Property } from '@openops/blocks-framework';
+import { accountSingleSelectProperty } from './account-property';
+import { anadotAuth } from './anodot-auth-property';
+import {
+  buildUserAccountApiKey,
+  createAnodotAuthHeaders,
+} from './common/anodot-requests-helpers';
+import { authenticateUserWithAnodot } from './common/auth';
+
+export const customAnodotApiAction = createAction({
```
rita-gorokhod: community request to support more actions

packages/blocks/anodot/src/lib/anodot-auth-property.ts
```diff
@@ -21,7 +21,7 @@ export const anadotAuth = BlockAuth.CustomAuth({
       description: 'The URL to use to request Umbrella API',
       required: true,
       validators: [Validators.url],
-      defaultValue: 'https://api.umbrellacost.io/api/v1',
+      defaultValue: 'https://api.umbrellacost.io/api',
```
rita-gorokhod, line 24: I originally went according to their new rebranded docs but it's incorrect

packages/blocks/anodot/src/lib/account-property.ts
```diff
@@ -57,9 +69,9 @@ async function getAccountDropdownState(
   } catch (error) {
     return {
       disabled: true,
+      error: (error as Error).message,
```
rita-gorokhod, line 72: minor improvement since we didn't have this at the time

packages/blocks/anodot/src/lib/account-property.ts
```diff
@@ -14,6 +14,18 @@ export function accountProperty() {
   });
 }
 
+export function accountSingleSelectProperty() {
```
rita-gorokhod, line 17: added single select so you would be forced to pick only one account when executing the custom API call

packages/blocks/anodot/src/lib/custom-anodot-api-action.ts
bigfluffycookie: Why not do

```
createCustomApiCallAction({
      baseUrl: () => {
        return 'https://slack.com/api';
      },
      auth: slackAuth,
      authMapping: async (auth) => {
        return {
          Authorization: `Bearer ${(auth as OAuth2PropertyValue).access_token}`,
        };
      },
    }),
```

just bc of the accounts?

rita-gorokhod: Yeah, there is no access_token in the connection, you need to do multiple calls to actually get the token

bigfluffycookie: I mean that you can still do, the function is async, you could do all the calls in there, the only thing im not sure how to do is add an extra property, just seems like a shame to have so much code duplication

packages/blocks/anodot/src/lib/custom-anodot-api-action.ts
```diff
@@ -0,0 +1,35 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
```
bigfluffycookie, line 1: yay much better now

bigfluffycookie (approved):

PR #776, opened 2025-06-11
Do not strip down sample data from flow version anymore

cezudas:
Fixes OPS-1921.


cezudas:
We purposely decided against it: we should not export the sample data when you export the workflow.
With export, people could put in sample data-sensitive information.

PR #777, opened 2025-06-11
Only include minimally necessary tools

bigfluffycookie:
Fixes OPS-1938

### DIFF
--- a/packages/server/api/src/app/ai/mcp/openops-tools.ts
+++ b/packages/server/api/src/app/ai/mcp/openops-tools.ts
@@ -13,19 +13,21 @@ import os from 'os';
 import path from 'path';
 import { MCPTool } from './mcp-tools';
 
-const INCLUDED_PATHS = [
-  '/v1/folders',
-  '/v1/flow-versions',
-  '/v1/flows',
-  '/v1/files',
-  '/v1/flow-templates',
-  '/v1/test-trigger',
-  '/v1/blocks',
-  '/v1/flow-runs',
-  '/v1/app-connections',
-];
-
-const EXCLUDED_OPERATIONS = ['delete'];
+const INCLUDED_PATHS: Record<string, string[]> = {
+  '/v1/files/{fileId}': ['get'],
+  '/v1/flow-versions/': ['get'],
+  '/v1/flows/{id}': ['get'],
+  '/v1/blocks/categories': ['get'],
+  '/v1/blocks/': ['get'],
+  '/v1/blocks/{scope}/{name}': ['get'],
+  '/v1/blocks/{name}': ['get'],
+  '/v1/flow-runs/': ['get'],
+  '/v1/flow-runs/{id}': ['get'],
+  '/v1/flow-runs/{id}/retry': ['post'],
+  '/v1/app-connections/': ['get', 'post', 'patch'],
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/ai/mcp/openops-tools.ts
```diff
@@ -39,45 +41,54 @@ type OpenApiPathItem = {
...
-        if (!EXCLUDED_OPERATIONS.includes(method.toLowerCase())) {
-          filteredPathItem[method] = operation;
-        }
-      }
-
-      if (Object.keys(filteredPathItem).length > 0) {
-        filteredPaths[path] = filteredPathItem;
-      }
-    }
+  if (!filteredSchema.paths) {
```
bigfluffycookie: the changes here are also to reduce cognitive complexity, which sonarcloud complained about, the relevant part is the removal of the EXCLUDED_OPERATIONS, as we now directly add it to the paths dictionary

packages/server/api/src/app/ai/mcp/openops-tools.ts
```diff
@@ -13,19 +13,21 @@ import os from 'os';
...
+  '/v1/flows/{id}': ['get'],
+  '/v1/blocks/categories': ['get'],
+  '/v1/blocks/': ['get'],
+  '/v1/blocks/{scope}/{name}': ['get'],
+  '/v1/blocks/{name}': ['get'],
+  '/v1/flow-runs/': ['get'],
+  '/v1/flow-runs/{id}': ['get'],
+  '/v1/flow-runs/{id}/retry': ['post'],
+  '/v1/app-connections/': ['get', 'post', 'patch'],
+  '/v1/app-connections/{id}': ['get'],
```
rita-gorokhod, line 28: isn't the patch on '/v1/app-connections/{id}' ?

packages/server/api/src/app/ai/mcp/openops-tools.ts
rita-gorokhod: hmm. so do we need this?

it's the equivalent of 

if (!schema.paths) {
   return schema
}

but this should've then do nothing
```
for (const [path, pathItem] of Object.entries(schema.paths))
```

packages/server/api/src/app/ai/mcp/openops-tools.ts
bigfluffycookie: hmm it isnt, but now that you mention it it probably should be? im not sure

packages/server/api/src/app/ai/mcp/openops-tools.ts
rita-gorokhod: @bigfluffycookie maybe just

```
async function filterOpenApiSchema(schema: OpenAPI.Document): Promise<OpenAPI.Document> {
  const filteredPaths: Record<string, any> = {};

  for (const [path, pathItem] of Object.entries(schema.paths ?? {})) {
    if (!INCLUDED_PATHS[path]) continue;

    filteredPaths[path] = {};
    for (const [method, op] of Object.entries(pathItem)) {
      if (INCLUDED_PATHS[path].includes(method.toLowerCase())) {
        filteredPaths[path][method] = op;
      }
    }
  }

  return { ...schema, paths: filteredPaths };
}


```

PR #780, opened 2025-06-11
Fix missing step id on replace step

alexandrudanpop:
Fixes OPS-1940

https://www.loom.com/share/0332daf8fe814895b1fab192dd18ec3e

### DIFF
--- a/packages/react-ui/src/app/features/builder/blocks-selector/block-selector-utils.ts
+++ b/packages/react-ui/src/app/features/builder/blocks-selector/block-selector-utils.ts
@@ -47,13 +47,11 @@ const getStepName = (block: StepMetadata, flowVersion: FlowVersion) => {
 };
 
 const getDefaultStep = ({
-  id,
   stepName,
   block,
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/app/features/builder/blocks-selector/block-selector-utils.ts
```diff
@@ -46,11 +47,13 @@ const getStepName = (block: StepMetadata, flowVersion: FlowVersion) => {
 };
 
 const getDefaultStep = ({
+  id,
   stepName,
   block,
   actionOrTriggerName,
   displayName,
 }: {
+  id: string;
```
rita-gorokhod: maybe I'm blind but I don't see where you actually pass null/undefined to force the new id generation

alexandrudanpop: fixed

cezudas (approved):

PR #782, opened 2025-06-11
Add migration to add step ids to missing steps

MarceloRGonc:
Fixes OPS-1958.

### DIFF
--- a/packages/server/api/src/app/database/migrations/1749644220381-AddStepIdToExistingFlowVersionsAndInsertOutputInTestOutputTable.ts
+++ b/packages/server/api/src/app/database/migrations/1749644220381-AddStepIdToExistingFlowVersionsAndInsertOutputInTestOutputTable.ts
@@ -1,0 +1,117 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { encryptUtils, fileCompressor, logger } from '@openops/server-shared';
+import { FileCompression, openOpsId } from '@openops/shared';
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddStepIdToExistingFlowVersionsAndInsertOutputInTestOutputTable1749644220381
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    logger.info(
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/database/migrations/1749644220381-AddStepIdToExistingFlowVersionsAndInsertOutputInTestOutputTable.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const { updatedJson, extractedData } = await processJsonObject(jsonData);
+
+    await queryRunner.query(
+      `UPDATE "${tableName}" SET "trigger" = $1 WHERE "id" = $2`,
+      [updatedJson, record.id],
+    );
+
+    for (const data of extractedData) {
+      const encryptOutput = encryptUtils.encryptObject(
```
rita-gorokhod: I think we don't need this @MarceloRGonc , we're only patching the step ids, no?

MarceloRGonc: You are right. Even if we don't have stepId we will not have the currentSelectedData. I already changed

packages/server/api/src/app/database/migrations/1749644220381-AddStepIdToExistingFlowVersions.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!obj || typeof obj !== 'object') {
+    return obj;
+  }
+
+  if (obj.name && obj.type && !obj.id) {
+    obj.id = openOpsId();
+  }
+
+  for (const key of Object.keys(obj)) {
+    if (Array.isArray(obj[key])) {
```
rita-gorokhod, line 53: can this happen? an array of step definitions?

MarceloRGonc: Yes, if the flow as conditions
image

rita-gorokhod (approved):

PR #783, opened 2025-06-11
Update fonts&margins and reduce layout shift when toggling sample data viewer/editor mode

cezudas:
Fixes OPS-1947.
Todo: 

- [x] turn off feature flag back again

Demo
https://www.loom.com/share/0913fea53e3749748283af5503bdd55f


packages/server/api/src/app/flags/dev-flags.service.ts
```diff
@@ -32,7 +32,7 @@ async function getAll(): Promise<Flag[]> {
     },
     {
       id: FlagId.SHOW_SAMPLE_DATA,
-      value: false,
+      value: true,
```
cezudas: just for testing on the deployed environment

PR #787, opened 2025-06-11
Fix step ids for copy paste

alexandrudanpop:
Fixes OPS-1967

### DIFF
--- a/packages/react-ui/src/app/features/builder/hooks/use-paste.ts
+++ b/packages/react-ui/src/app/features/builder/hooks/use-paste.ts
@@ -5,16 +5,60 @@ import {
 } from '@openops/components/ui';
 import {
   Action,
+  ActionType,
   flowHelper,
   FlowOperationType,
   FlowVersion,
   isNil,
+  openOpsId,
   StepLocationRelativeToParent,
 } from '@openops/shared';
 import { useCallback } from 'react';
 import { useBuilderStateContext } from '../builder-hooks';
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/app/features/builder/hooks/use-paste.ts
```diff
@@ -5,17 +5,61 @@ import {
...
+    newAction.nextAction = assignNewIdsToActions(newAction.nextAction);
+  }
+
+  if (newAction.type === ActionType.BRANCH) {
+    if (newAction.onSuccessAction) {
+      newAction.onSuccessAction = assignNewIdsToActions(
+        newAction.onSuccessAction,
+      );
+    }
+    if (newAction.onFailureAction) {
```
rita-gorokhod, line 35: Idea: we already have code that generates slugs.
Let's put the IDs only there, this way we know it's called by all relevant places.

rita-gorokhod (changes requested):

packages/react-ui/src/app/features/builder/hooks/use-paste.ts
alexandrudanpop: you mean in this file?
https://github.com/openops-cloud/openops/blob/739a450f0247fac7907ab1dd22d4ed4e6b06f5f9/packages/react-ui/src/app/features/builder/blocks-selector/block-selector-utils.ts#L36-L47

It needs to be on all operations, otherwise it won't work ADD_ACTION / UPDATE_ACTION, etc And those operations are invoked in multiple places

rita-gorokhod (approved):

PR #788, opened 2025-06-12
Remove projectId from app connection requests

bigfluffycookie:
Fixes OPS-1968


cezudas (approved):
`projectId` is no longer required in the payload and the FE no longer sends it when submitting create connection form!

Screenshot 2025-06-12 at 14 38 38

PR #790, opened 2025-06-12
Remove USE_CONNECTIONS_PROVIDER reference from BE

MarceloRGonc:
Fixes OPS-1889.

### DIFF
--- a/packages/server/api/src/app/app-connection/app-connection.controller.ts
+++ b/packages/server/api/src/app/app-connection/app-connection.controller.ts
@@ -5,7 +5,6 @@ import {
 } from '@fastify/type-provider-typebox';
 import {
   AppConnectionWithoutSensitiveData,
-  FlagId,
   ListAppConnectionsRequestQuery,
   OpenOpsId,
   PatchAppConnectionRequestBody,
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/app-connection/app-connection.controller.ts
```diff
@@ -73,26 +71,20 @@ export const appConnectionController: FastifyPluginCallbackTypebox = (
...
-      const { name, status, cursor, limit } = request.query;
-      let { blockNames, authProviders } = request.query;
+      const { name, status, cursor, limit, blockNames } = request.query;
+      let { authProviders } = request.query;
 
-      const featureFlag = await devFlagsService.getOne(
-        FlagId.USE_CONNECTIONS_PROVIDER,
-      );
-      if (blockNames && featureFlag?.value) {
+      if (blockNames) {
```
rita-gorokhod, line 77: why do we still need the IF? 
the FE now always sends authProviders, no?

MarceloRGonc: When I spoke to Alex it was necessary to keep filtering by block name. But something may have changed. @alexandrudanpop Can I remove the filter by blockName?

MarceloRGonc: I will do it in another PR

alexandrudanpop (approved):

PR #791, opened 2025-06-12
Do not use the blockName from the request in the GET and PATCH endpoints of connection controller

MarceloRGonc:
Part of OPS-1959.

## Additional notes:
- We can't change the models yet.


MarceloRGonc (commented):
We can't change the models yet. I'm preparing another PR

bigfluffycookie (approved):

PR #754, opened 2025-06-06
Merge sample data with test output for data selector

cezudas:
Fixes OPS-1916.

Quick test:

I set `step_20` (Get Repos) sample data using Postman.
Screenshot 2025-06-06 at 09 28 56
Test output data was already available on `step_20`
Screenshot 2025-06-06 at 09 30 20

In the data selector of the loop step `step_21`, I expand "Get repos" nodes and I see the merged result on the UI.
Screenshot 2025-06-06 at 09 27 43

PR #792, opened 2025-06-12
Sample data overrides when available test output data in the data selector

cezudas:
Fixes OPS-1951.

Follows the initial implementation from "Merge sample data with test output for data selector" #754

If the sample data has one of the following values - `{}, "", undefined, null ` - then the data selector will display the test output (if available). Otherwise (non-empty sample data available), sample data will be displayed instead.

Deployed to UX
Todo: Turn feature flag back to false before merging


packages/server/api/src/app/flags/dev-flags.service.ts
```diff
@@ -32,7 +32,7 @@ async function getAll(): Promise<Flag[]> {
     },
     {
       id: FlagId.SHOW_SAMPLE_DATA,
-      value: false,
+      value: true,
```
cezudas: this will be rolled-back before merging

PR #795, opened 2025-06-12
Remove USE_CONNECTIONS_PROVIDER reference from FE 

MarceloRGonc:
Fixes OPS-1890.


packages/react-ui/src/app/features/templates/components/connections-picker/connections-picker.tsx
```diff
@@ -73,13 +68,10 @@ const ConnectionsPicker = ({
...
     refetch,
-  } = appConnectionsHooks.useGroupedConnections(
-    {
-      blockNames: integrations.map((integration) => integration.name),
-      limit: 10000,
-    },
-    useConnectionsProvider ?? false,
-  );
+  } = appConnectionsHooks.useGroupedConnections({
+    blockNames: integrations.map((integration) => integration.name),
```
MarceloRGonc: @rita-gorokhod and @alexandrudanpop here we still use the blockNames

rita-gorokhod: @MarceloRGonc why do we need this instead of sending the block's auth provider?

alexandrudanpop: in connection picker we have a list of blocks, so it's easier query the connections by blockName. Is it so hard to maintain it in the API? 
We eighter keep it in API or need to query metadata every time to check each block and query by providers instead.

rita-gorokhod: we can keep in the API, it will help the LLM too

MarceloRGonc: I changed to this. And I think it works.

```
    authProviders: integrations.flatMap((integration) =>
      integration.auth?.authProviderKey
        ? [integration.auth.authProviderKey]
        : [],
    ),
```

PR #800, opened 2025-06-13
Add AI Streaming-specific setting for nginx.standard config

cezudas:
Fixes CI-51.

It uses the same settings as the [`nginx.gateway`](https://github.com/openops-cloud/openops/commit/7f5f8ecdc87b4e31320c54d9b9010e6e31843143) configuration.

Demo on Openops test environment:
https://www.loom.com/share/ae81b67b4bb941fc81f232a0776b444a

### DIFF
--- a/nginx.standard.conf
+++ b/nginx.standard.conf
@@ -77,6 +77,11 @@ http {
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection "Upgrade";
             proxy_set_header Host $host;
+
+            # --- AI Streaming-specific settings ---
+            proxy_buffering off;
+            proxy_cache off;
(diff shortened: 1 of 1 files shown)


nginx.standard.conf
```diff
@@ -77,6 +77,11 @@ http {
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection "Upgrade";
             proxy_set_header Host $host;
+
+            # --- AI Streaming-specific settings ---
```
cezudas, line 81: We can try to limit the impact by matching only the relevant /conversation endpoints, as applying these directives to the entire /api/ block may degrade backend performance by disabling buffering and caching.

MarceloRGonc (approved):

PR #803, opened 2025-06-13
Fix extra call when doing test step / trigger

alexandrudanpop:
Fixes OPS-1957

Put data from WS directly in the cache, we don't need an extra HTTP call

### DIFF
--- a/packages/react-ui/src/app/features/builder/test-step/test-trigger-section.tsx
+++ b/packages/react-ui/src/app/features/builder/test-step/test-trigger-section.tsx
@@ -27,13 +27,15 @@ import {
   CATCH_WEBHOOK,
   isNil,
   SeekPage,
-  StepOutputWithData,
   Trigger,
   TriggerEvent,
   TriggerTestStrategy,
 } from '@openops/shared';
 
-import { stepTestOutputCache } from '../data-selector/data-selector-cache';
+import {
+  setStepOutputCache,
+  stepTestOutputCache,
+} from '../data-selector/data-selector-cache';
 import { stepTestOutputHooks } from './step-test-output-hooks';
 import { TestSampleDataViewer } from './test-sample-data-viewer';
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/app/features/builder/test-step/test-trigger-section.tsx
```diff
@@ -249,6 +269,7 @@ const TestTriggerSection = React.memo(
                     );
                     if (triggerEvent) {
                       updateSelectedData(triggerEvent);
+                      setCurrentSelectedId(value);
```
alexandrudanpop, line 258: the dropdown for webhook was broken

packages/react-ui/src/app/features/builder/test-step/test-trigger-section.tsx
```diff
@@ -197,7 +205,7 @@ const TestTriggerSection = React.memo(
     });
 
     const currentTestOutput = testOutputData?.output;
-    const sampleDataSelected =
+    const outputDataSelected =
```
alexandrudanpop, line 206: just renaming for semantics

packages/react-ui/src/app/features/builder/test-step/test-trigger-section.tsx
```diff
@@ -131,8 +130,11 @@ const TestTriggerSection = React.memo(
       },
       onSuccess: async (results) => {
         if (results.length > 0) {
-          updateSelectedData(results[0]);
-          refetch();
+          const lastResult = results[results.length - 1];
+          updateSelectedData(lastResult);
+
+          // setStepOutputCache handles output and cache
```
cezudas: can we remove it?

alexandrudanpop: done

cezudas (approved):

PR #808, opened 2025-06-13
Search connections by blockName if importing from a template

MarceloRGonc:
Fixes #<issue_number>.

## Additional Notes
- When importing a template, the property authPropertyKey does not exist

### DIFF
--- a/packages/shared/src/lib/flows/flow-operations.ts
+++ b/packages/shared/src/lib/flows/flow-operations.ts
@@ -51,6 +51,7 @@ export const ImportFlowRequest = Type.Object({
   displayName: Type.String({}),
   description: Type.Optional(Type.String({})),
   trigger: Trigger,
+  blockToProviderMap: Type.Optional(Type.Object({})),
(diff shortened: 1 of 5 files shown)


packages/shared/src/lib/flows/flow-operations.ts
```diff
@@ -47,11 +47,22 @@ export const LockFlowRequest = Type.Object({});
...
+export type AppConnectionsWithSupportedBlocks = Static<
+  typeof AppConnectionsWithSupportedBlocks
+>;
+
 export const ImportFlowRequest = Type.Object({
   displayName: Type.String({}),
   description: Type.Optional(Type.String({})),
   trigger: Trigger,
-  connections: Type.Optional(Type.Array(AppConnectionWithoutSensitiveData)),
+  connections: Type.Optional(AppConnectionsWithSupportedBlocks),
```
alexandrudanpop: Looks good for create flow from template. Let's check also import.

PR #811, opened 2025-06-16
Add cloudzero block with custom api action

bigfluffycookie:
Fixes OPS-1935


.verdaccio/config.yml
```diff
@@ -0,0 +1,28 @@
+# path to a directory with all packages
```
rita-gorokhod: remove, will be handled as part of https://linear.app/openops/issue/OPS-1931/update-create-block-cli-command

project.json
```diff
@@ -10,6 +10,14 @@
           "npx playwright test -c packages/tests-e2e/playwright.config.ts packages/tests-e2e/tests/*.spec.ts"
         ]
       }
+    },
+    "local-registry": {
```
rita-gorokhod: undo

packages/blocks/cloudzero/src/lib/auth.ts
```diff
@@ -0,0 +1,9 @@
+import { BlockAuth } from '@openops/blocks-framework';
+
+export const cloudzeroAuth = BlockAuth.SecretAuth({
```
rita-gorokhod, line 20: Add a markdown please how to generate the auth.

Do you know also if the auth token has any expiry?

bigfluffycookie: from my understanding no https://docs.cloudzero.com/reference/authorization

packages/blocks/cloudzero/src/lib/auth.ts
```diff
@@ -0,0 +1,27 @@
...
+
+> ⚠️ WARNING: The API key will not be shown again after creation.
+
+For more information, visit the [CloudZero API documentation](https://docs.cloudzero.com/reference/authorization).
+`;
+
+export const cloudzeroAuth = BlockAuth.SecretAuth({
+  displayName: 'API Key',
+  required: true,
+  authProviderKey: 'cloudzero',
```
MarceloRGonc, line 23:
```suggestion
  authProviderKey: 'Cloudzero',
```

Just to be consistent with other authProviderKeys :)

PR #819, opened 2025-06-17
Add sample data step indication in canvas

alexandrudanpop:
Fixes OPS-1952

https://www.loom.com/share/9600e9a8ff244767aa51a50db0256d65


packages/server/api/src/app/flags/dev-flags.service.ts
```diff
@@ -26,7 +26,7 @@ async function getAll(): Promise<Flag[]> {
     },
     {
       id: FlagId.SHOW_SAMPLE_DATA,
-      value: false,
+      value: true,
```
alexandrudanpop, line 29: enable flag, as we plan to release the feature

packages/ui-components/src/tailwind.css
```diff
@@ -105,8 +108,9 @@
     --primary: 0 0% 100%;
     --primary-100: 204 38% 75%;
     --primary-200: 239 65% 61%;
-    --primary-300: 250 75% 12%;
+    --primary-300: 265 79% 71%;
```
alexandrudanpop, line 111: this color was too dark, so you could not read the text in darkmode (as seen in chromatic diffs)

PR #825, opened 2025-06-17
Add cloudhealth block with graphql action

bigfluffycookie:
Fixes OPS-1936

image

### DIFF
--- a/packages/blocks/cloudhealth/src/index.ts
+++ b/packages/blocks/cloudhealth/src/index.ts
@@ -1,5 +1,5 @@
 import { createCustomApiCallAction } from '@openops/blocks-common';
-import { createBlock } from '@openops/blocks-framework';
+import { createBlock, Property } from '@openops/blocks-framework';
 import { BlockCategory } from '@openops/shared';
 import { cloudhealthAuth } from './lib/auth';
 import { graphqlAction } from './lib/graphql-action';
@@ -15,6 +15,12 @@ export const cloudhealth = createBlock({
     createCustomApiCallAction({
       baseUrl: () => 'https://chapi.cloudhealthtech.com/v1',
       auth: cloudhealthAuth,
+      additionalProps: {
+        documentation: Property.MarkDown({
+          value:
+            'For more information, visit the [CloudHealth API documentation](https://apidocs.cloudhealthtech.com/).',
+        }),
+      },
       authMapping: async (context) => ({
         Authorization: `Bearer ${context.auth as string}`,
         'Content-Type': 'application/json',


packages/blocks/cloudhealth/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const cloudhealth = createBlock({
+  displayName: 'CloudHealth',
+  auth: cloudhealthAuth,
+  minimumSupportedRelease: '0.20.0',
+  logoUrl: 'https://static.openops.com/blocks/cloudhealth.png',
+  authors: [],
+  categories: [BlockCategory.FINOPS],
+  actions: [
+    createCustomApiCallAction({
```
bigfluffycookie, line 15: to make the normal api requests

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,57 @@
+import { createAction, Property } from '@openops/blocks-framework';
+import { cloudhealthAuth } from './auth';
+
+async function getAccessToken(apiKey: string) {
```
bigfluffycookie, line 34: so, they do have a refresh token, but its only valid for 9 hours so I dont think its worth to deal with it tbh

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,57 @@
...
+  displayName: 'Execute GraphQL Query',
+  description: 'Execute a GraphQL query against the CloudHealth API',
+  auth: cloudhealthAuth,
+  props: {
+    query: Property.LongText({
+      displayName: 'GraphQL Query',
+      description: 'The GraphQL query to execute',
+      required: true,
+    }),
+    variables: Property.Json({
```
rita-gorokhod, line 63: can this be instead like the query in custom API? 
param, value ?
just thinking that this is not so intuitive 

also, is there a way to infer from the query what kind of params you need to pass? maybe we can auto-fill those?

bigfluffycookie: This is what it looks like on cloudhealth though, query and variables, i think it would be more confusing to change it

And to your second question, im not sure probably not without making the request first

packages/blocks/cloudhealth/src/lib/graphql-action.ts
MarceloRGonc: So, we are requesting an access token every time, right? The request can't take 9 hours, so we are safe.

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,57 @@
...
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ query, variables: variables ?? {} }),
+    });
+
+    if (!response.ok)
+      throw new Error(`GraphQL request failed: ${response.statusText}`);
+    const result = await response.json();
+    if (result.errors)
+      throw new Error(`GraphQL Error: ${result.errors[0].message}`);
```
MarceloRGonc: Regarding logs, how is this being handled? Do we have enough information to try to resolve the issues?

packages/blocks/cloudhealth/src/lib/graphql-action.ts
bigfluffycookie: Yes Everytime, so let me clarify, the access token expires after 15 minutes, the refresh token after 9 hours, but i think as soon as we started the request the expiration shouldnt matter

packages/blocks/cloudhealth/src/lib/graphql-action.ts
rita-gorokhod: I think it's not clear to the user though what exactly you're supposed to put in the variables.
Isn't it key:value ?

packages/blocks/cloudhealth/src/lib/graphql-action.ts
bigfluffycookie: wdym with "Regarding logs" are you saying i should add logs here? 

and I mean i think so, the only errors ive gotten so far were unauthorized but that one was pretty clear on how to deal with it

MarceloRGonc: You are throwing errors. I think they are not being handled anywhere and are being sent directly to the UI. My question is more whether these errors are already being logged somewhere. If not, then you should add logs.

packages/blocks/cloudhealth/src/lib/graphql-action.ts
bigfluffycookie: @rita-gorokhod so? can i leave it or nah

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,67 @@
...
+      method: 'POST',
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ query, variables: variables ?? {} }),
+    });
+
+    if (!response.ok) {
+      const errorMessages = `GraphQL request failed: ${response.statusText}`;
```
rita-gorokhod: it would be useful to output also what the query was, if they return it

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,67 @@
...
+      logger.error(errorMessages);
+      throw new Error(errorMessages);
+    }
+
+    const result = await response.json();
+
+    if (result.errors) {
```
rita-gorokhod, line 86: same here: it would be useful to output also what the query was, if they return it

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,67 @@
...
+      const errorMessages = `GraphQL Error: ${result.errors[0].message}`;
```
rita-gorokhod: why are you taking just [0], not one string.join on all?

packages/blocks/cloudhealth/src/index.ts
rita-gorokhod, line 15: as we discussed offline, should we add maybe a link to their docs? perhaps as a markdown prop?
https://apidocs.cloudhealthtech.com/

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,67 @@
...
+    throw new Error('Invalid API key or login failed');
+  return result.data.loginAPI.accessToken;
+}
+
+export const graphqlAction = createAction({
+  name: 'graphql_query',
+  displayName: 'Execute GraphQL Query',
+  description: 'Execute a GraphQL query against the CloudHealth API',
+  auth: cloudhealthAuth,
+  props: {
```
rita-gorokhod, line 54: maybe here also markdown for https://help.cloudhealthtech.com/graphql-api/

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,102 @@
...
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({ query, variables: variables ?? {} }),
+    });
+
+    const result = await response.json();
+
+    if (result.errors) {
```
bigfluffycookie, line 85: had to switch this and the response.ok around, logically i think theres 3 cases

1. is not ok and has error
2. is not ok and does not have errors
3. is ok and does not have errors

packages/blocks/cloudhealth/src/lib/graphql-action.ts
```diff
@@ -0,0 +1,106 @@
...
+          .map((error: any) => error.message)
+          .join(' | ')}`,
+      );
+    }
+
+    if (!response.ok) {
+      errorMessages.push(`GraphQL request failed: ${response.statusText}`);
+    }
+
+    if (errorMessages.length > 0) {
```
bigfluffycookie, line 98: i had to change this a bit , logically i think theres 3 cases

is not ok and has error
is not ok and does not have errors
is ok and does not have errors

packages/blocks/cloudhealth/src/lib/graphql-action.ts
bigfluffycookie: from my outputs i can not see a query property in the response

packages/blocks/cloudhealth/src/lib/auth.ts
```diff
@@ -0,0 +1,23 @@
+import { BlockAuth } from '@openops/blocks-framework';
+
+const markdown = `
```
bigfluffycookie, line 3: Personally i find it a bit intruding

image

packages/blocks/cloudhealth/src/index.ts
bigfluffycookie: image

packages/blocks/cloudhealth/src/lib/graphql-action.ts
bigfluffycookie: from my outputs i can not see a query property in the response

packages/blocks/cloudhealth/src/lib/graphql-action.ts
bigfluffycookie: mistake

rita-gorokhod (approved):

PR #831, opened 2025-06-18
Add new endpoint for test step and add to tools

bigfluffycookie:
Fixes OPS-1996

image


packages/server/api/src/app/flows/flow.module.ts
```diff
@@ -24,13 +24,15 @@ import { flowVersionService } from './flow-version/flow-version.service';
...
 import { stepRunService } from './step-run/step-run-service';
+import { testStepController } from './test-step/test-step.controller';
 import { testTriggerController } from './test-trigger/test-trigger-controller';
 
 export const flowModule: FastifyPluginAsyncTypebox = async (app) => {
   await app.register(flowWorkerController, { prefix: '/v1/worker/flows' });
   await app.register(flowVersionController, { prefix: '/v1/flow-versions' });
   await app.register(flowController, { prefix: '/v1/flows' });
   await app.register(testTriggerController, { prefix: '/v1/test-trigger' });
+  await app.register(testStepController, { prefix: '/v1/test-step' });
```
bigfluffycookie: alternatively, we could also add an endpoint /test/ and put both test trigger and test step there. (and in the future test workflow)

packages/server/api/src/app/ai/mcp/openops-tools.ts
```diff
@@ -29,6 +29,8 @@ const INCLUDED_PATHS: Record<string, string[]> = {
   '/v1/app-connections/': ['get', 'patch'],
   '/v1/app-connections/{id}': ['get'],
   '/v1/app-connections/metadata': ['get'],
+  '/v1/test-step/{workflowId}/{stepId}': ['post'],
```
bigfluffycookie: i feel like its a bit awkward to pass it as params in the url but idk, maybe its because of the endpoint in general, see my comment below

packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -189,6 +191,52 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
       });
     },
   );
+
+  fastify.post(
```
rita-gorokhod: @MarceloRGonc why did you want this under this route? 
I think it's a bit weird, shouldn't this be a separate route/entity, of "runs"? 
we're calling stepRunService, which is another clue that this doesn't belong here, it's a separate entity.

packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -189,6 +191,52 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
...
+      if (!step) {
+        await reply.status(StatusCodes.NOT_FOUND).send({
+          success: false,
+          output: 'Step not found',
+        });
+        return;
+      }
+
+      try {
+        const result = await stepRunService.create({
```
rita-gorokhod: what happens if try to execute a step and it depends on output from other steps, and that output is missing? (i.e. if you put a dependency on {{step_3}} and that was not tested?)

bigfluffycookie: hmmm, it will try to run it and for the ai the field will have 'undefined' in it, it has the same behavior as when you test it manually, it will still run but just fail
image
image

packages/server/api/src/app/flows/test/test.controller.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const testController: FastifyPluginAsyncTypebox = async (fastify) => {
+  fastify.post(
+    '/:flowVersionId/:stepId',
+    TestStepRequest,
+    async (request, reply) => {
+      const { flowVersionId, stepId } = request.params;
+      const projectId = request.principal.projectId;
+
+      const flowVersion = await flowVersionService.getOneOrThrow(flowVersionId);
```
rita-gorokhod: do you know where is the check that the flow belongs to the current projectId?

bigfluffycookie: nowhere, added it now

packages/server/api/src/app/flows/flow/flow-version.controller.ts
MarceloRGonc: If the route is '/v1/test/', I think the flowVersionId and stepId should not be the path, they should be query parameters. If we want to use the flowVersionId in the path it must be below the '/flow-version/' which is the entity.

packages/server/api/src/app/flows/test/test.controller.ts
MarceloRGonc: I think you can use the 'entitiesMustBeOwnedByCurrentProject'. However, I'm not entirely convinced it will work for this case. But if you can't use it, you can do something similar.

bigfluffycookie: we cannot use entitiesMustBeOwnedByCurrentProject in here, the code im using now is the same as in flow version controller.

packages/server/api/src/app/flows/test/test.controller.ts
```diff
@@ -0,0 +1,109 @@
...
+  flowVersion: FlowVersion,
+  projectId: string,
+  reply: any,
+): Promise<boolean> {
+  const flow = await flowService.getOne({
+    id: flowVersion.flowId,
+    projectId,
+  });
+  if (flow === null || flow === undefined) {
+    await reply.status(StatusCodes.BAD_REQUEST).send({
```
MarceloRGonc: This is not a BadRequest, it is a permissions issue, should be a FORBIDDEN. You are trying to execute a workflow without your token being in the correct project.

If this code is somewhere else it might be better to extract it.

MarceloRGonc (approved):

PR #833, opened 2025-06-18
Fix resetting sample data on step updates

alexandrudanpop:
Fixes OPS-1994

## Additional Notes
Because the data is saved without updating the main settings form, every update the the step resulted in losing the sample data. We simply need to update the form and it will go through the same update process as the rest of the form updates.

https://github.com/openops-cloud/openops/blob/9d849da8a4a019f74ec3845ea04270ee2fe92a45/packages/react-ui/src/app/features/builder/step-settings/index.tsx#L204


cezudas (approved):
👌

PR #838, opened 2025-06-19
Fix schema validation for import and use template with mandatory ids

alexandrudanpop:
Fixes OPS-1912

Deployed to UX, both import of old flows without ids and creating from template should work fine.


packages/shared/src/lib/flows/flow-operations.ts
```diff
@@ -241,7 +238,7 @@ export const FlowOperationRequest = Type.Union([
   Type.Object(
     {
       type: Type.Literal(FlowOperationType.UPDATE_TRIGGER),
-      request: UpdateTriggerRequest,
+      request: TriggerWithOptionalId,
     },
```
alexandrudanpop, line 242: there is no add trigger request, both add and update are via this flow operation

packages/shared/src/lib/flows/dto/flow-template-request.ts
```diff
@@ -1,8 +1,8 @@
 import { Static, Type } from '@sinclair/typebox';
 import { BaseModelSchema } from '../../common';
-import { FlowVersion } from '../flow-version';
+import { NoIdFlowVersion } from '../flow-version';
 import { Trigger } from '../triggers/trigger';
-export const FlowVersionTemplate = Type.Omit(FlowVersion, [
+export const FlowVersionTemplate = Type.Omit(NoIdFlowVersion, [
```
MarceloRGonc: Instead of creating the 'NoIdFlowVersion' type I think we can change the FlowVersionTemplate to the code below. Many of the fields are being omitted, so I think it makes more sense.

```
export const FlowVersionTemplate = Type.Object({
  displayName: Type.String(),
  description: Type.Optional(Type.String()),
  trigger: Type.Omit(Trigger, ['id']),
  valid: Type.Boolean(),
});

File: packages/shared/src/lib/flows/flow-version.ts [hunk 3]
```diff
@@ -22,6 +22,17 @@ export const FlowVersion = Type.Object({
...
+  ...BaseModelSchema,
+  flowId: Type.String(),
+  displayName: Type.String(),
+  description: Type.Optional(Type.String()),
+  trigger: Type.Omit(Trigger, ['id']),
+  updatedBy: Nullable(Type.String()),
+  valid: Type.Boolean(),
+  state: Type.Enum(FlowVersionState),
+});
+
```
MarceloRGonc: See the comment above.

MarceloRGonc (approved):

PR #847, opened 2025-06-23
Move test trigger to new path

bigfluffycookie:
Fixes OPS-2003

image


packages/server/api/src/app/flows/flow.module.ts
```diff
@@ -24,15 +24,13 @@ import { flowVersionService } from './flow-version/flow-version.service';
...
 import { stepRunService } from './step-run/step-run-service';
-import { testTriggerController } from './test-trigger/test-trigger-controller';
 import { testController } from './test/test.controller';
 
 export const flowModule: FastifyPluginAsyncTypebox = async (app) => {
   await app.register(flowWorkerController, { prefix: '/v1/worker/flows' });
   await app.register(flowVersionController, { prefix: '/v1/flow-versions' });
   await app.register(flowController, { prefix: '/v1/flows' });
   await app.register(testController, { prefix: '/v1/test' });
-  await app.register(testTriggerController, { prefix: '/v1/test-trigger' });
```
bigfluffycookie, line 35: so i didnt find this actually being used in the frontend but would be happy if @cezudas could confirm

packages/server/api/src/app/flows/test/test.controller.ts
```diff
@@ -98,6 +104,18 @@ export const testController: FastifyPluginAsyncTypebox = async (fastify) => {
       });
     }
   });
+
+  fastify.post('/trigger', TestTriggerRequest, async (request) => {
```
bigfluffycookie, line 108: copied it over as is from test-trigger-controller.ts

packages/server/api/src/app/flows/flow.module.ts
cezudas: it's not used anywhere, all good

PR #849, opened 2025-06-23
Add CLI prompts to add boilerplate code for Authentication and opinionated folder structure

cezudas:
Fixes OPS-1931.

Because create-block was getting too big with the additional boilerplate template code, I created new modules and moved some of the code there.

### DIFF
--- a/packages/cli/src/lib/commands/create-block.ts
+++ b/packages/cli/src/lib/commands/create-block.ts
@@ -55,7 +55,7 @@ export const createBlock = async (
   if (createOpinionatedStructure) {
     console.log(
       chalk.blue(
-        `📁 Opinionated folder structure created with sample actions, tests, and service layer`,
+        `📁 Opinionated folder structure created with sample actions and custom API call action`,
       ),
     );
   }
--- a/packages/cli/src/lib/commands/create-block.ts
+++ b/packages/cli/src/lib/commands/create-block.ts
@@ -42,8 +42,6 @@ export const createBlock = async (
   authType: string,
   createOpinionatedStructure: boolean,
 ) => {
-  await validateBlockName(blockName);
-  await validatePackageName(packageName);
   await checkIfBlockExists(blockName);
   await nxGenerateNodeLibrary(blockName, packageName);
   await setupGeneratedLibrary(blockName, authType, createOpinionatedStructure);
@@ -71,13 +69,7 @@ export const createBlockCommand = new Command('create')
         type: 'input',
         name: 'blockName',
         message: 'Enter the block name:',
-        validate: (input: string) => {
-          const blockNamePattern = /^[A-Za-z0-9-]+$/;
-          if (!blockNamePattern.test(input)) {
-            return 'Block names can only contain lowercase letters, numbers, and hyphens.';
-          }
-          return true;
-        },
+        validate: validateBlockName,
       },
       {
         type: 'input',
@@ -85,13 +77,7 @@ export const createBlockCommand = new Command('create')
         message: 'Enter the package name:',
         default: (answers: any) => `@openops/block-${answers.blockName}`,
         when: (answers: any) => answers.blockName !== undefined,
-        validate: (input: string) => {
-          const packageNamePattern = /^(?:@[a-zA-Z0-9-]+\/)?[a-zA-Z0-9-]+$/;
-          if (!packageNamePattern.test(input)) {
-            return 'Package names can only contain lowercase letters, numbers, and hyphens.';
-          }
-          return true;
-        },
+        validate: validatePackageName,
       },
       {
         type: 'list',
--- a/packages/cli/src/lib/commands/create-block.ts
+++ b/packages/cli/src/lib/commands/create-block.ts
@@ -1,179 +1,66 @@
 import chalk from 'chalk';
 import { Command } from 'commander';
-import { readdir, unlink, writeFile } from 'fs/promises';
 import inquirer from 'inquirer';
-import assert from 'node:assert';
-import { findBlockSourceDirectory } from '../utils/block-utils';
-import { exec } from '../utils/exec';
 import {
-  readJestConfig,
-  readPackageEslint,
-  readProjectJson,
-  writeJestConfig,
-  writePackageEslint,
-  writeProjectJson,
-} from '../utils/files';
+  nxGenerateNodeLibrary,
+  removeUnusedFiles,
+  updateEslintFile,
+  updateJestConfigFile,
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/commands/create-block.ts
```diff
@@ -197,16 +71,65 @@ export const createBlockCommand = new Command('create')
         type: 'input',
         name: 'blockName',
         message: 'Enter the block name:',
+        validate: (input: string) => {
```
bigfluffycookie: why do we validate twice? whats the point of the import above? same for the package validation why have these ?   await validateBlockName(blockName);
  await validatePackageName(packageName); and then the below aswell

packages/cli/src/lib/commands/create-block.ts
```diff
@@ -1,192 +1,66 @@
...
+  generateIndexTsFile,
+  generateOpinionatedStructure,
+} from '../utils/block-templates';
+import {
+  checkIfBlockExists,
+  validateBlockName,
+  validatePackageName,
+} from '../utils/block-validators';
 
-const nxGenerateNodeLibrary = async (
```
bigfluffycookie, line 53: when moving things around please add comments next time, its hard to review whats new and whats moved else

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+    })
+    .join('');
+
+  let authTemplate = '';
+
+  switch (authType) {
+    case 'secret':
+      authTemplate = `import { BlockAuth } from '@openops/blocks-framework';
+
+const markdown = \`
```
bigfluffycookie: please indent properly

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+  required: true,
+  authProviderKey: '${blockName}',
+  authProviderDisplayName: '${capitalizeFirstLetter(blockName)}',
+  authProviderLogoUrl: 'https://static.openops.com/blocks/${blockName}.png',
+  description: markdown,
+});
+`;
+      break;
+
+    case 'custom':
```
bigfluffycookie, line 94: same here, please indent correclty, difficult to see whats string and whats code

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+  await mkdir(`packages/blocks/${blockName}/src/lib/actions`, {
+    recursive: true,
+  });
+  await mkdir(`packages/blocks/${blockName}/src/lib/common`, {
+    recursive: true,
+  });
+  await mkdir(`packages/blocks/${blockName}/test/actions`, { recursive: true });
+  await mkdir(`packages/blocks/${blockName}/test/common`, { recursive: true });
+
+  // Generate sample action
```
bigfluffycookie: remove comment, indent correctly

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+  const actionTemplate = `import { createAction, Property } from '@openops/blocks-framework';
+${
+  authType !== 'none'
+    ? `import { ${blockNameCamelCase}Auth } from '../auth';`
+    : ''
+}
+
+export const sampleAction = createAction({
```
bigfluffycookie: does this need to be exported?

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+
+  await writeFile(
+    `packages/blocks/${blockName}/src/lib/actions/sample-action.ts`,
+    actionTemplate,
+  );
+
+  // Generate common service
+  const serviceTemplate = `import { ${
+    authType !== 'none' ? `OAuth2PropertyValue, ` : ''
+  }Property } from '@openops/blocks-framework';
```
bigfluffycookie: space after }

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
+import { writeFile } from 'fs/promises';
```
bigfluffycookie, line 2: please remove all unnecessary comments, eg. all the ones that are not useful to users.

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+    .join('');
+
+  let authTemplate = '';
+
+  switch (authType) {
+    case 'secret':
+      authTemplate = `import { BlockAuth } from '@openops/blocks-framework';
+
+const markdown = \`
+To get your ${capitalizeFirstLetter(blockName)} API key:
```
bigfluffycookie: whats the point of this being there as a default? Its gonna be different for all auths no?

packages/cli/src/lib/utils/block-templates.ts
```diff
@@ -0,0 +1,479 @@
...
+  },
+});
+`;
+
+  await writeFile(
+    `packages/blocks/${blockName}/src/lib/actions/sample-action.ts`,
+    actionTemplate,
+  );
+
+  // Generate common service
```
bigfluffycookie: Was this discussed with anyone else? I dont see this as a requirement in the ticket, and im not sure how useful it is

packages/cli/src/lib/utils/block-templates.ts
cezudas: I think so, I couldn't find a block action which is not exported.

packages/cli/src/lib/utils/block-templates.ts
cezudas: I will pass an empty string literal for description then

packages/cli/src/lib/utils/block-templates.ts
cezudas: I make a few changes - after each multiline literal, I put a new line so that the literal starts left-aligned.
Unfortunately, every JS expression in the literal (ternary, switch) will be formatted by Prettier.

packages/cli/src/lib/commands/create-block.ts
cezudas: `validateBlockName` and `validatePackageName` will now only be invoked on the individual input level. The validation is interactive; it will give the user the chance to enter a correct value without exiting the CLI process.

bigfluffycookie (approved):

PR #850, opened 2025-06-23
Add new reference every time on new empty condition

alexandrudanpop:
Fixes OPS-2002


packages/react-ui/src/app/features/builder/step-settings/branch-settings/index.tsx
```diff
@@ -16,6 +16,15 @@ const emptyCondition: ValidBranchCondition = {
...
 
+const createEmptyCondition = (): ValidBranchCondition => {
+  return {
+    firstValue: '',
+    secondValue: '',
+    operator: BranchOperator.TEXT_CONTAINS,
+    caseSensitive: false,
+  };
+};
+
```
rmn-snpk: Do we still need `emptyCondition`? Maybe we can reuse it in the function or remove it

PR #856, opened 2025-06-23
Use codemirror to display step output

alexandrudanpop:
Fixes OPS-480

We don't need `react-json-view` as it's unmaintained and does not support virtual scroll.

Deployed on UX.

### DIFF
--- a/ai-prompts/mcp-aws-cost-unavailable.txt
+++ b/ai-prompts/mcp-aws-cost-unavailable.txt
@@ -1,5 +1,6 @@
 AWS COST MCP CONFIGURATION HINT
 -------------------------------
-If the user asks about their AWS cost, billing, and spending, reply with information how to view this in AWS console, or fetch it via AWS CLI, or other OpenOps tools.
-Do not attempt to simulate or infer AWS cost data unless given by the user, and instead recommend the user to enable AWS Cost MCP integration in \"AI\" section under the Settings page.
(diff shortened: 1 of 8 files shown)


packages/ui-components/src/components/json-viewer/json-content.tsx
```diff
@@ -63,7 +66,7 @@ export const JsonContent = ({
   }
 
   return (
-    <div className="pt-[11px] pl-3 border-t border-solid">
+    <div className="">
```
rmn-snpk:
```suggestion
    <div>
```

rmn-snpk:
Divider is not visible when showing not object value

image

rmn-snpk:
There is no scroll so I'm not able to see the bottom border of the viewer 
image

packages/ui-components/src/components/json-viewer/json-content.tsx
```diff
@@ -25,15 +24,17 @@ type JsonContentProps = {
   editorClassName?: string;
 };
 
+const isEmptyString = (value: any): boolean => {
+  return typeof value === 'string' && value.trim() === '';
+};
```
rmn-snpk: not a blocker, but I would move it inside component and add memorization

alexandrudanpop:
> Divider is not visible when showing not object value
> 
> image

fixed

packages/ui-components/src/components/json-viewer/json-content.tsx
alexandrudanpop: fixed

cezudas (approved):
neat!

PR #859, opened 2025-06-24
Add new V2 endpoint for 'Add step'

bigfluffycookie:
Fixes OPS-2017


packages/server/api/src/app/ai/mcp/openops-tools.ts
```diff
@@ -19,8 +19,10 @@ const INCLUDED_PATHS: Record<string, string[]> = {
   '/v1/flows/': ['get'],
   '/v1/flows/count': ['get'],
   '/v1/flows/{id}': ['get'],
+  '/v2/flows/{id}/steps': ['post'],
   '/v1/blocks/categories': ['get'],
   '/v1/blocks/': ['get'],
+  '/v1/blocks/options': ['post'],
```
bigfluffycookie: for blocks with dynamic properties it needs this endpoint

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  Permission,
+  PopulatedFlow,
+  PrincipalType,
+  SERVICE_KEY_SECURITY_OPENAPI,
+} from '@openops/shared';
+import { StatusCodes } from 'http-status-codes';
+import { entitiesMustBeOwnedByCurrentProject } from '../../authentication/authorization';
+import { flowService } from './flow.service';
+
+export const flowV2Controller: FastifyPluginAsyncTypebox = async (app) => {
```
bigfluffycookie, line 19: the logic that was implemented is basically indetical to flow post, the reason we cannot just use flow post is because it is not granular enough and gives the AI too much access. This way we have a new v2 endpoint for the bot to use that is not too broad.

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const AddStepRequestOptions = {
+  config: {
+    allowedPrincipals: [PrincipalType.USER, PrincipalType.SERVICE],
+    permission: Permission.UPDATE_FLOW_STATUS,
+  },
+  schema: {
+    tags: ['flows-v2'],
+    description:
+      'Add a new step to a workflow. This endpoint allows you to add actions, triggers, or other step types to an existing workflow. The step will be added at the specified location relative to the parent step. Note: The parentStep parameter should be the name of the step (e.g., "step_1", "step_3", "trigger"), not the step ID.',
```
bigfluffycookie, line 58: the note is there because the bot assumes parentStep: string is the id, instead of the name, not sure if we should change the api there or not

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+  SERVICE_KEY_SECURITY_OPENAPI,
+} from '@openops/shared';
+import { StatusCodes } from 'http-status-codes';
+import { entitiesMustBeOwnedByCurrentProject } from '../../authentication/authorization';
+import { flowService } from './flow.service';
+
+export const flowV2Controller: FastifyPluginAsyncTypebox = async (app) => {
+  app.addHook('preSerialization', entitiesMustBeOwnedByCurrentProject);
+
+  app.post('/:id/steps', AddStepRequestOptions, async (request) => {
```
bigfluffycookie, line 22: Also, the ui doesnt update immediately, you need to refresh to see the change

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -0,0 +1,67 @@
+import {
```
bigfluffycookie, line 1: im no longer gonna add theses paths to be included in the mcp, they serve to test the agent functionality, they dont work well anyways for the assistant. I will remove some other endpoints aswell its cluttering the mcp. -> https://linear.app/openops/issue/OPS-2029/remove-recent-additions-from-included-paths-in-openops-tools

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -0,0 +1,67 @@
...
+          request: request.body,
+        },
+      });
+
+      return updatedFlow;
+    } catch (error) {
+      logger.error('Error in V2 Add Step endpoint', {
+        error: error instanceof Error ? error.message : String(error),
+        stack: error instanceof Error ? error.stack : undefined,
+      });
```
MarceloRGonc:
```suggestion
      logger.error('Error adding step to workflow.', {
        body: request.body,
        flowId: request.params.id,
        error: error instanceof Error ? error.message : String(error),
      });
```

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
MarceloRGonc: Did you try to change the parentStep property to parentStepName?

bigfluffycookie: We could but this property comes from "addActionRequest" which is used in other places aswell so we'd have to change it in all the places that use the original post add action request ( or i could also make a new request type if we want )

PR #864, opened 2025-06-25
Add v2 endpoint for update step

bigfluffycookie:
Fixes OPS-2028

### DIFF
--- a/packages/server/api/src/app/flows/flow/flow-v2.controller.ts
+++ b/packages/server/api/src/app/flows/flow/flow-v2.controller.ts
@@ -70,7 +70,6 @@ export const flowV2Controller: FastifyPluginAsyncTypebox = async (app) => {
       return updatedFlow;
     } catch (error) {
       logger.error('Error in V2 Update Step endpoint', {
-      logger.error('Error updating workflow step.', {
         body: request.body,
         flowId: request.params.id,
         stepName: request.params.stepName,


packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -44,6 +45,36 @@ export const flowV2Controller: FastifyPluginAsyncTypebox = async (app) => {
       throw error;
     }
   });
+
+  app.put('/:id/steps/:stepName', UpdateStepRequestOptions, async (request) => {
```
bigfluffycookie, line 50: as with the other endpoint, its basically what happens in the original endpoint of post flow but this way the ai doesnt get immediate access to all the post logic

packages/server/api/src/app/flows/flow/flow-v2.controller.ts
```diff
@@ -44,6 +45,36 @@ export const flowV2Controller: FastifyPluginAsyncTypebox = async (app) => {
...
+          request: {
+            ...request.body,
+            name: stepName,
+          },
+        },
+      });
+
+      return updatedFlow;
+    } catch (error) {
+      logger.error('Error in V2 Update Step endpoint', {
```
MarceloRGonc, line 72:
```suggestion
      logger.error('Error in V2 Update Step endpoint', {
      logger.error('Error updating workflow step.', {
        body: request.body,
        flowId: request.params.id,
        stepName: request.params.stepName,
```

MarceloRGonc (approved):

PR #865, opened 2025-06-25
Fix AWS cost MCP availability check for mcp-aws-cost-unavailable.txt prompt loading

cezudas:
Fixes OPS-2032.

Out of scope: fixing `ai-prompts/mcp-aws-cost-unavailable.txt` prompt itself


packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -121,9 +121,9 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
       aiConfig,
     });
 
-    const isAwsCostMcpAvailable =
-      hasToolProvider(filteredTools, 'cost-analysis') ||
-      hasToolProvider(filteredTools, 'cost-explorer');
+    const isAwsCostMcpDisabled =
+      !hasToolProvider(tools, 'cost-analysis') &&
```
cezudas, line 125: depending on the user prompt, using filteredTools with `hasToolProvider` can lead to `isAwsCostMcpAvailable` getting evaluated as false when it should be true  (AWS cost mcp is enabled in the mcp configuration).

I had an initial "primer" discution with Leyla and Marcelo on why this change is needed.

PR #866, opened 2025-06-25
Validate email and password when creating an user

MarceloRGonc:
Fixes OPS-1658.


packages/server/api/src/app/authentication/authentication-service/index.ts
```diff
@@ -21,16 +21,21 @@ import { Provider } from './hooks/authentication-service-hooks';
...
 
     const { token, refresh_token } = await openopsTables.createUser({
-      name: `${params.firstName} ${params.lastName}`,
+      name,
       email: params.email,
       password: params.password,
       authenticate: true,
     });
 
+    const user = await createUser(params);
```
MarceloRGonc, line 37: Changed order, if it fails in tables, we don't create the user in the editor

packages/shared/src/lib/password-validation.ts
```diff
@@ -60,6 +70,30 @@ const passwordValidation = {
...
-export { emailRegex, passwordRules, passwordValidation };
+const assertValidPassword = (password: string): void => {
+  const isLengthValid =
+    password.length >= MIN_LENGTH && password.length <= MAX_LENGTH;
+
+  if (!isLengthValid) {
+    throw new ApplicationError({
+      code: ErrorCode.INVALID_USER_PASSWORD,
+      params: {
+        message: `Password must be between ${MIN_LENGTH} and ${MAX_LENGTH} characters.`,
```
bigfluffycookie, line 81: Do passwords commonly have a max length?

packages/shared/src/lib/password-validation.ts
```diff
@@ -60,6 +70,30 @@ const passwordValidation = {
     NUMBER_REGEX.test(value) || validationMessages.number,
 };
 
-const emailRegex =
-  /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-export { emailRegex, passwordRules, passwordValidation };
+const assertValidPassword = (password: string): void => {
```
bigfluffycookie, line 73: btw, not entirely related but do we currently have a way to reset a password? Like what happens if a user forgets it

bigfluffycookie (approved):

packages/shared/src/lib/password-validation.ts
MarceloRGonc: We don't have a way to do it.

bigfluffycookie: Yikes, i guess worst case they can change it in the DB haha

PR #877, opened 2025-06-27
Block making AI chat calls for pending statuses

rSnapkoOpenOps:
Fixes OPS-1740.

https://www.loom.com/share/a4ec27b03b8e4a91ad6330322fb9ed4d?sid=29b0ad50-7e01-4a41-8277-4f01e18bc1a3

### DIFF
--- a/packages/ui-components/src/components/ai-chat-container/ai-chat-input.tsx
+++ b/packages/ui-components/src/components/ai-chat-container/ai-chat-input.tsx
@@ -6,15 +6,17 @@ import { Button } from '../../ui/button';
 import { ScrollArea } from '../../ui/scroll-area';
 import { AiModelSelector, AiModelSelectorProps } from './ai-model-selector';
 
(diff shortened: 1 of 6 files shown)


packages/ui-components/src/components/ai-chat-container/ai-chat-input.tsx
```diff
@@ -6,9 +6,15 @@ import { Button } from '../../ui/button';
...
 
+export const ChatStatus = {
+  STREAMING: 'streaming',
+  SUBMITTED: 'submitted',
+};
+
 export type AiChatInputProps = {
   className?: string;
   placeholder?: string;
+  status?: string;
```
cezudas: I think Copilot's remark is legit,  I think it should not be string, but a type based on ChatStatus

rmn-snpk: Done

PR #878, opened 2025-06-27
Add live progress for runs

alexandrudanpop:
Fixes OPS-1123

## Additional Notes
Update requests are sent to client through websocket. Only the runId is sent through the WS. The react client will check if the user views the runId from the WS and it will refetch it. If there are too many requests from WS to the react-app while requests to get runs are pending previous requests will be cancelled

## Testing Checklist

Check all that apply:

Deployed on test.
https://www.loom.com/share/30eef78de0e94c62be5e5fc59ee52c13


packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -135,15 +147,28 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
...
+      !isNil(executionStateContentLength) && executionStateContentLength > 0;
+    if (!updateLogs) {
+      if (progressUpdateType === ProgressUpdateType.TEST_FLOW) {
+        app.io
+          .to(request.principal.projectId)
+          .emit(WebsocketClientEvent.TEST_FLOW_RUN_PROGRESS, runId);
+      } else {
+        app.io
+          .to(request.principal.projectId)
+          .emit(WebsocketClientEvent.FLOW_RUN_PROGRESS, runId);
```
alexandrudanpop: I wonder if we need both `TEST_FLOW_RUN_PROGRESS` and `FLOW_RUN_PROGRESS`, I would keep only `FLOW_RUN_PROGRESS` unless we want to have some telemetry based on them?

alexandrudanpop: I received info from Marco, that scheduled runs (production runs don't need live update). This will make it better from a technical perspective as we don't send too many websocket requests in case there are many concurrent scheduled runs.

jest.env.js
```diff
@@ -1,2 +1,2 @@
-
 process.env.OPS_QUEUE_MODE = 'MEMORY';
+process.env.OPS_ENGINE_URL = 'http://localhost:3005';
```
cezudas, line 2: Is it needed?
Does `packages/server/api/test/unit/engine/engine-controller.test.ts` or `packages/engine/test/services/progress.service.test.ts` fail without it?

packages/react-ui/src/app/features/builder/hooks/use-run-progress.ts
```diff
@@ -0,0 +1,74 @@
...
+
+import { useSocket } from '@/app/common/providers/socket-provider';
+import { flowRunsApi } from '@/app/features/flow-runs/lib/flow-runs-api';
+import { FlowRun, FlowVersion, WebsocketClientEvent } from '@openops/shared';
+
+interface UseRunProgressParams {
+  run: FlowRun | null;
+  setRun: (run: FlowRun, flowVersion: FlowVersion) => void;
+  flowVersion: FlowVersion | null;
+}
```
rmn-snpk: [nitpick] we use `type` in most of the files

jest.env.js
alexandrudanpop: yep, we need it for `engine-controller` tests

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -135,15 +136,14 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
       tags: runDetails.tags ?? [],
     });
 
-    if (populatedRun.status === FlowRunStatus.RUNNING) {
-      return;
-    }
```
MarceloRGonc, line 140: Don't remove this. Move this code to after the web-socket event, but before marking the job as complete.

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -135,15 +136,14 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
...
-    if (runDetails.status === FlowRunStatus.PAUSED) {
+    if (
+      runDetails.status === FlowRunStatus.PAUSED &&
+      !isNil(runDetails.pauseMetadata)
+    ) {
       await flowRunService.pause({
         flowRunId: runId,
         pauseMetadata: {
-          ...runDetails.pauseMetadata!,
+          ...runDetails.pauseMetadata,
```
MarceloRGonc: ...(runDetails.pauseMetadata ?? {}),

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -135,15 +136,14 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
...
     });
 
-    if (populatedRun.status === FlowRunStatus.RUNNING) {
-      return;
-    }
-
-    if (runDetails.status === FlowRunStatus.PAUSED) {
+    if (
+      runDetails.status === FlowRunStatus.PAUSED &&
+      !isNil(runDetails.pauseMetadata)
```
MarceloRGonc:
```suggestion
      runDetails.status === FlowRunStatus.PAUSED
```

rita-gorokhod: how can it be null and paused at the same time? if it's paused, we have to have the pauseMetadata 🤔

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -152,9 +152,10 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
         },
       });
     }
+
     app.io
-      .to(populatedRun.projectId)
-      .emit(WebsocketClientEvent.TEST_FLOW_RUN_PROGRESS, populatedRun);
+      .to(request.principal.projectId)
+      .emit(WebsocketClientEvent.FLOW_RUN_PROGRESS, runId);
```
rita-gorokhod, line 155: I'm curious why we're not passing the populatedRun as before, what was the problem?

packages/react-ui/src/app/features/flows/lib/flows-api.ts
```diff
@@ -172,7 +142,7 @@ function getInitialRun(
   socket: Socket,
   flowVersionId: string,
 ): Promise<FlowRun> {
-  return new Promise<FlowRun>((resolve, reject) => {
```
rita-gorokhod, line 175: can you explain why this was removed?

packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -77,20 +77,14 @@ export const flowExecutor = {
         break;
       }
 
-      const isNotNested = flowExecutionContext.currentPath.path.length === 0;
-      const sendContinuousProgress =
-        isNotNested &&
-        constants.progressUpdateType === ProgressUpdateType.TEST_FLOW;
```
alexandrudanpop: we don't need those filters if we want continuous progress when executing the flow

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -1,52 +1,70 @@
...
-
-      await fetch(url.toString(), {
-        method: 'POST',
-        headers: {
-          'Content-Type': 'application/json',
-          Authorization: `Bearer ${engineConstants.engineToken}`,
-        },
-        body: JSON.stringify(request),
-      });
+      await sendUpdateRunRequest(params);
```
alexandrudanpop, line 16: Just extracted sendUpdateRunRequest to a separate function, and changed the implementation to do request deduplication like in AP

packages/server/api/src/app/workers/engine-controller.ts
alexandrudanpop: I don't think it's a good practice to send large payloads to the clients through websockets. For example:
React app user is viewing Run1, but engine is sending updates because it's executing Run2, now those payloads for Run2 are send to the app. With this approach it just receives a runId instead of a full run payload.

packages/react-ui/src/app/features/flows/lib/flows-api.ts
alexandrudanpop: This was previously supposed to track the flow run progress, now we the hook from `use-run-progress` to track the progress and this just initiates the run.

packages/react-ui/src/app/features/builder/hooks/use-run-progress.ts
alexandrudanpop: fixed

packages/server/api/src/app/workers/engine-controller.ts
alexandrudanpop: fixed

packages/server/api/src/app/workers/engine-controller.ts
alexandrudanpop: done

packages/server/api/src/app/workers/engine-controller.ts
alexandrudanpop: done

alexandrudanpop: please re-check

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -1,52 +1,70 @@
...
+    workerHandlerId: engineConstants.serverHandlerId ?? null,
+    runDetails: await flowExecutorContext.toResponse(),
+    progressUpdateType: engineConstants.progressUpdateType,
+  };
+
+  // Request deduplication using hash comparison
+  const requestHash = crypto
+    .createHash('sha256')
+    .update(JSON.stringify(request))
+    .digest('hex');
```
MarceloRGonc: You can use the hashObject method. Also this is not working because the duration time changes.
```suggestion
  const requestHash = hashUtils.hashObject(request, (key, value) => {
    if (key === 'duration') return undefined;
    return value;
  });
```

packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -77,20 +77,14 @@ export const flowExecutor = {
...
-        progressService
-          .sendUpdate({
-            engineConstants: constants,
-            flowExecutorContext: flowExecutionContext,
-          })
-          .catch((error) => {
-            logger.error('Error sending progress update', error);
-          });
-      }
+      progressService
```
rita-gorokhod: I see that we're doing this after "if (flowExecutionContext.verdict !== ExecutionVerdict.RUNNING) {", so we won't get to it if it was failed or paused, no?

packages/engine/src/lib/services/progress.service.ts
alexandrudanpop: fixed, thanks

packages/engine/src/lib/handler/flow-executor.ts
alexandrudanpop: adjusted, I'll deploy again to test

rSnapkoOpenOps (approved):

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -1,52 +1,69 @@
...
+    workerHandlerId: engineConstants.serverHandlerId ?? null,
+    runDetails: await flowExecutorContext.toResponse(),
+    progressUpdateType: engineConstants.progressUpdateType,
+  };
+
+  // Request deduplication using hash comparison
+  const requestHash = hashUtils.hashObject(request, (key, value) => {
+    if (key === 'duration') return undefined;
+    return value;
+  });
```
MarceloRGonc, line 49: Because we are calling executeFromAction recursively, the event we are sending may have already been sent. We are hashing the request to validate if we have already sent the event and reduce the number of requests.

Example: We have a branch with a code block inside.

executeFromAction (Branch) -> executeFromAction (Code)

When the code block finishes, it calls the sendUpdateRunRequest and returns the result. Then the Branch finishes and calls the sendUpdateRunRequest with the same result. The only difference is the duration.

PR #884, opened 2025-06-30
Use a valid string to test the AI connection

MarceloRGonc:
Fixes OPS-2064

### DIFF
--- a/packages/openops/src/lib/ai/providers.ts
+++ b/packages/openops/src/lib/ai/providers.ts
@@ -102,7 +102,7 @@ export const validateAiProviderConfig = async (
 
     await generateText({
       model: languageModel,
-      prompt: '',
+      prompt: 'Is the connection valid?',
       ...config.modelSettings,
     });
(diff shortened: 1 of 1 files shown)


packages/openops/src/lib/ai/providers.ts
```diff
@@ -102,7 +102,7 @@ export const validateAiProviderConfig = async (
 
     await generateText({
       model: languageModel,
-      prompt: '',
+      prompt: 'Is the connection valid?',
```
rita-gorokhod, line 105: Nice, does it cause any problems with the other models?

rita-gorokhod (approved):

packages/openops/src/lib/ai/providers.ts
MarceloRGonc: I also tested with OpenAI, and it worked. Additionally, it's the same as we do with streaming. We pass a message to the LLM.

PR #891, opened 2025-07-01
Ensure error instances passed as the second argument to the logger are included in result log

cezudas:
Fixes OPS-2067.

- [x] Added more test coverage for `cleanLogEvent` relevant for the usecase when event data is an error

I tested using local-dev with `prettier-pino`
Before:
The error object passed as the second argument to the logger is excluded, and it does not end up in the result log.
Screenshot 2025-07-01 at 13 51 01

After:
The error object passed as the second argument to the logger is included; event data will contain "message" and "stack" from the provided error instance.
Screenshot 2025-07-01 at 13 48 54

The events are still automatically enriched when some request context is available
Screenshot 2025-07-01 at 14 05 12


packages/server/shared/test/log-cleaner.test.ts
```diff
@@ -70,22 +70,6 @@ describe('log-cleaner', () => {
     });
   });
 
-  it('should map error object', () => {
```
cezudas, line 73: I moved this test down below in the 'should map error object' describe block.

packages/server/shared/src/lib/logger/index.ts
```diff
@@ -94,8 +103,18 @@ function enrichEvent(event: object): object {
     clientIp: requestContext.get('clientIp' as never) ?? undefined,
   };
 
+  const baseEvent =
+    event instanceof Error
+      ? {
+          ...event,
+          message: event.message,
+          name: event.name,
+          stack: event.stack,
```
MarceloRGonc:
```suggestion
          stack: truncate(event.stack, 2000)
```

cezudas: I think stack is already truncated in log-cleaner https://github.com/openops-cloud/openops/blob/feat/runs-progress/packages/server/shared/src/lib/logger/log-cleaner.ts#L61-L68

cezudas:
> There is a cleanLogEvent method that checks if the event is an error. I think it will never happen now. Please review and clarify the logic.

yes, there is a cleanLogEvent branch checking if the event is an error.
it wasn't reaching before, and it is reached now

https://github.com/openops-cloud/openops/blob/feat/runs-progress/packages/server/shared/src/lib/logger/log-cleaner.ts#L61-L68

packages/server/shared/src/lib/logger/index.ts
```diff
@@ -53,10 +57,15 @@ function initLogger(): Logger {
           const levelString = pino.levels.labels[level] ?? 'info';
           const message =
             inputArgs && inputArgs.find((arg) => typeof arg === 'string');
+
+          if (isError(eventData) || !isEmpty(eventData)) {
```
MarceloRGonc: Do we need this condition? We can always enrich the event

packages/server/shared/src/lib/logger/log-cleaner.ts
```diff
@@ -59,13 +59,15 @@ export const cleanLogEvent = (logEvent: any) => {
...
+    eventData.stack = truncate(stack, 2000);
+    eventData.name = name;
     if (!logEvent.message) {
-      logEvent.message = truncate(logEvent.event.message);
+      logEvent.message = truncate(message);
     } else {
-      eventData.message = truncate(logEvent.event.message);
+      eventData.message = truncate(message);
     }
+    Object.assign(eventData, rest);
```
MarceloRGonc: Do we hit this block?

cezudas: yes

packages/server/shared/src/lib/logger/index.ts
cezudas: good idea, done

PR #900, opened 2025-07-01
Add Vega Cloud block

rita-gorokhod:
Fixes OPS-1937

### DIFF
--- a/packages/blocks/vegacloud/src/lib/auth.ts
+++ b/packages/blocks/vegacloud/src/lib/auth.ts
@@ -1,0 +1,32 @@
+import { BlockAuth, Property } from '@openops/blocks-framework';
+
+const markdown = `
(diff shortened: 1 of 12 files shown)


packages/blocks/vegacloud/src/lib/auth.ts
```diff
@@ -0,0 +1,32 @@
...
+`;
+
+export const vegacloudAuth = BlockAuth.CustomAuth({
+  authProviderKey: 'vegacloud',
+  authProviderDisplayName: 'Vega Cloud',
+  authProviderLogoUrl: 'https://static.openops.com/blocks/vegacloud.png',
+  description: markdown,
+  required: true,
+  props: {
+    clientSecret: Property.SecretText({
```
rita-gorokhod, line 14: small comment: I'm putting the secret before the id because that's how it appears in Vega's UI, so if the user is copy pasting, they can make a mistake otherwise (I did..)

PR #901, opened 2025-07-02
Fix Delay block inside a loop

MarceloRGonc:
Fixes OPS-2071.

image


packages/server/worker/src/lib/executors/repeating-job-executor.ts
```diff
@@ -52,7 +54,12 @@ async function executeRepeatingJob({
     });
     return;
   }
-  if (populatedFlow?.status === FlowStatus.DISABLED) {
+
+  if (
+    data.jobType === RepeatableJobType.EXECUTE_TRIGGER &&
+    data.environment === RunEnvironment.PRODUCTION &&
+    populatedFlow?.status === FlowStatus.DISABLED
```
MarceloRGonc: If it's a test, the workflow can be disabled.

packages/server/api/src/app/workers/worker-controller.ts
```diff
@@ -183,7 +183,9 @@ export const flowWorkerController: FastifyPluginAsyncTypebox = async (app) => {
     async (request) => {
       const data = request.body;
       await flowRunService.start({
-        payload: null,
+        payload: {
+          queryParams: data.resumePayload,
+        },
```
MarceloRGonc, line 188: The resume workflow is expecting the execution "path" to be inside the payload as queryParams.

https://github.com/openops-cloud/openops/blob/53e0466591b18c3c24b28dea1408548219c43850/packages/engine/src/lib/handler/loop-executor.ts#L43

packages/server/shared/src/lib/job/job-data.ts
```diff
@@ -36,13 +36,19 @@ export const RepeatingJobData = Type.Object({
...
+});
+
 // Never change without increasing LATEST_JOB_DATA_SCHEMA_VERSION, and adding a migration
 export const DelayedJobData = Type.Object({
   projectId: Type.String(),
   environment: Type.Enum(RunEnvironment),
   schemaVersion: Type.Number(),
   flowVersionId: Type.String(),
   runId: Type.String(),
+  resumePayload: Type.Optional(ResumePayload),
```
MarceloRGonc, line 51: Adding the resumePayload to the Delayed queue. We need this information to proper resume the workflows

packages/server/worker/src/lib/executors/flow-job-executor.ts
```diff
@@ -140,6 +140,10 @@ async function executeFlow(
       await handleTimeoutError(jobData, engineToken);
     } else {
       await handleInternalError(jobData, engineToken, e as Error);
+      jobStatus = JobStatus.FAILED;
+      jobFinalMessage = `Internal error reported by engine. Error message: ${
+        (e as Error).message
+      }`;
```
MarceloRGonc: Missing from the previous PR

packages/server/worker/src/lib/executors/repeating-job-executor.ts
```diff
@@ -140,6 +144,17 @@ const consumeDelayedJob = async (
...
+  data: ScheduledJobData,
+  populatedFlow: PopulatedFlow | null,
+): boolean => {
+  return (
+    data.jobType === RepeatableJobType.EXECUTE_TRIGGER &&
+    data.environment === RunEnvironment.PRODUCTION &&
+    populatedFlow?.status === FlowStatus.DISABLED
+  );
+};
+
```
MarceloRGonc, line 157: If it's a test, the workflow can be disabled; however, we want to maintain the schedule and execute it.

packages/server/worker/src/lib/executors/flow-job-executor.ts
rita-gorokhod: please extract to a separate PR, I know it's overhead but it's a very difficult area to understand and I don't want to mix the commits

packages/server/worker/src/lib/executors/repeating-job-executor.ts
```diff
@@ -140,6 +144,17 @@ const consumeDelayedJob = async (
   await workerApiService(workerToken).resumeRun(data);
 };
 
+const shouldSkipDisabledFlow = (
+  data: ScheduledJobData,
+  populatedFlow: PopulatedFlow | null,
+): boolean => {
+  return (
+    data.jobType === RepeatableJobType.EXECUTE_TRIGGER &&
```
rita-gorokhod, line 152: why do we need this different condition?

MarceloRGonc: Only the EXECUTE_TRIGGER type will start a new run. And we still want to skip if it's in a published workflow and execution that it's disabled.

rita-gorokhod (approved):

PR #903, opened 2025-07-02
Refactor to always send workflow progress.

MarceloRGonc:
Part of OPS-2071.

## Additional Notes
- We will always send the workflow progress
- We only update the job status after the response from the Engine


packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -31,7 +37,31 @@ export const flowExecutor = {
...
+    const newContext =
+      output.verdict === ExecutionVerdict.RUNNING
+        ? output.setVerdict(ExecutionVerdict.SUCCEEDED, output.verdictResponse)
+        : output;
+
+    sendProgress(newContext, constants);
+
+    return newContext.toResponse();
+  },
+  async executeFromAction({
```
MarceloRGonc, line 64: Rename the execute to executeFromAction

packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -73,29 +103,31 @@ export const flowExecutor = {
         duration: stepEndTime - stepStartTime,
       });
 
+      sendProgress(flowExecutionContext, constants);
+
       if (flowExecutionContext.verdict !== ExecutionVerdict.RUNNING) {
         break;
       }
```
MarceloRGonc, line 110: If different from Running we don't want to continue the workflow execution

packages/engine/src/lib/handler/flow-executor.ts
rita-gorokhod: any particular reason?

MarceloRGonc: Just "execute" after having the new method. I don't think it's clear enough.

packages/engine/src/lib/operations.ts
```diff
@@ -31,28 +31,20 @@ import { testExecutionContext } from './handler/context/test-execution-context';
...
 
 const executeFlow = async (
   input: ExecuteFlowOperation,
   context: FlowExecutorContext,
 ): Promise<EngineResponse<Pick<FlowRunResponse, 'status' | 'error'>>> => {
   const constants = EngineConstants.fromExecuteFlowInput(input);
-  const output = await flowExecutor.execute({
-    action: input.flowVersion.trigger.nextAction,
+
+  const response = await flowExecutor.triggerFlowExecutor({
```
rita-gorokhod, line 42: 1. I think we should keep it as "execute", as before
2. move 
```
await progressService.sendUpdate({
    engineConstants: constants,
    flowExecutorContext: newContext,
  });
```
to be there always
3. this should then just be
```
  const newContext =
    output.verdict === ExecutionVerdict.RUNNING
      ? output.setVerdict(ExecutionVerdict.SUCCEEDED, output.verdictResponse)
      : output;
  const response = await newContext.toResponse();
```

wdyt?

packages/server/worker/src/lib/executors/flow-job-executor.ts
```diff
@@ -196,6 +205,18 @@ async function handleInternalError(
...
+  status: JobStatus = JobStatus.COMPLETED,
+  message = 'Flow succeeded',
+): Promise<void> {
+  await engineApiService(engineToken).updateJobStatus({
+    status,
+    message,
+    queueName: QueueName.ONE_TIME,
+  });
+}
+
```
MarceloRGonc, line 219: After the engine execution, we need to update the job status. To be moved from the QUEUE to the complete or failed state.

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -155,13 +155,6 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
...
       .to(populatedRun.projectId)
       .emit(WebsocketClientEvent.TEST_FLOW_RUN_PROGRESS, populatedRun);
-
-    await markJobAsCompleted(
-      populatedRun.status,
-      executionCorrelationId,
-      request.principal as unknown as EnginePrincipal,
-      runDetails.error,
-    );
   });
```
MarceloRGonc, line 158: We are sending progress updates, but this does not mean that the job is completed. So, I'm removing it from here and add it to the end of the engine execution

packages/engine/src/lib/operations.ts
MarceloRGonc: Unfortunately, this doesn't work. The execute can be recursive. So we will miss the last update, and for example, the loop execution does not resume.

image

image

packages/server/api/src/app/workers/engine-controller.ts
rita-gorokhod: Can remove also the method markJobAsCompleted

rita-gorokhod (approved):

PR #904, opened 2025-07-02
Add path to pause metadata

MarceloRGonc:
Part of OPS-2071.

## Additional notes
- When resuming in a loop, we need the path to be filled to  be able to fetch the right iteration


packages/engine/src/lib/handler/block-executor.ts
```diff
@@ -293,10 +298,13 @@ function createPauseHook(
       ).executionCorrelationId;
     }
 
+    const path = req.pauseMetadata?.path ?? currentExecutionPath;
```
rita-gorokhod, line 301: does it make sense to take it from the previous pause, and not the current?

rita-gorokhod (approved):

packages/engine/src/lib/handler/block-executor.ts
MarceloRGonc: It's not from the previous. It's the object that we are currently passing. If, for some reason, we want to pass the path.

PR #905, opened 2025-07-02
Preserve empty values when redacting secrets

cezudas:
Fixes OPS-1979.

### DIFF
--- a/packages/server/api/test/unit/app-connection/app-connection-utils.test.ts
+++ b/packages/server/api/test/unit/app-connection/app-connection-utils.test.ts
@@ -292,70 +292,51 @@ describe('redactSecrets', () => {
     },
   );
 
-  test('should return undefined value when OAUTH2 client_secret is null', () => {
-    const connection: AppConnection = {
-      ...baseConnection,
-      type: AppConnectionType.OAUTH2,
-      value: {
+  test.each([
+    { value: null, description: 'null', expectUndefined: true },
+    { value: undefined, description: 'undefined', expectUndefined: true },
+    { value: '', description: 'empty string', expectUndefined: false },
+  ])(
+    'should handle OAUTH2 client_secret when $description',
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/app-connection/app-connection-utils.ts
```diff
@@ -62,12 +69,15 @@ export function redactSecrets(
         typeof redacted.client_id === 'string' &&
         typeof redacted.redirect_url === 'string'
       ) {
-        return {
-          type: AppConnectionType.OAUTH2,
-          client_id: redacted.client_id,
-          client_secret: REDACTED_MESSAGE,
```
rita-gorokhod, line 68: minor but in this case I think it might be easier to read
```
clientSecret: redactIfNotNilOrEmpty(redacted.client_secret)
```

packages/server/api/test/unit/app-connection/app-connection-utils.test.ts
```diff
@@ -188,6 +188,271 @@ describe('redactSecrets', () => {
     );
     expect(result).toEqual(undefined);
   });
+
+  test('should return original value when SECRET_TEXT is null', () => {
```
rita-gorokhod: use test cases for null undefined and ''

packages/server/api/test/unit/app-connection/app-connection-utils.test.ts
```diff
@@ -188,6 +188,271 @@ describe('redactSecrets', () => {
...
+      displayName: 'Secret',
+      valueSchema: 'some schema',
+      required: true,
+    };
+
+    const result = redactSecrets(auth, connection.value);
+    expect(result?.secret_text).toBe('');
+  });
+
+  test('should return original value when BASIC_AUTH password is null', () => {
```
rita-gorokhod: same comment as above for all the properties

PR #908, opened 2025-07-03
Update @frontegg/js and use ensure cleanup operations after logout are completed

cezudas:
Fixes OPS-1358.

The bug does not reproduce consistently, and I suspect it's a bug in the client(browser) state.
Yes, there is a redux store under the hood ( see [cloud-connection-page](https://github.com/openops-cloud/openops/blob/cezudas/OPS-1358/packages/react-ui/src/app/routes/cloud-connection/cloud-connection-page.tsx#L51))
```
app.store.subscribe(() => {
  const { auth } = app.store.getState();
  // ... you then use auth.isLoading, auth.isAuthenticated, etc.
});
```

I expect the problem to be solved by [updating @frontegg/js to the latest](https://github.com/openops-cloud/openops/pull/908#discussion_r2182407578), but we need to merge and deploy on `app.openops.com` to actually validate it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "@fastify/swagger-ui": "5.2.3",
     "@fastify/type-provider-typebox": "5.1.0",
     "@frontegg/client": "5.3.2",
-    "@frontegg/js": "7.56.0",
+    "@frontegg/js": "7.79.0",
     "@hookform/resolvers": "3.9.0",
     "@microsoft/microsoft-graph-client": "3.0.7",
     "@microsoft/microsoft-graph-types": "2.40.0",
--- a/packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
+++ b/packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
@@ -1,14 +1,15 @@
 import { flagsHooks } from '@/app/common/hooks/flags-hooks';
+import { FronteggApp } from '@frontegg/js';
 import Cookies from 'js-cookie';
 import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { initializeFrontegg } from './frontegg-setup';
 
-const handleAppReady = (app: any) => {
+const handleAppReady = (app: FronteggApp) => {
   app.ready(() => handleLogout(app));
 };
 
-const handleLogout = (app: any) => {
+const handleLogout = (app: FronteggApp) => {
   app.logout(handleWindowClose);
 };
 
--- a/packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
+++ b/packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
@@ -4,6 +4,23 @@ import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { initializeFrontegg } from './frontegg-setup';
 
+const handleLogout = (app: any) => {
+  app.logout(handleWindowClose);
+};
+
+const handleAppReady = (app: any) => {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "@fastify/swagger-ui": "5.2.3",
     "@fastify/type-provider-typebox": "5.1.0",
     "@frontegg/client": "5.3.2",
-    "@frontegg/js": "7.56.0",
+    "@frontegg/js": "7.79.0",
```
cezudas, line 79: @frontegg/js does not reference any public GitHub repository, and npm releases do not contain a changelog.
Thankfully, the npm package contains the non-obfuscated code, and I was able to do a deep diff between the currently installed and the latest versions.

1. More Robust State Reset on Logout (resetAuthState)

In node_modules/@frontegg/redux-store/auth/index.js:
Old (7.56.0): const resetAuthState = () => { deepResetState(store, ['auth'], snapshotAuthState); }
New (7.79.0): const resetAuthState = (state = {}) => { deepResetState(store, ['auth'], _extends({}, snapshotAuthState, state)); }
Analysis:
A more thorough state reset on logout would ensure that no remnants of the previous user's session (like tokens or user info) persist. This clean state would force the application to initiate a fresh login flow with Google, correctly including the prompt=select_account parameter in the OAuth request, which is what triggers the account chooser.

2. Enhancements to the logout Function

In node_modules/@frontegg/js/FronteggApp/FronteggApp.d.ts:
Old (7.56.0): logout(): void;
New (7.79.0): logout(callback?: () => void): void;
Analysis:
The addition of a callback function to logout suggests that the logout process might now involve asynchronous cleanup operations. Adding a callback provides a more reliable way to ensure all cleanup tasks (including state reset and session invalidation) are fully completed before any subsequent actions are taken.

packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
```diff
@@ -1,9 +1,27 @@
...
 import { useEffect } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { initializeFrontegg } from './frontegg-setup';
 
+const handleAppReady = (app: FronteggApp) => {
+  app.ready(() => handleLogout(app));
+};
+
+const handleLogout = (app: FronteggApp) => {
+  app.logout(handleWindowClose);
```
cezudas, line 13: Making sure all internal cleanup operations are complete before closing the window.
This optional callback was not available in @frontegg/js@7.56.0

packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
```diff
@@ -28,15 +46,7 @@ const CloudLogoutPage = () => {
...
-      app.logout();
-
-      if (window.opener) {
-        setTimeout(() => {
-          window.close();
-        }, 300);
-      }
-    });
+    handleAppReady(app);
   }, [flags, isLoading, navigate]);
```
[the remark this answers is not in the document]
cezudas: nope, check line 36, it is used in useffect

packages/react-ui/src/app/routes/cloud-connection/cloud-logout-page.tsx
```diff
@@ -28,15 +46,7 @@ const CloudLogoutPage = () => {
     Cookies.remove('cloud-token');
     Cookies.remove('cloud-refresh-token');
 
-    app.ready(() => {
```
cezudas, line 31: I had to factor this simply because Sonarqube was complaining about increasing the code nesting :( 

> [Refactor this code to not nest functions more than 4 levels deep.]

PR #911, opened 2025-07-03
Decrease the TTL for the openops tables token

MarceloRGonc:
Fixes OPS-2081.

## Additional Notes
- Decreasing the TTL for the token in the cache
- Adding request time in the errors

### DIFF
--- a/packages/server/api/src/app/app-connection/app-connection-utils.ts
+++ b/packages/server/api/src/app/app-connection/app-connection-utils.ts
@@ -5,9 +5,22 @@ import {
   AppConnectionType,
   AppConnectionValue,
   AppConnectionWithoutSensitiveData,
+  isNil,
 } from '@openops/shared';
 
 const REDACTED_MESSAGE = '**REDACTED**';
+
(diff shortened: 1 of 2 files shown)


packages/openops/src/lib/openops-tables/auth-user.ts
```diff
@@ -7,7 +7,7 @@ const tokenLifetimeMinutes = system.getNumber(
 );
 
 const tokenLifetimeSeconds = tokenLifetimeMinutes
-  ? tokenLifetimeMinutes * 60 - 60 // Subtract 60 seconds to ensure the cache expired before the token
+  ? (tokenLifetimeMinutes - 10) * 60 // Subtract 10 minutes to ensure the cache expired before the token
```
rita-gorokhod, line 10: why are you not changing the env variable?

MarceloRGonc: The environment variable is also used in the tables container. So If I decrease the value it will be decreased also in the token

rita-gorokhod (approved):

PR #912, opened 2025-07-03
Use correct jobStatus when there is an internal error

MarceloRGonc:
Part of OPS-2071.


packages/server/worker/src/lib/executors/flow-job-executor.ts
```diff
@@ -140,6 +140,10 @@ async function executeFlow(
       await handleTimeoutError(jobData, engineToken);
```
rita-gorokhod, line 140: as we discussed offline, timeouts are currently marked as successful in the queue. We're not changing it for now.

PR #917, opened 2025-07-04
Make request body limit configurable and increase default to 10MB

cezudas:
Fixes OPS-2031.


alexandrudanpop (approved):

packages/engine/src/api-handler.ts
```diff
@@ -11,8 +11,16 @@ import { StatusCodes } from 'http-status-codes';
...
 
+const isLambdaEnvironment = (): boolean => {
+  return process.env.AWS_LAMBDA_FUNCTION_NAME !== undefined;
+};
+
+const LAMBDA_FASTIFY_BODY_LIMIT = 6 * 1024 * 1024; // 6MB
+const DEFAULT_FASTIFY_BODY_LIMIT = 10 * 1024 * 1024; // 10MB
 const app = fastify({
-  bodyLimit: 6 * 1024 * 1024, // 6MB same as engine api-handler & nginx.conf
+  bodyLimit: isLambdaEnvironment()
```
rita-gorokhod: use an env var

alexandrudanpop: so if it is an existing installation, and we add an env var, will it be active in that installation? Asking because I think the template .env is only copied when you first install.

cezudas: Indeed, even if the env variable is not available I will fall back to 10MB.  And I think it will cover the that scenario.

docker-entrypoint.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/sh
 
+if [ -n "$OPS_FASTIFY_BODY_LIMIT_MB" ]; then
+  export NGINX_BODY_LIMIT="${OPS_FASTIFY_BODY_LIMIT_MB}m"
+else
+  export NGINX_BODY_LIMIT="10m"  # Default fallback
+fi
```
maor-rozenfeld:
```suggestion
  export NGINX_BODY_LIMIT="${OPS_FASTIFY_BODY_LIMIT_MB:-10}m"
```

maor-rozenfeld (commented):
See the problem of calling it "FASTIFY" and then using it to configure nginx?

docker-entrypoint.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+export NGINX_CLIENT_MAX_BODY_SIZE="${OPS_REQUEST_BODY_LIMIT:-10}m"
+
+# Generate nginx config from template
+envsubst '${NGINX_CLIENT_MAX_BODY_SIZE}' < /etc/nginx/nginx.template.conf > /etc/nginx/nginx.conf
```
cezudas, line 6: [Subtitute only NGINX_CLIENT_MAX_BODY_SIZE to prevent map directive from getting broken](https://github.com/openops-cloud/openops/pull/917/commits/19aa1560a4db464aa02bea15674d0e19f24d18bd)

PR #918, opened 2025-07-04
Add generate with AI to code block

alexandrudanpop:
Fixes OPS-2092

## Testing Checklist

Check all that apply:

Deployed on UX

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,7 @@ COPY --link ai-prompts ai-prompts
 LABEL service=openops
 
 # Copy Nginx configuration template and static files
-COPY nginx.standard.conf /etc/nginx/nginx.conf
+COPY nginx.template.conf /etc/nginx/nginx.template.conf
 COPY dist/packages/react-ui/ /usr/share/nginx/html/
 
 ARG VERSION=unknown
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+export NGINX_CLIENT_MAX_BODY_SIZE="${OPS_REQUEST_BODY_LIMIT:-10}m"
+
+# Generate nginx config from template
+envsubst '${NGINX_CLIENT_MAX_BODY_SIZE}' < /etc/nginx/nginx.template.conf > /etc/nginx/nginx.conf
+
 # Start Nginx server
 nginx -g "daemon off;" &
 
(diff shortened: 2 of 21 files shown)


packages/react-ui/src/app/features/builder/step-settings/hooks/use-apply-code-to-inject.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    switch (midpanelState.aiChatProperty.inputName) {
+      case 'settings.sourceCode':
+        form.setValue(
+          midpanelState.aiChatProperty.inputName,
+          {
+            code: midpanelState.codeToInject,
+            packageJson: '{}',
+          },
```
alexandrudanpop: Updating packageJson TBD after the API is created. Probably need to use structured output.

packages/server/api/src/app/flags/dev-flags.service.ts
```diff
@@ -20,7 +20,7 @@ async function getAll(): Promise<Flag[]> {
   flags = [
     {
       id: FlagId.CODE_WITH_AI,
-      value: false,
+      value: true,
```
alexandrudanpop: to reset

packages/react-ui/src/app/features/builder/ai-chat/lib/step-settings-ai-chat-hook.ts
```diff
@@ -125,3 +132,26 @@ export const useStepSettingsAiChat = (
     isEmpty: !messages.length,
   };
 };
+
+const CODE_BLOCK_NAME = '@openops/code';
+const CODE_ACTION_NAME = 'code';
```
alexandrudanpop, line 137: We don't have a block and an action name because Code is a frontend-defined block (like switch/branch)

packages/react-ui/src/app/features/builder/step-settings/hooks/use-apply-code-to-inject.ts
rmn-snpk: Can we keep existing `packageJson` instead of replacing it with empty object?

alexandrudanpop: the idea is we will generate it with LLM (OPS-2103), so it will be replaced if the user uses the Inject Code command. If the user has some code already that depends on a package I think it's fine to overwrite it since the new code might not depend on that package.

cezudas (approved):

PR #922, opened 2025-07-07
Add variable resolution endpoint

alexandrudanpop:
Fixes OPS-2108

Simple Proxy API so we can resolve a variable in a flow, that can be exposed later as MCP

image

Resolving sample output data:
image

Without Auth header
image

### DIFF
--- a/packages/server/api/test/integration/cloud/block-variable/block-variable-module.test.ts
+++ b/packages/server/api/test/integration/cloud/block-variable/block-variable-module.test.ts
@@ -6,13 +6,26 @@ jest.mock('@openops/common', () => commonMock);
 
 const flowServiceMock = {
   getOnePopulatedOrThrow: jest.fn(),
+  getAllEnabled: jest.fn(),
 };
 jest.mock('../../../../src/app/flows/flow/flow.service', () => ({
   flowService: flowServiceMock,
 }));
 
+const flowVersionServiceMock = {
+  getOneOrThrow: jest.fn(),
+};
+jest.mock(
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/block-variable/block-variable-module.ts
```diff
@@ -0,0 +1,101 @@
...
+      projectId,
+      stepName: request.stepName ?? '',
+      stepTestOutputs,
+    });
+
+    return {
+      success: result.success,
+      value: result.censoredValue,
+      error: result.error,
+    };
```
alexandrudanpop, line 55: th enginer returnes both `resolvedValue` and `censoredValue`, so we just return the censored one

packages/server/api/src/app/block-variable/block-variable-module.ts
```diff
@@ -0,0 +1,101 @@
...
+      "Execute variable resolution to get the resolved value of a variable expression within a flow context. This endpoint resolves variables like {{trigger['headers']}} or {{step_2}} and returns the resolved value.",
+    body: Type.Object({
+      flowId: Type.String({
+        description: 'The ID of the flow containing the variable context',
+      }),
+      flowVersionId: Type.String({
+        description: 'The specific version ID of the flow',
+      }),
+      variableExpression: Type.String({
+        description:
```
cezudas, line 75: maybe adding reasonable limits for this input, e.g. `minLength: 1, maxLength: 10000`

alexandrudanpop: I somehow missed this comment, will fix

PR #923, opened 2025-07-07
Fix engine progress calls

alexandrudanpop:
Fixes OPS-2109

## Additional Notes

Add a retry with backoff to the call

### DIFF
--- a/packages/engine/src/lib/handler/flow-executor.ts
+++ b/packages/engine/src/lib/handler/flow-executor.ts
@@ -57,7 +57,14 @@ export const flowExecutor = {
         ? output.setVerdict(ExecutionVerdict.SUCCEEDED, output.verdictResponse)
         : output;
 
-    sendProgress(newContext, constants);
+    try {
+      await progressService.sendUpdate({
+        engineConstants: constants,
+        flowExecutorContext: newContext,
+      });
+    } catch (error) {
+      logger.error('CRITICAL: Final progress update failed', error);
+    }
 
     return newContext.toResponse();
   },
--- a/packages/engine/src/lib/handler/flow-executor.ts
+++ b/packages/engine/src/lib/handler/flow-executor.ts
@@ -106,7 +106,7 @@ export const flowExecutor = {
         duration: stepEndTime - stepStartTime,
       });
 
-      sendProgress(flowExecutionContext, constants);
+      await sendProgress(flowExecutionContext, constants);
 
       if (flowExecutionContext.verdict !== ExecutionVerdict.RUNNING) {
         break;
@@ -124,13 +124,9 @@ export const flowExecutor = {
 function sendProgress(
   flowExecutionContext: FlowExecutorContext,
   constants: EngineConstants,
-): void {
-  progressService
-    .sendUpdate({
-      engineConstants: constants,
-      flowExecutorContext: flowExecutionContext,
-    })
-    .catch((error) => {
-      logger.error('Error sending progress update', error);
-    });
+): Promise<void> {
+  return progressService.sendUpdate({
+    engineConstants: constants,
+    flowExecutorContext: flowExecutionContext,
+  });
 }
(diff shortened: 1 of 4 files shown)


packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -57,7 +57,14 @@ export const flowExecutor = {
         ? output.setVerdict(ExecutionVerdict.SUCCEEDED, output.verdictResponse)
         : output;
 
-    sendProgress(newContext, constants);
+    try {
```
rita-gorokhod: why not do it inside the sendProgress method?

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -1,20 +1,17 @@
...
 let lastRequestHash: string | undefined = undefined;
 
-const lock = new Mutex();
-
 export const progressService = {
   sendUpdate: async (params: UpdateStepProgressParams): Promise<void> => {
     throwIfExecutionTimeExceeded();
-    return lock.runExclusive(async () => {
-      await sendUpdateRunRequest(params);
-    });
```
alexandrudanpop: fetch() can hang with no timeout so and next requests to update the progress would be blocked by the lock.

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -1,20 +1,17 @@
...
 import { throwIfExecutionTimeExceeded } from '../timeout-validator';
 
 let lastRequestHash: string | undefined = undefined;
 
-const lock = new Mutex();
-
 export const progressService = {
   sendUpdate: async (params: UpdateStepProgressParams): Promise<void> => {
     throwIfExecutionTimeExceeded();
-    return lock.runExclusive(async () => {
```
rita-gorokhod: any reason you removed it?

packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -57,7 +57,10 @@ export const flowExecutor = {
         ? output.setVerdict(ExecutionVerdict.SUCCEEDED, output.verdictResponse)
         : output;
 
-    sendProgress(newContext, constants);
+    await progressService.sendUpdate({
```
rita-gorokhod: use await sendProgress

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -1,18 +1,21 @@
...
+const MAX_RETRIES = 3;
 
+let lastRequestHash: string | undefined = undefined;
 const lock = new Mutex();
 
 export const progressService = {
   sendUpdate: async (params: UpdateStepProgressParams): Promise<void> => {
-    throwIfExecutionTimeExceeded();
     return lock.runExclusive(async () => {
+      throwIfExecutionTimeExceeded();
```
alexandrudanpop, line 18: todo

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -54,14 +61,28 @@ const sendUpdateRunRequest = async (
...
+        retries: MAX_RETRIES,
+        retryDelay: (retryCount: number) => {
+          return (retryCount + 1) * 1000; // 1s, 2s, 3s
+        },
+      },
+    );
+  } catch (error) {
+    logger.error(
+      `Progress update failed after ${MAX_RETRIES} retries for status ${request.runDetails.status} on run ${request.runId}`,
+      error,
```
rita-gorokhod: needs to be { error } as an object

alexandrudanpop: done

packages/engine/src/lib/handler/flow-executor.ts
alexandrudanpop: done

PR #934, opened 2025-07-09
Add context enrichment service 

alexandrudanpop:
Fixes OPS-2097

## Additional Notes
Uses the new variable resolution in the engine & truncates when output is larger than 1000 characters. New messages in the conversation endpoint will receive this context from the frontend. This service can be extended to return step outputs in cases where variables are not specified.


packages/server/api/src/app/ai/chat/context-enrichment.service.ts
```diff
@@ -0,0 +1,250 @@
...
+          }
+          return val;
+        },
+        2,
+      );
+    } else {
+      stringValue = String(value);
+    }
+  } catch (error) {
+    stringValue = String(value);
```
cezudas: I think it's best to log the error

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
```diff
@@ -0,0 +1,250 @@
...
+      flowVersionId: additionalContext.flowVersionId,
+      steps: enrichedSteps,
+    };
+  } catch (error) {
+    logger.error('Failed to enrich context', { error });
+    throw error;
+  }
+}
+
+function truncateForLLM(value: unknown, maxLength = 1000): string {
```
cezudas: If you think we might use it somewhere else then maybe it's worth exactring a shared utility and give it a more explicit name like "safeStringifyAndTruncate". If not, we can keep `truncateForLLM`.

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
```diff
@@ -0,0 +1,250 @@
...
+    flowVersion: flowData.flow.version,
+    projectId,
+    stepName: step.stepName,
+    stepTestOutputs: flowData.stepTestOutputs,
+  });
+
+  return {
+    name: variable.name,
+    value: result.success
+      ? truncateForLLM(result.censoredValue)
```
cezudas: Test coverage would be good for `truncateForLLM`

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
alexandrudanpop: done

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
alexandrudanpop: extracted and added tests

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
alexandrudanpop: done, but to be have a safe function we just return an empty string

cezudas (approved):

PR #941, opened 2025-07-09
Prevent reconnect popup from losing pending form changes on tab refocus

cezudas:
Fixes OPS-2126.

### DIFF
--- a/packages/react-ui/src/app/features/builder/step-settings/block-settings/connection-select.tsx
+++ b/packages/react-ui/src/app/features/builder/step-settings/block-settings/connection-select.tsx
@@ -59,7 +59,7 @@ const ConnectionSelect = memo((params: ConnectionSelectProps) => {
     limit: FETCH_ALL_CONNECTIONS_LIMIT,
   });
 
-  const { data: reconnectConnection, isFetching: isFetchingConnection } =
+  const { data: reconnectConnection, isLoading: isLoadingConnection } =
     appConnectionsHooks.useConnection({
       id: reconnectConnectionId,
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/app/features/builder/step-settings/block-settings/connection-select.tsx
```diff
@@ -59,7 +59,7 @@ const ConnectionSelect = memo((params: ConnectionSelectProps) => {
     limit: FETCH_ALL_CONNECTIONS_LIMIT,
   });
 
-  const { data: reconnectConnection, isFetching: isFetchingConnection } =
+  const { data: reconnectConnection, isLoading: isLoadingConnection } =
```
cezudas, line 62: isLoading is `true` only when the first fetch for a query is in-flight, Is the same as `isFetching && isPending`.

packages/react-ui/src/app/features/builder/step-settings/block-settings/connection-select.tsx
```diff
@@ -109,7 +109,7 @@ const ConnectionSelect = memo((params: ConnectionSelectProps) => {
               inputName="settings.input.auth"
               allowDynamicValues={!params.isTrigger}
             >
-              {connectionDialogOpen && !isFetchingConnection && (
+              {connectionDialogOpen && !isLoadingConnection && (
```
cezudas, line 112: upon tab re-focus `isFetchingConnection` transitions from `false` to `true`, the component will be unmounted.

PR #958, opened 2025-07-11
Fix loop output

rSnapkoOpenOps:
Fixes OPS-2146

https://www.loom.com/share/26acaa024bb149d5809dee47ce71dec8?sid=b0f16c5c-668d-48af-8eca-a2382e2f920f

### DIFF
--- a/packages/react-ui/src/app/features/builder/run-details/index.tsx
+++ b/packages/react-ui/src/app/features/builder/run-details/index.tsx
@@ -13,13 +13,7 @@ import React, { useMemo } from 'react';
 
 import { flagsHooks } from '@/app/common/hooks/flags-hooks';
 import { useBuilderStateContext } from '@/app/features/builder/builder-hooks';
-import {
-  ActionType,
-  FlagId,
-  isNil,
-  RunEnvironment,
-  StepOutput,
-} from '@openops/shared';
+import { FlagId, isNil, RunEnvironment } from '@openops/shared';
 import { LeftSideBarType } from '../builder-types';
 
 import { useEffectOnce } from 'react-use';
@@ -57,37 +51,14 @@ const FlowRunDetails = React.memo(() => {
   });
 
   const selectedStepTestData = useMemo(() => {
-    const testData =
-      run && selectedStep && run.steps
-        ? flowRunUtils.extractStepOutput(
-            selectedStep,
-            loopsIndexes,
-            run.steps,
-            flowVersion.trigger,
-          )
-        : null;
-    if (
-      !testData ||
-      testData.type !== ActionType.LOOP_ON_ITEMS ||
-      !testData.output?.index ||
-      !testData.input ||
-      !selectedStep
-    ) {
-      return testData;
-    }
-    try {
-      return {
-        ...testData,
-        output: {
-          item: (testData.input as { items: any }).items?.[
-            loopsIndexes[selectedStep]
-          ],
-          index: loopsIndexes[selectedStep] + 1,
-        },
-      } as StepOutput;
-    } catch {
-      return testData;
-    }
+    return run && selectedStep && run.steps
+      ? flowRunUtils.extractLoopItemStepOutput(
+          selectedStep,
+          loopsIndexes,
+          run.steps,
+          flowVersion.trigger,
+        )
+      : null;
   }, [run, selectedStep, loopsIndexes, flowVersion.trigger]);
 
   const message = getRunMessage(run, rententionDays);
--- a/packages/react-ui/src/app/features/builder/run-details/index.tsx
+++ b/packages/react-ui/src/app/features/builder/run-details/index.tsx
@@ -13,7 +13,13 @@ import React, { useMemo } from 'react';
 
 import { flagsHooks } from '@/app/common/hooks/flags-hooks';
 import { useBuilderStateContext } from '@/app/features/builder/builder-hooks';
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/app/features/builder/run-details/index.tsx
```diff
@@ -50,15 +56,38 @@ const FlowRunDetails = React.memo(() => {
...
+            selectedStep,
+            loopsIndexes,
+            run.steps,
+            flowVersion.trigger,
+          )
+        : null;
+    if (
+      !testData ||
+      testData.type !== ActionType.LOOP_ON_ITEMS ||
+      !testData.output?.index ||
```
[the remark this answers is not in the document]
rmn-snpk: In the case of watches, it's not a real index, but a sequence number, so it is fine.

packages/react-ui/src/app/features/builder/run-details/index.tsx
```diff
@@ -50,15 +56,38 @@ const FlowRunDetails = React.memo(() => {
...
+        ...testData,
+        output: {
+          item: (testData.input as { items: any }).items?.[
+            loopsIndexes[selectedStep]
+          ],
+          index: loopsIndexes[selectedStep] + 1,
+        },
+      } as StepOutput;
+    } catch {
+      return testData;
```
alexandrudanpop: can we extract the function and add unit tests for it? Also, could we avoid the try-catch?

packages/react-ui/src/app/features/builder/run-details/index.tsx
```diff
@@ -50,15 +56,38 @@ const FlowRunDetails = React.memo(() => {
...
-      ? flowRunUtils.extractStepOutput(
-          selectedStep,
-          loopsIndexes,
-          run.steps,
-          flowVersion.trigger,
-        )
-      : null;
+  const selectedStepTestData = useMemo(() => {
+    const testData =
+      run && selectedStep && run.steps
```
rita-gorokhod: how can run/run.steps be undefined, with selectedStep defined?

packages/react-ui/src/app/features/builder/run-details/index.tsx
```diff
@@ -50,15 +56,38 @@ const FlowRunDetails = React.memo(() => {
...
+            run.steps,
+            flowVersion.trigger,
+          )
+        : null;
+    if (
+      !testData ||
+      testData.type !== ActionType.LOOP_ON_ITEMS ||
+      !testData.output?.index ||
+      !testData.input ||
+      !selectedStep
```
rita-gorokhod: no need to check again for selectedStep

packages/react-ui/src/app/features/builder/run-details/index.tsx
```diff
@@ -50,15 +56,38 @@ const FlowRunDetails = React.memo(() => {
...
+      return testData;
+    }
+    try {
+      return {
+        ...testData,
+        output: {
+          item: (testData.input as { items: any }).items?.[
+            loopsIndexes[selectedStep]
+          ],
+          index: loopsIndexes[selectedStep] + 1,
```
rita-gorokhod: is there a race condition risk, i.e. that if my input is
[a, b, c]
the actual execution finished could be
[b, c, a]

packages/react-ui/src/app/features/builder/run-details/index.tsx
rmn-snpk: For ex right after clicking on test 

Image

packages/react-ui/src/app/features/builder/run-details/index.tsx
rmn-snpk: Yes, but it helps to avoid this error 

TS2538: Type null cannot be used as an index type.

`loopsIndexes[selectedStep]` 

or this warning 
ESLint: Forbidden non-null assertion. (@typescript-eslint/no-non-null-assertion)

`loopsIndexes[selectedStep!]`

packages/react-ui/src/app/features/builder/run-details/index.tsx
rmn-snpk: I tested it with code block

During executions we get iterations one by one

Image

Execution time doesn't matter, we still getting correct order 

Image

packages/react-ui/src/app/features/builder/run-details/index.tsx
rmn-snpk: Extracted! We cannot avoid using try catch as there can be not iterable data as input

packages/react-ui/src/app/features/flow-runs/lib/tests/flow-run-utils.test.ts
```diff
@@ -492,7 +492,139 @@ describe('flowRunUtils.findLoopsState', () => {
       run,
       currentLoopsState,
     );
-    // Since the failed step is the parent loop, should just return the current state
     expect(result).toEqual({ loop1: 1 });
   });
 });
+
+describe('flowRunUtils.extractLoopItemStepOutput', () => {
```
alexandrudanpop, line 499: please add  a test case for loopIndex 2, that the output is correct

rmn-snpk: done

alexandrudanpop (approved):

PR #962, opened 2025-07-14
Add `success` field to `FlowStepTestOutput` and related logic

rSnapkoOpenOps:
Part of OPS-2154.

### DIFF
--- a/packages/server/api/src/app/flows/flow/flow-version.controller.ts
+++ b/packages/server/api/src/app/flows/flow/flow-version.controller.ts
@@ -153,6 +153,7 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
             input: flowStepTestOutput.input,
             output: flowStepTestOutput.output,
             lastTestDate: flowStepTestOutput.updated,
+            success: flowStepTestOutput.success,
           },
         ]),
       );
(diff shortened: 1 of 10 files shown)


packages/server/api/src/app/flows/trigger-events/trigger-event.service.ts
```diff
@@ -63,6 +63,7 @@ export const triggerEventService = {
         flowVersionId: flow.version.id,
         output: triggerEventToSave.payload,
         input,
+        success: true,
```
rmn-snpk, line 66: We don't save failed trigger outputs

packages/server/api/src/app/flows/flow/flow-version.controller.ts
```diff
@@ -194,7 +196,7 @@ export const flowVersionController: FastifyPluginAsyncTypebox = async (
     },
     async (request, reply) => {
       const { flowVersionId } = request.params;
-      const { stepId, input, output } = request.body;
+      const { stepId, input, output, success } = request.body;
```
rita-gorokhod, line 199: where are we sending this?

rmn-snpk: Seems like we don't use POST request

rmn-snpk: https://linear.app/openops/issue/OPS-2163/remove-unused-endpoints-from-flow-versioncontroller

MarceloRGonc: I think it will be used by the agent. I removed it, and Leyla added it back.

PR #970, opened 2025-07-16
Add resizable panel for AI chat 

rSnapkoOpenOps:
Fixes OPS-2165.

## Additional Notes

I updated the way we save panel sizes. Based on existing logic we care only about left side bar. Now we save sizes for each panel separately by panel id. 

Loom to show that existing behavior is not broken: https://www.loom.com/share/2c055bca2acb4c3e9bf27d5291cc2ada?sid=d1c45257-d016-42c8-965b-0b1317879bfe

It can be checked on UX 
image


packages/server/api/src/app/flags/dev-flags.service.ts
```diff
@@ -24,6 +24,12 @@ async function getAll(): Promise<Flag[]> {
       created,
       updated,
     },
+    {
+      id: FlagId.SHOW_CHAT_IN_RESIZABLE_PANEL,
+      value: true,
```
alexandrudanpop: reset flag

PR #973, opened 2025-07-16
Fix workflow progress when using loops

MarceloRGonc:
Fixes OPS-2175.

## Additional Notes
- We are no longer running iterations in parallel. So we are removing the environment variable
- We increment the loop state with each iteration, so the progress is sent with the current state


packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -149,25 +154,29 @@ function triggerLoopIterations(
...
-        action: firstLoopAction,
-        constants,
-      });
+    loopExecutionState = await flowExecutor.executeFromAction({
+      executionState: loopExecutionState,
+      action: firstLoopAction,
+      constants,
+    });
+
+    loopIterations[i] = cloneDeep(loopExecutionState);
```
MarceloRGonc, line 168: Using deep clone to save only the state up to the current iteration.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -149,25 +154,29 @@ function triggerLoopIterations(
...
+    loopExecutionState = loopExecutionState
+      .setVerdict(ExecutionVerdict.RUNNING)
+      .setCurrentPath(loopExecutionState.currentPath.removeLast())
+      .setPauseId(originalPauseId);
```
MarceloRGonc, line 172: Restoring the execution state to the original one (path, verdict and the pauseId)

packages/server/shared/src/lib/system/system-prop.ts
```diff
@@ -139,8 +139,6 @@ export enum SharedSystemProp {
   INTERNAL_OAUTH_PROXY_URL = 'INTERNAL_OAUTH_PROXY_URL',
   INTERNAL_BACKEND_PUBLIC_URL = 'INTERNAL_BACKEND_PUBLIC_URL',
 
-  INTERNAL_PARALLEL_LOOP_ITERATIONS_LIMIT = 'INTERNAL_PARALLEL_LOOP_ITERATIONS_LIMIT',
-
```
MarceloRGonc, line 143: We are no longer running iterations in parallel.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -149,25 +154,29 @@ function triggerLoopIterations(
 
     // Generate new pauseId for each iteration
     const newId = nanoid();
-    const newExecutionContext = loopExecutionState
+    loopExecutionState = loopExecutionState
```
MarceloRGonc, line 157: The 'loopExecutionState' has all the execution states. So, as the iterations are completed, the state changes. In the end, we have the complete state.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -203,25 +205,18 @@ async function waitForIterationsToFinishOrPause(
...
 
-  const { iterationContext: lastIterationContext } =
-    iterationResults[iterationResults.length - 1];
-
-  populateLastIterationContext(lastIterationContext, iterationResults);
-
-  const executionState = lastIterationContext.setCurrentPath(
-    lastIterationContext.currentPath.removeLast(),
-  );
-
```
MarceloRGonc, line 214: Since we have the complete state, it is not necessary to collect the states of all iterations.

rita-gorokhod (approved):

PR #977, opened 2025-07-17
Add new Umbrella recommendation types

MarceloRGonc:
Fixes OPS-2160.


packages/blocks/anodot/src/lib/api-filters/get-x-recommendation-property.ts
```diff
@@ -15,58 +15,114 @@ export function getPredefinedRecommendationsDropdownProperty() {
...
-  /* {
+  {
     label: 'AMI Orphaned Snapshot',
-    type_id: [],
-  },*/
+    filters: { type_id: ['ami-orphaned-snapshot'] },
+  },
   {
     label: 'AWS Backup Outdated Snapshot',
-    filters: { type_id: ['aws-backup-outdated-snapshot'] },
```
MarceloRGonc, line 24: I'm not sure which id we should use.

rita-gorokhod: could explain why it's not working for me right now in prod :D

packages/blocks/anodot/src/lib/api-filters/get-x-recommendation-property.ts
```diff
@@ -15,58 +15,114 @@ export function getPredefinedRecommendationsDropdownProperty() {
 }
 
 const recommendationTypes = [
-  /* {
+  {
     label: 'AMI Orphaned Snapshot',
-    type_id: [],
-  },*/
+    filters: { type_id: ['ami-orphaned-snapshot'] },
```
rita-gorokhod: I really hate this, is there any use case where we can't just do 
`value: 'ami-orphaned-snapshot'` as the dropdown value? (will require migration, but much simpler for the user when changing to dynamic)

MarceloRGonc: The only situation where I think this might be useful is if we want to add multiple filters associated with a label. But I don't think that's the case. Even the variable says 'recommendation type,' so it should be just one filter.

rita-gorokhod: Then let's simplify and change it to just
{ label: 'AMI Orphaned Snapshot', value: 'ami-orphaned-snapshot' }

packages/server/api/src/app/database/migrations/common/apply-update-to-flow-version.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
MarceloRGonc, line 1: We repeat this code every time we perform a flow version update. So I created a common file

packages/blocks/anodot/src/lib/api-filters/get-x-recommendation-property.ts
MarceloRGonc: Done

packages/blocks/anodot/src/lib/api-filters/get-x-recommendation-property.ts
```diff
@@ -1,94 +1,364 @@
...
-      })),
+      disabled: false,
+      options: recommendationTypes,
     },
-    required: true,
+    required: false,
   });
 }
 
 const recommendationTypes = [
```
rita-gorokhod: Please add AWS/Azure/GCP prefix to the labels

packages/blocks/anodot/test/api-filters/get-x-recommendation-property.test.ts
```diff
@@ -5,119 +5,359 @@ describe('getPredefinedRecommendationsDropdownProperty', () => {
...
 
     expect(result).toMatchObject({
-      required: true,
+      required: false,
       displayName: 'Recommendation',
       description: 'The type of recommendations to fetch',
-      type: 'STATIC_DROPDOWN',
+      type: 'STATIC_MULTI_SELECT_DROPDOWN',
       options: {
         options: [
```
rita-gorokhod: Please extract it to a json file of <key: label> so it would be easier to compare and update, and then re-use it

packages/server/api/src/app/database/migrations/1752758891771-UpdateUmbrellaRecommendationTypesAcrossFlowVersions.ts
```diff
@@ -0,0 +1,36 @@
...
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function updateJsonObject(obj: any): void {
+  if (
+    obj.settings?.blockName === '@openops/block-anodot' &&
+    obj.settings?.actionName === 'get_recommendations_predefined' &&
+    obj.settings?.input?.recommendationType !== undefined
+  ) {
+    obj.settings.input.recommendationTypes =
+      obj.settings?.input?.recommendationType?.filters?.type_id || [];
+    obj.settings.input.recommendationType = undefined;
```
rita-gorokhod: let's not remove the old field, if the user rolls back to the prev version

packages/blocks/anodot/src/lib/api-filters/get-x-recommendation-property.ts
```diff
@@ -1,94 +1,364 @@
 import { Property } from '@openops/blocks-framework';
 
 export function getPredefinedRecommendationsDropdownProperty() {
```
rita-gorokhod: Unrelated but can we finally rename the file name

rita-gorokhod (changes requested):
Some cosmetics, namely grouping together and labeling AWS/Azure/GCP so it would easier for the user

packages/blocks/anodot/src/lib/api-filters/get-predefined-recommendation-property.ts
```diff
@@ -0,0 +1,17 @@
+import { Property } from '@openops/blocks-framework';
+import recommendationTypes from './assets/recommendation-types.json';
+
+export function getPredefinedRecommendationsDropdownProperty() {
+  return Property.StaticMultiSelectDropdown({
+    displayName: 'Recommendation',
+    description: 'The type of recommendations to fetch',
+    options: {
+      disabled: false,
+      options: Object.entries(recommendationTypes).map(([label, value]) => ({
```
rita-gorokhod: sort alphabetically please

rita-gorokhod (approved):

PR #978, opened 2025-07-17
Fix 'Select all' selection when the value is an object

MarceloRGonc:
Fixes OPS-2188.


cezudas (approved):
https://www.loom.com/share/1fccfa9ba885471a904b9a915961dee8

We don't have to do anything else, the change is minimal and should not break anything

PR #979, opened 2025-07-17
Add Thread from Assistant UI and needed child components

cezudas:
Fixes OPS-2187.


alexandrudanpop:
can you please add some stories

cezudas:
> can you please add some stories

I added a couple in the meantime, but Thread and other child components are using these primitives
import {
  ActionBarPrimitive,
  BranchPickerPrimitive,
  ComposerPrimitive,
  MessagePrimitive,
  ThreadPrimitive,
} from '@assistant-ui/react';

which expects an AssistantRuntimeProvider, and cannot be conveniently mocked.

packages/ui-components/src/components/assistant-ui/markdown-text/markdown-text.tsx
```diff
@@ -0,0 +1,257 @@
...
+      .catch((error) => {
+        // eslint-disable-next-line no-console
+        console.debug('Failed to copy to clipboard:', error);
+      });
+  };
+
+  return { isCopied, copyToClipboard };
+};
+
+const defaultComponents = memoizeMarkdownComponents({
```
alexandrudanpop, line 71: Is this rendering the same as our other markdown component? I don't mean exact styles, as the current MD component is not the prettiest, but in terms of behavior?

alexandrudanpop (approved):

package.json
```diff
@@ -45,6 +45,8 @@
     "@ai-sdk/react": "1.2.9",
     "@ai-sdk/togetherai": "0.2.12",
     "@ai-sdk/xai": "1.2.14",
+    "@assistant-ui/react": "0.10.25",
+    "@assistant-ui/react-markdown": "0.10.6",
```
alexandrudanpop, line 49: we will need to lazy load those new packages

PR #980, opened 2025-07-17
Add Flexera One block

rita-gorokhod:
Fixes OPS-2179

image
image
image
image
image

### DIFF
--- a/packages/blocks/flexera-one/project.json
+++ b/packages/blocks/flexera-one/project.json
@@ -1,0 +1,33 @@
+{
+  "name": "blocks-flexera-one",
+  "$schema": "../../../node_modules/nx/schemas/project-schema.json",
+  "sourceRoot": "packages/blocks/flexera-one/src",
+  "projectType": "library",
+  "tags": [],
+  "targets": {
+    "build": {
+      "executor": "@nx/js:tsc",
+      "outputs": ["{options.outputPath}"],
+      "options": {
+        "outputPath": "dist/packages/blocks/flexera-one",
+        "tsConfig": "packages/blocks/flexera-one/tsconfig.lib.json",
+        "packageJson": "packages/blocks/flexera-one/package.json",
+        "main": "packages/blocks/flexera-one/src/index.ts",
+        "assets": ["packages/blocks/flexera-one/*.md"],
+        "buildableProjectDepsInPackageJsonType": "dependencies",
+        "updateBuildableProjectDepsInPackageJson": true
+      }
+    },
+    "test": {
+      "executor": "@nx/jest:jest",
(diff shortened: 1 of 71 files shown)


packages/blocks/flexera/src/auth.ts
```diff
@@ -4,10 +4,10 @@ const markdown = `
 Authenticate with your Flexera API Token to access Flexera services.
 `;
 
-export const flexeraAuth = BlockAuth.SecretAuth({
+export const flexeraSpotAuth = BlockAuth.SecretAuth({
   authProviderKey: 'Flexera',
-  authProviderDisplayName: 'Flexera',
-  authProviderLogoUrl: `https://static.openops.com/blocks/flexera.png`,
+  authProviderDisplayName: 'Flexera Spot',
```
rita-gorokhod, line 9: changed just the display name and logo, the key is the same for any existing connections

packages/blocks/flexera-one/project.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "blocks-flexera-one",
```
rita-gorokhod, line 2: called flexera-one just to avoid conflict with the existing block

packages/blocks/flexera-one/src/lib/common/governance-api.ts
```diff
@@ -0,0 +1,28 @@
+import { makeGetRequest } from './make-request';
+
+// https://reference.rightscale.com/governance-policies/
+
+const baseUrl = 'https://governance-3.rightscale.com/api/governance';
```
MarceloRGonc, line 5: Doesn't it depend on the region?

packages/blocks/flexera-one/src/lib/common/make-request.ts
```diff
@@ -0,0 +1,65 @@
...
+  switch (region) {
+    case 'us':
+      return 'flexera.com';
+    case 'eu':
+      return 'flexera.eu';
+    case 'apac':
+      return 'flexera.au';
+    default:
+      throw new Error(`Unsupported region: ${region}`);
+  }
```
MarceloRGonc, line 64: These methods should be in a single file. If we have to change an URL or add a new region, we need to look in different files.

packages/blocks/flexera-one/src/lib/common/governance-api.ts
rita-gorokhod: not according to their docs, but I'm not sure  https://reference.rightscale.com/governance-policies/

packages/blocks/flexera-one/src/lib/common/make-request.ts
rita-gorokhod: Will do in a separate PR

PR #983, opened 2025-07-18
Allow multiple HttpMethod in sendTernaryRequest

MarceloRGonc:
Part of OPS-2138.

## Additional Notes
- Also adding some unit tests

### DIFF
--- a/packages/blocks/ternary/src/lib/actions/get-budgets.ts
+++ b/packages/blocks/ternary/src/lib/actions/get-budgets.ts
@@ -1,5 +1,6 @@
 import { HttpMethod } from '@openops/blocks-common';
 import { createAction } from '@openops/blocks-framework';
+import { logger } from '@openops/server-shared';
 import { sendTernaryRequest } from '../common';
 import { ternaryCloudAuth } from '../common/auth';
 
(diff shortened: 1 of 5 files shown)


packages/blocks/ternary/src/lib/actions/get-budgets.ts
```diff
@@ -20,10 +21,9 @@ export const getBudgets = createAction({
         },
       });
       return response.body as any[];
-    } catch (e) {
-      console.error('Error getting budgets!');
-      console.error(e);
-      return e;
+    } catch (error) {
+      logger.error('Error getting budgets.', error);
```
rita-gorokhod, line 25: I'm not sure we need to log it, there is nothing we can do with this when we receive it in logzio

packages/blocks/ternary/src/lib/common/index.ts
```diff
@@ -12,11 +11,14 @@ export async function sendTernaryRequest(
...
-  if (!validJwt) throw new Error('Invalid JWT');
+
+  if (!validJwt) {
+    throw new Error('Invalid JWT');
+  }
+
   return httpClient.sendRequest({
     ...request,
     url: `${request.auth.apiURL}/api/${request.url}`,
-    method: HttpMethod.GET,
```
MarceloRGonc, line 19: This is the main change.

packages/blocks/ternary/src/lib/actions/get-budgets.ts
MarceloRGonc: It depends on the error. If it's a problem with our implementation, it's better to have the log. Otherwise, we will only know if someone complains.

PR #986, opened 2025-07-18
Add langfuse telemetry (only when configured)

alexandrudanpop:
Fixes OPS-2189

## Additional Notes
It's only enabled if the keys are present.
https://ai-sdk.dev/providers/observability/langfuse
https://langfuse.com/docs/sdk/typescript/guide

### DIFF
--- a/packages/server/api/test/integration/ce/authentication/signin.test.ts
+++ b/packages/server/api/test/integration/ce/authentication/signin.test.ts
@@ -12,6 +12,7 @@ jest.mock('jwt-decode', () => ({
 jest.mock('@openops/common', () => ({
   authenticateUserInOpenOpsTables: authUserMock,
   authenticateDefaultUserInOpenOpsTables: authUserMock,
+  isLLMTelemetryEnabled: jest.fn().mockReturnValue(false),
 }));
 
 import { faker } from '@faker-js/faker';
--- a/packages/server/api/test/integration/ce/authentication/signup.test.ts
+++ b/packages/server/api/test/integration/ce/authentication/signup.test.ts
@@ -14,6 +14,7 @@ jest.mock('jwt-decode', () => ({
 jest.mock('@openops/common', () => ({
   authenticateUserInOpenOpsTables: authUserMock,
   authenticateDefaultUserInOpenOpsTables: authUserMock,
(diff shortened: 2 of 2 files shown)


packages/openops/src/lib/ai/providers.ts
```diff
@@ -81,7 +82,6 @@ export const getAiProviderLanguageModel = async (aiConfig: {
     model: aiConfig.model,
     providerSettings: {
       ...sanitizedSettings,
-      experimental_telemetry: { enabled: false },
```
alexandrudanpop, line 84: this was not needed, and can't globally set it from here

rita-gorokhod: @alexandrudanpop are you sure? the telemetry in ai-sdk was enabled by default.

packages/server/api/src/app/app.ts
```diff
@@ -67,6 +71,17 @@ import { workerModule } from './workers/worker-module';
 export const setupApp = async (
   app: FastifyInstance,
 ): Promise<FastifyInstance> => {
+  const otelSDK = new NodeSDK({
```
maor-rozenfeld: Why not wrap this with "isLLMTelemetryEnabled"?

packages/openops/src/lib/ai/providers.ts
alexandrudanpop: Yeah, but without an Open telemetry exporter it's not sent anywhere.

packages/server/api/src/app/app.ts
alexandrudanpop: done

packages/server/shared/src/lib/system/system.ts
```diff
@@ -94,6 +94,7 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
   [AppSystemProp.AWS_MCP_COST_PATH]: '/root/.mcp/aws-cost',
   [AppSystemProp.SAMPLE_DATA_SIZE_LIMIT_KB]: '100',
   [AppSystemProp.REQUEST_BODY_LIMIT]: '10',
+  [SharedSystemProp.LANGFUSE_HOST]: 'https://us.cloud.langfuse.com',
```
alexandrudanpop, line 97: default host

PR #990, opened 2025-07-20
Support literal items and objects as loop input

rita-gorokhod:
Fixes OPS-2196

image
image
image
image
image


packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -45,13 +46,33 @@ export const loopExecutor: BaseExecutor<LoopOnItemsAction> = {
...
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      censoredInput: any;
+    } = await constants.variableService.resolve<LoopOnActionResolvedSettings>({
+      unresolvedInput: {
+        items: action.settings.items,
+      },
+      executionState,
+    });
+
+    if (isString(resolvedInput.items)) {
```
rita-gorokhod: this is ugly as hell but I figured it would be safest to do it here, where we know a valid input must be an array. I think it's better than the behavior today, where it just doesn't return anything, or passes individual chars if you give it a string.

packages/engine/src/lib/variables/variable-service.ts
```diff
@@ -197,12 +197,16 @@ export class VariableService {
 
     const flattenedSteps = executionState.currentState();
     const resolvedInput = await this.resolveInternally(
-      JSON.parse(JSON.stringify(unresolvedInput)),
+      isString(unresolvedInput)
```
rita-gorokhod: will revert and test without it

MarceloRGonc (approved):

PR #1000, opened 2025-07-22
Add chat history tool support when using Assistant-UI chat

cezudas:
Fixes OPS-2211.

### DIFF
--- a/packages/react-ui/src/app/features/ai/assistant-ui/assistant-ui-chat.tsx
+++ b/packages/react-ui/src/app/features/ai/assistant-ui/assistant-ui-chat.tsx
@@ -1,35 +1,26 @@
-import { AssistantRuntimeProvider } from '@assistant-ui/react';
+import {
+  AssistantRuntimeProvider,
+  ThreadMessageLike,
+} from '@assistant-ui/react';
 
-import { useVercelUseChatRuntime } from '@assistant-ui/react-ai-sdk';
+import { useChatRuntime } from '@assistant-ui/react-ai-sdk';
 
-import { AI_ASSISTANT_LS_KEY, ASSISTANT_UI_CHAT_ID } from '@/app/constants/ai';
+import { AI_ASSISTANT_LS_KEY } from '@/app/constants/ai';
 import { QueryKeys } from '@/app/constants/query-keys';
 import { authenticationSession } from '@/app/lib/authentication-session';
-import { Message, useChat } from '@ai-sdk/react';
 import { Thread, ThreadWelcomeProvider } from '@openops/components/ui';
 import { OpenChatResponse } from '@openops/shared';
 import { useQuery } from '@tanstack/react-query';
 import { t } from 'i18next';
-import { useMemo, useRef } from 'react';
-import { v4 as uuidv4 } from 'uuid';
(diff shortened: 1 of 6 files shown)


packages/react-ui/src/app/features/ai/assistant-ui/assistant-ui-chat.tsx
```diff
@@ -45,70 +38,56 @@ const AssistantUiChat = () => {
...
-          id: msg.id || uuidv4(),
-          role: msg.role as Message['role'],
-          content: contentString,
-        };
-      }
+  const initialMessages: ThreadMessageLike[] = useMemo(() => {
+    if (isLoading || !openChatResponse?.messages) {
+      return [];
+    }
+    return mergeToolResults(openChatResponse.messages);
```
alexandrudanpop: as discussed, I think we should do it server side when we save chat history.

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -134,10 +134,35 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
...
+          !Array.isArray(lastMessage.content) ||
+          lastMessage.content.length === 0
+        ) {
+          return await reply.code(400).send({
+            message:
+              'Last message must have valid content array with at least one element.',
+          });
+        }
 
+        messageContent = String(lastMessage.content[0].text);
```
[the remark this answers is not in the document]
alexandrudanpop: valid comment @cezudas

alexandrudanpop: maybe could extract the entire logic in a separate function:
getUserMessage

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -67,7 +67,11 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
         );
 
         if (existingContext) {
-          const messages = await getChatHistory(inputChatId, userId, projectId);
+          const messages = await getChatHistoryWithMergedTools(
+            inputChatId,
+            userId,
+            projectId,
+          );
```
alexandrudanpop, line 74: @cezudas why we don't do this in the `saveChatHistory` function?

cezudas: yes. good observation: AI-SDK actually expects `role: tool` , if I make the change in saveChatHistory `streamText` will fail with  "Invalid prompt: messages must be an array of CoreMessage or UIMessage"

packages/server/api/src/app/flags/dev-flags.service.ts
```diff
@@ -26,7 +26,7 @@ async function getAll(): Promise<Flag[]> {
     },
     {
       id: FlagId.ASSISTANT_UI_ENABLED,
-      value: false,
+      value: true,
```
cezudas: I only changed this to deploy on a hosted environment for a sanity test. Will revert it before merging

PR #1002, opened 2025-07-22
Add additional Cloudability actions

rita-gorokhod:
Fixes OPS-2197


packages/blocks/cloudability/src/lib/common/recommendations-api.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  const response = await makeRequest({
+    auth,
+    endpoint: url,
+    method: HttpMethod.GET,
+    queryParams: {
+      duration,
+      basis,
+      filters: filters.join(','),
+      ...(includeSnoozed === 'NO_SNOOZED' ? {} : { options: includeSnoozed }),
```
rita-gorokhod: if you want only active, you need to not pass the "options" argument.

p.s. the docs say it's "option" but it's in fact "options"

packages/blocks/cloudability/src/lib/common/recommendations-api.ts
```diff
@@ -0,0 +1,123 @@
...
+      ...(!limit || isEmpty(limit) ? {} : { limit: limit, offset: '0' }),
```
rita-gorokhod: if you pass limit you need to pass offset as well

packages/blocks/cloudability/src/lib/actions/snooze-recommendations-action.ts
```diff
@@ -0,0 +1,50 @@
...
+    }),
+    resourceIds: Property.Array({
+      displayName: 'Resource IDs',
+      description: 'The IDs of the resources to snooze recommendations for.',
+      required: true,
+    }),
+    snoozeUntil: Property.DateTime({
+      displayName: 'Snooze Until',
+      description:
+        'The date and time until the recommendation should be snoozed. Format: YYYY-MM-DDTHH:mm:ssZ. Use "never" to snooze indefinitely.',
```
rita-gorokhod: if you want to snooze forever you need to pass the string "never". Not sure if it's worth maybe adding a separate action for it?

packages/blocks/cloudability/src/lib/common/recommendations-api.ts
```diff
@@ -0,0 +1,123 @@
...
+export enum Duration {
+  TenDay = 'ten-day',
+  ThirtyDay = 'thirty-day',
+}
+export enum CostBasis {
+  OnDemand = 'on-demand',
+  Effective = 'effective',
+}
+
+export enum IncludeSnoozed {
```
bigfluffycookie: includeSnoozed sounds like a boolean, maybe SnoozedFilter ? idk

packages/blocks/cloudability/src/lib/common/common-properties.ts
```diff
@@ -0,0 +1,84 @@
...
+      displayName: 'Recommendation Type',
+      description: 'The type of recommendations to fetch.',
+      required: true,
+      refreshers: ['vendor'],
+      options: async ({ vendor }) => {
+        const recommendationTypes = getRecommendationTypesByVendor(
+          vendor as Vendor,
+        );
+
+        if (!recommendationTypes || recommendationTypes.length === 0) {
```
bigfluffycookie: if (!recommendationTypes?.length) {

packages/blocks/cloudability/src/lib/actions/snooze-recommendations-action.ts
bigfluffycookie: hmmm, i think its fine to keep it in here, another action doesnt seem super useful

packages/blocks/cloudability/src/lib/actions/get-recommendations-action.ts
```diff
@@ -0,0 +1,99 @@
...
+      },
+    }),
+    includeSnoozed: Property.StaticDropdown({
+      displayName: 'Recommendations Status',
+      description: 'Whether to include recommendations that have been snoozed.',
+      required: true,
+      defaultValue: 'NO_SNOOZED',
+      options: {
+        options: [
+          { label: 'Show Only Active', value: 'NO_SNOOZED' },
```
bigfluffycookie: why not use the enum here

packages/blocks/cloudability/src/lib/actions/get-recommendations-action.ts
```diff
@@ -0,0 +1,99 @@
...
+          { label: 'Show All (including snoozed)', value: 'INCLUDE_SNOOZED' },
+          { label: 'Show Only Snoozed', value: 'ONLY_SNOOZED' },
+        ],
+      },
+    }),
+    limit: Property.ShortText({
+      displayName: 'Limit',
+      description: 'The maximum number of recommendations to return.',
+      required: false,
+      defaultValue: undefined,
```
bigfluffycookie: isnt default undefined if you just dont add the property?

packages/blocks/cloudability/src/lib/actions/get-recommendations-action.ts
```diff
@@ -0,0 +1,99 @@
+import { createAction, Property, Validators } from '@openops/blocks-framework';
```
bigfluffycookie: please add tests for all three actions

packages/blocks/cloudability/src/lib/common/recommendations-api.ts
```diff
@@ -0,0 +1,123 @@
...
+  NO_SNOOZED = 'NO_SNOOZED',
+  INCLUDE_SNOOZED = 'INCLUDE_SNOOZED',
+  ONLY_SNOOZED = 'ONLY_SNOOZED',
+}
```
MarceloRGonc: If we are creating these enums, I think we should use them in properties rather than repeating the values.

packages/blocks/cloudability/src/lib/actions/get-recommendations-action.ts
rita-gorokhod: added for the base recommendations-api since the actions don't much besides calling it

PR #1003, opened 2025-07-22
Use stepId instead of stepName when generating a chatId

ravikiranvm:
Fixes OPS-2100.

## Testing Checklist

Check all that apply:

- Screenshots
- Loom video
- Preview deployment link

### DIFF
--- a/packages/react-ui/src/app/features/builder/ai-chat/lib/step-settings-ai-chat-hook.ts
+++ b/packages/react-ui/src/app/features/builder/ai-chat/lib/step-settings-ai-chat-hook.ts
@@ -67,7 +67,7 @@ export const useStepSettingsAiChat = (
       return aiChatApi.open(
         flowVersion.flowId,
         getBlockName(stepDetails),
-        selectedStep,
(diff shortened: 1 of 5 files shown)


alexandrudanpop:
You also need to remove `deleteChatRequest` in https://github.com/openops-cloud/openops/blob/main/packages/react-ui/src/app/features/builder/update-flow-version.ts#L41 (you can remove the function)

packages/react-ui/src/app/features/builder/ai-chat/lib/step-settings-ai-chat-hook.ts
```diff
@@ -67,7 +67,7 @@ export const useStepSettingsAiChat = (
       return aiChatApi.open(
         flowVersion.flowId,
         getBlockName(stepDetails),
-        selectedStep,
+        stepDetails?.id ?? '',
```
alexandrudanpop, line 70: add a check below on `enabled` param that `stepDetails?.id` is defined

PR #1019, opened 2025-07-23
Add 'get users' action to ternary block

MarceloRGonc:
Part of OPS-2138.


packages/blocks/ternary/src/lib/actions/get-users.ts
```diff
@@ -0,0 +1,28 @@
+import { createAction, Property } from '@openops/blocks-framework';
+import { logger } from '@openops/server-shared';
+import { ternaryCloudAuth } from '../common/auth';
+import { getUsersList } from '../common/users';
+
+export const getUsers = createAction({
+  name: 'get_users',
+  displayName: 'Get Users',
+  description: 'Get list of Users.',
```
bigfluffycookie: "Get a list" also users with small u

PR #1028, opened 2025-07-24
Remove unnecessary details from table request logs

MarceloRGonc:
Part of OPS-2158.


packages/openops/src/lib/openops-tables/rows.ts
```diff
@@ -182,7 +181,6 @@ export async function addRow(addRowParams: AddRowParams) {
       logger.error('Error while adding row:', {
         error,
         url,
-        ...addRowParams,
```
MarceloRGonc, line 185: This may contain sensitive data.

rita-gorokhod: @MarceloRGonc I think we print it also in other places, no? delete, upsert?

rita-gorokhod (approved):

packages/openops/src/lib/openops-tables/rows.ts
MarceloRGonc: Fixed. I'm adding the properties that we don't have in other parameters. For example the tableId and rowId is defined in the URL

PR #1035, opened 2025-07-28
Fix slack interactions returning Unexpected token issue

bigfluffycookie:
Fixes OPS-2242

### DIFF
--- a/packages/blocks/slack/src/lib/common/wait-for-interaction.ts
+++ b/packages/blocks/slack/src/lib/common/wait-for-interaction.ts
@@ -75,15 +75,7 @@ export async function onReceivedInteraction(
     };
   }
 
-  let userSelection;
-  try {
-    userSelection = JSON.parse(resumePayload.actionClicked);
-  } catch {
-    userSelection = {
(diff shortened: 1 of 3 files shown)


packages/blocks/slack/src/lib/common/wait-for-interaction.ts
```diff
@@ -75,7 +75,15 @@ export async function onReceivedInteraction(
     };
   }
 
-  const userSelection = JSON.parse(resumePayload.actionClicked);
+  let userSelection;
+  try {
```
rita-gorokhod: why not fix it here?
[https://github.com/openops-cloud/openops/blob/89675b2e308d644023063af8b6a5a8ea39d5[…]packages/blocks/slack/src/lib/actions/request-action-message.ts](https://github.com/openops-cloud/openops/blob/89675b2e308d644023063af8b6a5a8ea39d5d9c4/packages/blocks/slack/src/lib/actions/request-action-message.ts#L118)

rita-gorokhod (approved):

PR #1039, opened 2025-07-28
Mark the workflow run as failed when a general engine error occurs

MarceloRGonc:
Fixes OPS-2239.


packages/react-ui/src/app/features/builder/run-details/run-details-helpers.ts
```diff
@@ -11,7 +11,7 @@ export function getRunMessage(
     run.status === FlowRunStatus.SCHEDULED
   )
     return null;
-  if (run.status === FlowRunStatus.INTERNAL_ERROR) {
+  if (run.status === FlowRunStatus.INTERNAL_ERROR && isNil(run.logsFileId)) {
```
MarceloRGonc, line 14: Show the steps that ran even if the workflow fails.

image

packages/engine/src/lib/operations.ts
```diff
@@ -278,7 +279,12 @@ export async function execute(
     logger.warn('Engine operation failed.', error);
     return {
       status: EngineResponseStatus.ERROR,
-      response: utils.tryParseJson((error as Error).message),
+      response: {
+        status: FlowRunStatus.INTERNAL_ERROR,
+        error: {
+          message: utils.tryParseJson((error as Error).message),
+        },
+      },
```
MarceloRGonc, line 287: Object expected to update flow as failed.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -136,6 +137,10 @@ export const loopExecutor: BaseExecutor<LoopOnItemsAction> = {
 };
 
 function processItems(items: readonly unknown[]): readonly unknown[] {
+  if (isEmpty(items)) {
+    return [];
+  }
```
MarceloRGonc, line 142: If the provided items are an empty string, the JSON parse fails. The isEmpty method checks if the value is a string. Also checks if it's an empty array or an empty object.

packages/react-ui/src/app/features/builder/run-details/run-details-helpers.ts
rita-gorokhod: Can we show the failed step? like in this example, why are we not showing that the loop is the one that failed, and what was the input to it?

MarceloRGonc: In this situation, we don't know which step failed. It's an unexpected error. However, for the workflow that fails in staging, it will successfully execute the empty loop.

PR #1040, opened 2025-07-28
Rename cost-alert action of Ternary block

MarceloRGonc:
Part of OPS-2138


rita-gorokhod:
As we discussed, let's not do it

PR #1044, opened 2025-07-28
Update styles for Assistant UI chat container and resizable panel.

rSnapkoOpenOps:
Fixes #<issue_number>.

## Testing Checklist

Check all that apply:

- Screenshots
- Loom video
- Preview deployment link

### DIFF
--- a/packages/react-ui/src/app/features/builder/ai-chat/ai-chat-resizable-panel.tsx
+++ b/packages/react-ui/src/app/features/builder/ai-chat/ai-chat-resizable-panel.tsx
@@ -20,7 +20,7 @@ const AiChatResizablePanelContent = ({
   if (!showChat) return null;
 
   return (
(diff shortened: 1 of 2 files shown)


packages/react-ui/src/app/features/builder/ai-chat/ai-chat-resizable-panel.tsx
```diff
@@ -71,6 +67,12 @@ const AiChatResizablePanel = ({
     showChatInResizablePanel,
   ]);
 
+  useEffect(() => {
+    setTimeout(() => {
+      setShowChatDelayed(showChat);
+    }, 300);
+  }, [showChat]);
+
```
rmn-snpk: We need to have 'min-width' to prevent user from collapsing the panel by dragging. But at the same time it breaks the opening behavior. So we set min-width but do it when the panel is already opened

PR #1045, opened 2025-07-28
Extract the message stream code from the controller to the user-message-handler file

MarceloRGonc:
Part of OPS-2123.


packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  sendAiChatMessageSendEvent({
+    projectId,
+    userId,
+    chatId,
+    provider: aiConfig.provider,
+  });
+}
+
+async function streamMessages(
```
alexandrudanpop, line 124: I think `streamLLMResponse` would be a better name

packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -0,0 +1,235 @@
...
+    authToken,
+    aiConfig,
+    messages,
+    chatContext,
+    languageModel,
+  );
+
+  pipeDataStreamToResponse(serverResponse, {
+    execute: async (dataStreamWriter) => {
+      logger.debug('Send user message to LLM.');
```
alexandrudanpop, line 81: do we need this log?

packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -0,0 +1,235 @@
...
+    };
+  }
+
+  return {
+    role: 'assistant',
+    content,
+  };
+}
+
+async function closeMCPClients(mcpClients: unknown[]): Promise<void> {
```
alexandrudanpop, line 230: would wrap in try catch and do `closeMCPClients` as first call in onFinish / onError

packages/server/api/src/app/ai/chat/user-message-handler.ts
MarceloRGonc: Just moved the code. I will do a proper refactor in another PR.

packages/server/api/src/app/ai/chat/user-message-handler.ts
MarceloRGonc: This will disappear :)

packages/server/api/src/app/ai/chat/user-message-handler.ts
MarceloRGonc: After the refactor, it will be outside the stream.

alexandrudanpop (approved):

PR #1051, opened 2025-07-29
Add action "List All" to Storage block

bigfluffycookie:
Fixes OPS-2007

image

### DIFF
--- a/packages/blocks/store/src/lib/actions/store-list-action.ts
+++ b/packages/blocks/store/src/lib/actions/store-list-action.ts
@@ -1,5 +1,5 @@
 import { createAction, Property } from '@openops/blocks-framework';
-import { BlockStoreScope, common, getScopeAndKey } from './common';
+import { common, getScopeAndKey } from './common';
 
 export const storageListAction = createAction({
   name: 'list',
--- a/packages/server/api/src/app/store-entry/store-entry.service.ts
+++ b/packages/server/api/src/app/store-entry/store-entry.service.ts
@@ -1,4 +1,3 @@
-import { logger } from '@openops/server-shared';
 import {
   openOpsId,
   ProjectId,


packages/blocks/store/src/lib/actions/store-list-action.ts
```diff
@@ -0,0 +1,59 @@
...
+        throw new Error(`Invalid regex pattern: ${keyFilter}`);
+      }
+    }
+
+    const entries = await context.store.list(scope);
+
+    if (filterRegex) {
+      return entries.filter((entry) => {
+        let keyName = entry.key;
+        if (context.run.isTest) {
```
bigfluffycookie: This is to counteract the logic in https://github.com/openops-cloud/openops/blob/main/packages/blocks/store/src/lib/actions/common.ts#L23

packages/server/api/src/app/store-entry/store-entry.controller.ts
```diff
@@ -74,3 +86,13 @@ const DeleteStoreRequest = {
       "Delete a key-value store entry. This endpoint removes a specific key-value pair from the project's store. Note that this operation is restricted to engine principals only.",
   },
 };
+
+const ListRequest = {
+  schema: {
+    querystring: Type.Object({
+      prefix: Type.Optional(Type.String()),
```
MarceloRGonc: In what situations is this optional?

bigfluffycookie: honestly no idea, i just making things optional sometimes

packages/blocks/store/src/lib/actions/store-list-action.ts
```diff
@@ -0,0 +1,68 @@
...
+        filterRegex = new RegExp(keyFilter);
+      } catch (error) {
+        throw new Error(`Invalid regex pattern: ${keyFilter}`);
+      }
+    }
+
+    const entries = await context.store.list(scope, keyPrefix);
+
+    if (filterRegex) {
+      return entries.filter((entry) => {
```
MarceloRGonc: In my opinion, we should not apply filters on the Engine side. We should only return from the server the values that were requested.

packages/blocks/framework/src/lib/context.ts
```diff
@@ -188,6 +188,10 @@ export interface Store {
   put<T>(key: string, value: T, scope?: StoreScope): Promise<T>;
   get<T>(key: string, scope?: StoreScope): Promise<T | null>;
   delete(key: string, scope?: StoreScope): Promise<void>;
+  list(
+    scope?: StoreScope,
+    keyPrefix?: string,
```
MarceloRGonc: These values are not optional

packages/blocks/store/src/lib/actions/store-list-action.ts
bigfluffycookie: how would you apply the regex filtering then?

MarceloRGonc: Postgres accepts regex when filtering. But if that doesn't work, you can pass this filtering to the server.

packages/blocks/framework/src/lib/context.ts
```diff
@@ -188,6 +188,12 @@ export interface Store {
   put<T>(key: string, value: T, scope?: StoreScope): Promise<T>;
   get<T>(key: string, scope?: StoreScope): Promise<T | null>;
   delete(key: string, scope?: StoreScope): Promise<void>;
+  list(
+    scope: StoreScope,
+    keyPrefix: string,
+    filterRegex?: string,
+    isTestRun?: boolean,
```
MarceloRGonc: You can remove 'isTestRun'

packages/server/api/src/app/store-entry/store-entry.service.ts
```diff
@@ -5,6 +5,7 @@ import {
   sanitizeObjectForPostgresql,
   StoreEntry,
 } from '@openops/shared';
+import { Like } from 'typeorm';
```
MarceloRGonc: You can remove this.

packages/server/api/src/app/store-entry/store-entry.service.ts
```diff
@@ -62,4 +63,36 @@ export const storeEntryService = {
...
     });
   },
+
+  async list({
+    projectId,
+    prefix,
+    filterRegex,
+  }: {
+    projectId: ProjectId;
+    prefix?: string;
```
MarceloRGonc: Is the prefix optional?

packages/server/api/src/app/store-entry/store-entry.service.ts
```diff
@@ -62,4 +63,36 @@ export const storeEntryService = {
...
+      const escapedPrefix = prefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+      query.andWhere('storeEntry.key LIKE :prefix', { prefix: `${prefix}%` });
+      if (filterRegex) {
+        keyExpression = `REGEXP_REPLACE(${keyExpression}, '^${escapedPrefix}', '', 'g')`;
+      }
+    }
+
+    if (filterRegex) {
+      query.andWhere(`${keyExpression} ~ :filterRegex`, { filterRegex });
+    }
```
MarceloRGonc: I think the prefix isn't optional, so the condition will disappear. If not, I think there's an inconsistency here with the keyExpression.

packages/server/api/src/app/store-entry/store-entry.service.ts
```diff
@@ -62,4 +62,34 @@ export const storeEntryService = {
...
+
+    const escapedPrefix = prefix.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    query.andWhere('storeEntry.key LIKE :prefix', { prefix: `${prefix}%` });
+    if (filterRegex) {
+      keyExpression = `REGEXP_REPLACE(${keyExpression}, '^${escapedPrefix}', '', 'g')`;
+    }
+
+    if (filterRegex) {
+      query.andWhere(`${keyExpression} ~ :filterRegex`, { filterRegex });
+    }
```
MarceloRGonc:
```suggestion
    if (filterRegex) {
      keyExpression = `REGEXP_REPLACE(${keyExpression}, '^${escapedPrefix}', '', 'g')`;
      query.andWhere(`${keyExpression} ~ :filterRegex`, { filterRegex });
    }
```

I don't think we need the keyExpression variable

PR #1054, opened 2025-07-29
Refactor the conversation endpoint to better control the stream

MarceloRGonc:
Part of OPS-2123

## Additional Notes
- Refactoring to separate the AI chat logic from the controller.


packages/server/api/src/app/ai/chat/llm-stream-handler.ts
```diff
@@ -0,0 +1,59 @@
...
+  const toolChoice = hasTools ? 'auto' : 'none';
+
+  const { fullStream } = streamText({
+    model: languageModel,
+    system: systemPrompt,
+    messages: chatHistory,
+    ...aiConfig.modelSettings,
+    tools,
+    toolChoice,
+    maxRetries: 1,
```
alexandrudanpop: could we extract those decisions as constants at the top of the file?

const TOOL_CHOISE = 'auto'
const MAX_RETRIES = 1

MarceloRGonc: The tool choice is not always auto. But I added the MAX_RETRIES

PR #1063, opened 2025-08-01
Extend conversation endpoint to handle code generation

cezudas:
Fixes OPS-2258.

https://www.loom.com/share/f6e60b9f64064c35a5bcdd9e4b19753e


packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -133,20 +135,33 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
         return; // Error response already sent
       }
 
+      const { chatContext } = await getConversation(chatId, userId, projectId);
+      const isCodeGenerationRequest =
+        chatContext?.blockName === CODE_BLOCK_NAME;
```
MarceloRGonc: If you need the conversation here, it's better to pass it as a parameter to the methods:

- handleCodeGenerationRequest
- handleUserMessage

It's one less call.

packages/server/api/src/app/ai/chat/code-generation-handler.ts
```diff
@@ -0,0 +1,328 @@
...
+  serverResponse.write(
+    `b:{"type":"tool-call-streaming-start","toolCallId":"${toolCallId}","toolName":"${GENERATE_CODE_TOOL_NAME}"}\n`,
+  );
+  serverResponse.write(
+    `c:{"type":"tool-call-delta","toolCallId":"${toolCallId}","toolName":"${GENERATE_CODE_TOOL_NAME}","argsTextDelta":"{\\"message\\":\\"${newMessage.content}\\"}"}\n`,
+  );
+
+  serverResponse.write(
+    `9:{"type":"tool-call","toolCallId":"${toolCallId}","toolName":"${GENERATE_CODE_TOOL_NAME}","args":{"message":"${newMessage.content}"}}\n`,
+  );
```
MarceloRGonc: I don't think you need the 'tool-call-delta' message. Either way, extract this into a method to be readable.

packages/server/api/src/app/ai/chat/code-generation-handler.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  const llmConfigResult = await getLLMConfig(projectId);
+  const conversationResult = await getConversation(chatId, userId, projectId);
+
+  conversationResult.chatHistory.push(newMessage);
+
+  const { chatContext, chatHistory } = conversationResult;
+  const { aiConfig, languageModel } = llmConfigResult;
```
MarceloRGonc: This is common. Maybe we can have a common file that has the try{} finally{} and the condition from the controller inside.

packages/server/api/src/app/ai/chat/code-generation-handler.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    const finishMessage = {
+      finishReason: 'tool-calls',
+      usage: { promptTokens: null, completionTokens: null },
+      isContinued: false,
+    };
+    serverResponse.write(`e:${JSON.stringify(finishMessage)}\n`);
+    serverResponse.write(`d:${JSON.stringify(finishMessage)}\n`);
+
+    serverResponse.write(`0:"${finalCodeResult.textAnswer}"\n`);
```
MarceloRGonc: Maybe we should have a file with methods to build the messages. Like
- textMessage() => `0:"${value}"\n`
- toolCallMessage() => `9:"${JSON.stringify(value)}"\n`

packages/server/api/src/app/ai/chat/code-generation-handler.ts
cezudas: If i simply remove the tool-call-delta I don't see the args on the FE tool component. ¯\_(ツ)_/¯

packages/server/api/src/app/ai/chat/code-generation-handler.ts
cezudas: done, made the changes in user-message-handler where it was obvious, but didn't do it in function sendMessageToStream since In code-generation-handler I don't really use the built-in types/interfaces for tools, only using simple mocks.

packages/server/api/src/app/ai/chat/code-generation-handler.ts
cezudas: The handlers are different enough that forcing them into a common pattern might make the code more complex rather than simpler. The duplication is minimal and the differences are meaningful.
What do you think? Should we keep the current structure or try to extract some common patterns?

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
cezudas: done

packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -10,36 +10,27 @@ import {
...
-  projectId: string;
-  serverResponse: ServerResponse;
-};
-
-type UserMessageParams = RequestContext & {
-  authToken: string;
-  app: FastifyInstance;
-  newMessage: CoreMessage;
-};
+export type UserMessageParams = RequestContext &
```
MarceloRGonc: I think the export is not needed.

packages/server/api/src/app/ai/chat/types.ts
```diff
@@ -14,3 +17,23 @@ export type MessageWithMergedToolResults = {
   role: 'user' | 'assistant' | 'system';
   content: string | (TextPart | ToolCallPartWithResult)[];
 };
+
+export type RequestContext = {
+  userId: string;
+  chatId: string;
+  projectId: string;
+
```
MarceloRGonc: Extra new line.

PR #1064, opened 2025-08-01
Fix chevron not pointing down when selecting category

cezudas:
Fixes OPS-2234.


packages/ui-components/src/components/flow-template/flow-template-category-collapsible.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+type FlowTemplateCategoryCollapsibleProps = {
+  category: TemplateSidebarCategory;
+  selectedCategories: string[];
+  selectedServices: string[];
+  onCategoryFilterClick: (category: string) => void;
+  onServiceFilterClick: (service: string) => void;
+};
+
+const FlowTemplateCategoryCollapsible = ({
```
cezudas, line 20: Extracted Inline JSX from FlowTemplateFilterItem, crux of the changes are
```
  const isCategorySelected = selectedCategories.includes(category.name);
  const ChevronIcon = isCategorySelected ? ChevronDown : ChevronRight;
```
```

PR #1065, opened 2025-08-01
Add enterprise package loading if available

MarceloRGonc:
Part of OPS-2267.

## Additional Notes
- The hot reloading is not working for the new package.


rita-gorokhod:
As we discussed, we're not doing it @MarceloRGonc

PR #1068, opened 2025-08-04
Prevent other template filters being reset when user types in the search input

cezudas:
Part of OPS-1108.

Adjusts templates filtering behavior so that typing in search input when a filter is enabled will further filter (AND) the results.


packages/react-ui/src/app/features/templates/components/select-flow-template-dialog.tsx
```diff
@@ -349,7 +349,7 @@ const SelectFlowTemplateDialog = ({
     setSelectedServices([]);
     setSelectedCategories([]);
     resetTemplateDialog();
-  }, [isOpen, resetTemplateDialog, searchText]);
```
cezudas, line 352: Removing `searchText` from the deps list will ensure that any template filter previously applied is no longer being reset.

rSnapkoOpenOps (approved):

PR #1076, opened 2025-08-04
Fix CloudFix custom api auth

rita-gorokhod:
Fixes OPS-2315

### DIFF
--- a/packages/blocks/cloudfix/src/index.ts
+++ b/packages/blocks/cloudfix/src/index.ts
@@ -25,7 +25,7 @@ export const cloudfix = createBlock({
       baseUrl: (auth: any) => auth.apiUrl,
       auth: cloudfixAuth,
       authMapping: async (context) => ({
-        Bearer: (context.auth as any).apiKey,
+        Authorization: 'Bearer ' + (context.auth as any).apiKey,
       }),
     }),
   ],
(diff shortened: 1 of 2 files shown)


packages/blocks/cloudfix/src/lib/common/make-request.ts
```diff
@@ -1,6 +1,6 @@
 import { httpClient, HttpMethod } from '@openops/blocks-common';
 import qs from 'qs';
-import { cloudfixAuth, CloudfixAuth } from './auth';
```
rita-gorokhod, line 3: was unused

packages/blocks/cloudfix/src/index.ts
```diff
@@ -25,7 +25,7 @@ export const cloudfix = createBlock({
       baseUrl: (auth: any) => auth.apiUrl,
       auth: cloudfixAuth,
       authMapping: async (context) => ({
-        Bearer: (context.auth as any).apiKey,
+        Authorization: 'Bearer ' + (context.auth as any).apiKey,
```
rita-gorokhod, line 28: FYI I used Bearer, because that's also what we use in makeRequest, so we should be consistent https://github.com/openops-cloud/openops/blob/5060acede21128a1c41809728f96c0d9f0a91494/packages/blocks/cloudfix/src/lib/common/make-request.ts#L31

bigfluffycookie (approved):

PR #1077, opened 2025-08-04
Handle UI tools in AI SDK

alexandrudanpop:
Part of OPS-2314


packages/server/api/src/app/ai/mcp/tool-utils.ts
```diff
@@ -0,0 +1,99 @@
...
+    ]),
+  );
+
+/**
+ * Adds separate tool messages for UI tool calls (tools with names starting with 'ui-')
+ * This ensures frontend tools have results in chat history for proper conversation flow
+ * In the future, we should refactor to make the FE save the real results to the chat history
+ */
+export function addUiToolResults(messages: CoreMessage[]): CoreMessage[] {
+  const newMessages: CoreMessage[] = [];
```
alexandrudanpop, line 34: without this we get an error that a tool call is not followed by a tool message after invoking a UI tool. The issue is that we always read for history, but when fully switching to assistant UI the history is passed as messages and we only need to save the history in the frontend `onFinish` hook (also need to an API for the session saving)

cezudas (approved):

PR #1089, opened 2025-08-05
Add new endpoint to manually trigger workflows

bigfluffycookie:
Fixes OPS-2273


packages/server/api/src/app/flows/flow/flow.controller.ts
```diff
@@ -197,6 +252,33 @@ async function extractUserIdFromPrincipal(
   return project.ownerId;
 }
 
+async function validateTriggerType(
```
bigfluffycookie, line 272: If anyone knows an easier way to verify that the trigger is of strategy polling, lmk

packages/server/api/src/app/flows/flow/flow.controller.ts
```diff
@@ -197,6 +268,43 @@ async function extractUserIdFromPrincipal(
...
+  flow: PopulatedFlow,
+  projectId: string,
+): Promise<{ success: boolean; message: string }> {
+  const blockTrigger = flow.version.trigger;
+
+  if (blockTrigger.type !== TriggerType.BLOCK) {
```
MarceloRGonc, line 278: This will never happen. If the workflow is published, you must have a trigger. And there are only block triggers.

packages/server/api/src/app/flows/flow/flow.controller.ts
```diff
@@ -325,3 +433,32 @@ const DeleteFlowRequestOptions = {
     },
   },
 };
+
+const RunFlowRequestOptions = {
+  config: {
+    allowedPrincipals: [PrincipalType.USER, PrincipalType.SERVICE],
```
MarceloRGonc: Why service?

packages/server/api/src/app/flows/flow/flow.controller.ts
```diff
@@ -139,6 +147,69 @@ export const flowController: FastifyPluginAsyncTypebox = async (app) => {
...
+      if (
+        error instanceof ApplicationError &&
+        error.error?.code === ErrorCode.ENTITY_NOT_FOUND
+      ) {
+        return reply.status(StatusCodes.BAD_REQUEST).send({
+          success: false,
+          message:
+            'Workflow must be published before it can be triggered manually',
+        });
+      }
```
MarceloRGonc: So, we check if the workflow is published. Is it possible to reach this catch for that reason?

packages/server/api/src/app/flows/flow/flow.controller.ts
bigfluffycookie: Theres TriggerType.Empty in the enum, which is why im doing it

packages/server/api/src/app/flows/flow/flow.controller.ts
bigfluffycookie: fixed the message

packages/server/api/src/app/flows/flow/flow.controller.ts
```diff
@@ -197,6 +268,43 @@ async function extractUserIdFromPrincipal(
...
+  }
+
+  try {
+    const metadata = await triggerUtils.getBlockTriggerOrThrow({
+      trigger: blockTrigger,
+      projectId,
+    });
+
+    return {
+      success: metadata.type === TriggerStrategy.POLLING,
```
rmn-snpk, line 292: I think we also need `WEBHOOK` triggers here

bigfluffycookie: Why? with the webhook i thought you can just go through the normal webhook logic, once you built it cant you just invoke it?

rmn-snpk: True. But won't it affect running workflows by AI agent in future?

bigfluffycookie: Rita says no need for a new endpoint for webhooks, I guess we will get to it when it comes up  then, but i think agents are able to invoke urls also

rSnapkoOpenOps (approved):

packages/server/api/src/app/flows/flow/flow.controller.ts
MarceloRGonc: It's okay to do this. But if this happens, we have a bug. TriggerType.Empty only exists when you create the workflow and have no trigger selected. And without selecting a trigger, you can't publish.

PR #1091, opened 2025-08-06
Return Jira issue HTML url

rita-gorokhod:


packages/blocks/jira-cloud/src/index.ts
```diff
@@ -1,6 +1,5 @@
-import { createBlock } from '@openops/blocks-framework';
-
 import { createCustomApiCallAction } from '@openops/blocks-common';
+import { createBlock } from '@openops/blocks-framework';
```
rita-gorokhod, line 2: auto format

packages/blocks/jira-cloud/src/auth.ts
```diff
@@ -1,12 +1,6 @@
 import { HttpError, HttpMethod } from '@openops/blocks-common';
-import {
```
rita-gorokhod, line 2: was unused

packages/blocks/jira-cloud/src/lib/common/index.ts
```diff
@@ -22,6 +22,12 @@ export async function sendJiraRequest(
         password: request.auth.apiToken,
       },
     });
+
+    if (response?.body?.key) {
```
rita-gorokhod, line 26: this is the change

packages/blocks/jira-cloud/src/lib/common/index.ts
```diff
@@ -22,6 +22,12 @@ export async function sendJiraRequest(
...
+      response.body.htmlUrl = `${request.auth.instanceUrl}/browse/${response.body.key}`;
```
rita-gorokhod, line 27: jira doesn't care if there are two "/", it will redirect:
https://openops.atlassian.net//browse/KAN-1586

PR #1096, opened 2025-08-06
Add new column for trigger source in flow_run table

bigfluffycookie:
OPS-2276


packages/server/api/src/app/database/migrations/1754489349004-AddTriggerSourceToFlowRun.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class AddTriggerSourceToFlowRun1754489349005
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      ALTER TABLE flow_run 
+      ADD COLUMN IF NOT EXISTS "triggerSource" varchar;
+
+      UPDATE flow_run
```
bigfluffycookie, line 12: this is the only thing im not 100% convinced on, should we fill them retroactively? if not we cannot enforce the NOT NULL

PR #1133, opened 2025-08-12
Fix LLM loading indicator

alexandrudanpop:
Fixes OPS-2354

https://www.loom.com/share/a0464961fb434eb6ae05ae763a3baebc

### DIFF
--- a/packages/ui-components/src/components/assistant-ui/markdown-text/markdown-text.tsx
+++ b/packages/ui-components/src/components/assistant-ui/markdown-text/markdown-text.tsx
@@ -1,5 +1,3 @@
-import '@assistant-ui/react-markdown/styles/dot.css';
-
 import {
   MarkdownTextPrimitive,
   unstable_memoizeMarkdownComponents as memoizeMarkdownComponents,
--- a/packages/ui-components/src/components/assistant-ui/thread/thread.tsx
+++ b/packages/ui-components/src/components/assistant-ui/thread/thread.tsx
@@ -70,6 +70,16 @@ export const Thread = ({
 
         <ThreadPrimitive.Messages components={messageComponents} />
 
+        <ThreadPrimitive.If running>
+          <div className="w-full max-w-[var(--thread-max-width)] mb-2">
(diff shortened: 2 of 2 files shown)


packages/ui-components/src/components/assistant-ui/thread/thread.tsx
```diff
@@ -218,7 +228,7 @@ const AssistantMessage: FC<{ theme: Theme }> = ({ theme }) => {
   );
 
   return (
-    <MessagePrimitive.Root className="grid grid-cols-[auto_auto_1fr] grid-rows-[auto_1fr] relative w-full max-w-[var(--thread-max-width)] py-4">
+    <MessagePrimitive.Root className="grid grid-cols-[auto_auto_1fr] grid-rows-[auto_1fr] relative w-full max-w-[var(--thread-max-width)] py-2">
```
alexandrudanpop: it was too much vertical padding between messages

packages/ui-components/src/components/assistant-ui/markdown-text/markdown-text.tsx
```diff
@@ -1,5 +1,3 @@
-import '@assistant-ui/react-markdown/styles/dot.css';
-
```
alexandrudanpop, line 2: this rendered their default spinner. This is problematic as it does not render after tool calls, even though conversation is not finished

packages/ui-components/src/components/assistant-ui/thread/thread.tsx
```diff
@@ -70,6 +70,16 @@ export const Thread = ({
 
         <ThreadPrimitive.Messages components={messageComponents} />
 
+        <ThreadPrimitive.If running>
+          <div className="w-full max-w-[var(--thread-max-width)] mb-2">
```
rmn-snpk: Do we need `--thread-max-width` as variable? Where do we set it?

alexandrudanpop: from what I see we did not have it but it was referenced in multiple parts in this file, I removed it everywhere

PR #1137, opened 2025-08-12
Add some descriptions to the types used by action endpoints

MarceloRGonc:
Part of OPS-2272.


packages/server/api/src/app/server.ts
```diff
@@ -39,7 +39,7 @@ async function setupBaseApp(): Promise<FastifyInstance> {
     trustProxy: true,
     ajv: {
       customOptions: {
-        removeAdditional: 'all',
+        removeAdditional: false,
```
MarceloRGonc, line 42: Necessary because AI models seem to dislike Records, so I needed to change and use an object of any type.

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -19,12 +19,35 @@ export enum RiskLevel {
...
+      'Indicates whether the action is currently valid and can be executed',
+  }),
+  displayName: Type.String({
+    description: 'Human-readable name for the action, intended for UI display',
+  }),
 };
 
+export const InputsSchema = Type.Object(
+  {},
+  {
```
MarceloRGonc, line 40: This was previously a Type.Record(Type.String({}), Type.Any())
Now it's an object that allows any key/value

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -19,12 +19,35 @@ export enum RiskLevel {
...
+    additionalProperties: true,
+    description:
+      'All block-specific properties must be contained within the `input` object, not at the root level. ' +
+      'The `input` object is a mapping of property keys to their chosen values (`input: { propertyKey → propertyValue }`). ' +
+      'Keys must exactly match the action property keys returned by the details endpoint, and values must follow the expected types and constraints. ' +
```
alexandrudanpop:
```suggestion
      'Keys must exactly match the action property keys returned by the action details tool, and values must follow the expected types and constraints. ' +
```

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -44,15 +67,21 @@ export type ActionErrorHandlingOptions = Static<
...
-  packageJson: Type.String({}),
-  code: Type.String({}),
+  packageJson: Type.String({
+    description:
+      'JSON string for a minimal package.json snippet focused on runtime dependencies. It must be a JSON object with an optional `dependencies` field mapping package names to version ranges. Example: { "dependencies": { "stylus": "0.64.0" } }. Use \'{}\' when there are no dependencies.',
+  }),
+  code: Type.String({
+    description:
+      'JavaScript/TypeScript source for this step. To use data from previous steps, pass them as key–value pairs and access them via `inputs.key` in your code. The entry point must be an exported function named `code`. If it is removed or renamed, the step will fail.\n\nExample:\n```ts\nexport const code = async (inputs) => {\n  // use inputs.someKey\n  return true;\n};\n```',
+  }),
```
alexandrudanpop:
```suggestion
  code: Type.String({
    description:
      'JavaScript/TypeScript source for this step. To use data from previous steps, pass them as key–value pairs and access them via `inputs.key` in your code. The entry point must be an exported function named `code`. If it is removed or renamed, the step will fail.\n\ncode MUST follow this exact interface:

```javascript
export const code = async (inputs) => {
  // Your transformation logic here
  // Process the inputs object
  // Return the result
  return result;
};
```',
  }),
```

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -765,7 +765,8 @@ const prefillConnection = (
     'auth' in action.settings.input &&
```
alexandrudanpop, line 765: What are those changes in this file for?

packages/shared/src/lib/flows/flow-operations.ts
```diff
@@ -110,11 +110,21 @@ export const MoveActionRequest = Type.Object({
...
+    Type.Enum(StepLocationRelativeToParent, {
+      description:
+        'Indicates the relative position of this step compared to its parent (e.g., before, after, inside)',
+    }),
+  ),
+  branchNodeId: Type.Optional(
+    Type.String({
+      description:
+        'The unique identifier of the branch node this step belongs to, if applicable',
+    }),
```
alexandrudanpop:
```suggestion
  branchNodeId: Type.Optional(
    Type.String({
      description:
        'The unique identifier of the branch node this step belongs to for placing the step inside a Split branch. Each branch in a Split node has a unique branchNodeId',
    }),
```

packages/shared/src/lib/flows/flow-helper.ts
MarceloRGonc: Because I changed the type from Record to an object with no declared properties, I need to cast to 'any' to avoid a 'Property does not exist on type '{}' error.

packages/shared/src/lib/flows/actions/action.ts
alexandrudanpop: did you check if this affects the frontend in any way? like displaying and editing step settings?

MarceloRGonc: It doesn't affect.

alexandrudanpop (approved):

PR #1142, opened 2025-08-13
Fix aws-cost mcp in codespaces

rita-gorokhod:


.devcontainer/Dockerfile
```diff
@@ -72,7 +72,7 @@ WORKDIR /root/.mcp/aws-cost
 RUN <<-```
     set -ex
     git clone https://github.com/awslabs/mcp.git .
-    git checkout 2025.6.2025131704
+    git checkout 2025.08.20250807085838
```
rita-gorokhod, line 75: https://github.com/openops-cloud/openops/blob/main/Dockerfile#L40

PR #1144, opened 2025-08-13
Allow higher values in storage block

MarceloRGonc:
Fixes OPS-2371.

### DIFF
--- a/chromatic.config.json
+++ b/chromatic.config.json
@@ -8,5 +8,6 @@
     "packages/ui-components/src/tailwind.css",
     "packages/ui-components/tailwind.base.config.cjs"
   ],
-  "onlyChanged": true
+  "onlyChanged": true,
+  "autoAcceptChanges": "main"
 }
--- a/packages/react-ui/src/app/features/ai/lib/assistant-ui-chat-hook.ts
+++ b/packages/react-ui/src/app/features/ai/lib/assistant-ui-chat-hook.ts
@@ -202,6 +202,7 @@ export const useAssistantChat = (props: UseAssistantChatProps) => {
   }, [chatId, isLoading, openChatResponse?.messages]);
 
   const runtime = useAISDKRuntime(chat);
+  runtimeRef.current = runtime;
 
   const createNewChat = useCallback(async () => {
     const oldChatId = chatId;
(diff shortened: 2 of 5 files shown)


packages/engine/src/lib/services/storage.service.ts
```diff
@@ -53,8 +56,12 @@ export const createStorageService = ({
 
       try {
         const sizeOfValue = sizeof(request.value);
-        if (sizeOfValue > STORE_VALUE_MAX_SIZE) {
-          throw new StorageLimitError(request.key, STORE_VALUE_MAX_SIZE);
+        const maxStorageSizeMB =
```
rita-gorokhod: minor: we can extract it to a var like it was before (here and not in the package), just to avoid the calculation each time and shorten the code here

packages/shared/src/lib/store-entry/store-entry.ts
```diff
@@ -1,11 +1,10 @@
-import { BaseModel } from '../common/base-model';
+import { BaseModel } from '../common';
 import { OpenOpsId } from '../common/id-generator';
-import { ProjectId } from '../project/project';
+import { ProjectId } from '../project';
 
 export type StoreEntryId = OpenOpsId;
 
 export const STORE_KEY_MAX_LENGTH = 128;
```
rita-gorokhod, line 7: Why not have it here actually, along with the STORE_KEY_MAX_LENGTH const?

MarceloRGonc: I don't think this file should exist. At least on the shared package

packages/engine/src/lib/services/storage.service.ts
MarceloRGonc: I changed the way we create variables. We only perform the calculation once.

packages/server/shared/src/lib/fastify.ts
```diff
@@ -1,21 +1,27 @@
...
       `System property OPS_${AppSystemProp.REQUEST_BODY_LIMIT} is not a valid integer`,
     );
   }
-  return ONE_MB_IN_BYTES * bodyLimitInMb;
-};
+
+  return value;
+})();
+
+export const MAX_REQUEST_BODY_MB = REQUEST_BODY_LIMIT_MB;
```
rita-gorokhod: export const MAX_REQUEST_BODY_MB = REQUEST_BODY_LIMIT_MB;

What's the diff between them? they're even called identically

rita-gorokhod: @MarceloRGonc I think it's leftovers, it doesn't seem to be used

MarceloRGonc: Removed

PR #1149, opened 2025-08-13
Fix chromatic diffs

alexandrudanpop:
Fixes OPS-2358


packages/ui-components/chromatic.config.json
```diff
@@ -0,0 +1,5 @@
+{
+  "autoAcceptChanges": "main",
+  "onlyChanged": true,
+  "diffThreshold": 0.2
```
alexandrudanpop, line 4: I'm setting a higher diff threshold than default, hopefully we will not have more flacky diffs
https://www.chromatic.com/docs/threshold/#why-werent-diffs-found-when-a-color-or-background-changed

packages/ui-components/src/styles/global.css
```diff
@@ -12,3 +12,12 @@ body {
...
   opacity: 0.9 !important;
 }
+
+/* Hide scrollbars in Chromatic to prevent diff issues */
+.monaco-editor .scrollbar {
+  display: none !important;
+}
+
+.monaco-editor .monaco-scrollable-element > .scrollbar {
+  display: none !important;
```
alexandrudanpop, line 22: This might be a timing issue as scrollbars are not always shown. This CSS is only used in storybook / chromatic

PR #1154, opened 2025-08-14
Display error messages in the conversation

MarceloRGonc:
Fixes OPS-2385.

image

### DIFF
--- a/packages/react-ui/src/app/features/templates/components/select-flow-template-dialog.tsx
+++ b/packages/react-ui/src/app/features/templates/components/select-flow-template-dialog.tsx
@@ -27,6 +27,7 @@ import {
   StepPlaceHolder,
   TemplateDetails,
   TemplateEdge,
+  TemplateSidebarCategory,
   Theme,
   toast,
   VerticalDivider,
@@ -37,7 +38,7 @@ import {
   FlowTemplateDto,
 } from '@openops/shared';
 import { useMutation } from '@tanstack/react-query';
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { useDebounceValue } from 'usehooks-ts';
 import { blocksHooks } from '../../blocks/lib/blocks-hook';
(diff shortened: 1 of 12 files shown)


packages/server/api/src/app/ai/chat/chat-request-router.ts
```diff
@@ -1,6 +1,6 @@
 import { logger } from '@openops/server-shared';
 import { CODE_BLOCK_NAME, NewMessageRequest, Principal } from '@openops/shared';
-import { CoreMessage } from 'ai';
+import { ModelMessage } from 'ai';
```
MarceloRGonc, line 3: The old value was deprecated

packages/server/api/src/app/ai/chat/message-extractor.ts
```diff
@@ -0,0 +1,35 @@
...
+  fallback = 'An unknown error occurred. Please start a new chat.',
+): string {
+  if (error instanceof Error) {
+    return error.message;
+  }
+
+  const messageProperty = findFirstMessageProperty(error);
+  if (messageProperty) {
+    return messageProperty;
+  }
```
MarceloRGonc: New condition.

packages/server/api/src/app/ai/chat/message-extractor.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  if (typeof error === 'string') {
+    return error.trim() || fallback;
+  }
+
+  return fallback;
+}
+
+function findFirstMessageProperty(obj: unknown): string | undefined {
```
MarceloRGonc: This is a new method that searches for a message property

packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -280,6 +280,7 @@ function sendMessageToStream(
     case 'finish-step':
     case 'finish':
     case 'tool-input-end':
+    case 'error':
```
MarceloRGonc, line 283: We don't want to send the error message. We are intersecting the message and we just send the string as text

packages/server/api/src/lib/error-utils.ts
```diff
@@ -1,31 +0,0 @@
...
-  if (typeof error === 'object' && error !== null && 'message' in error) {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const message = (error as any).message;
-    if (typeof message === 'string' && message.trim()) {
-      return message.trim();
-    }
-  }
-
-  return fallback;
-}
```
MarceloRGonc, line 31: Used in the message-extractor

packages/server/api/test/unit/ai/model-message-converter.test.ts
MarceloRGonc, line 1: Moved

PR #1163, opened 2025-08-18
Fix OpenAI structured outputs for generateObject

alexandrudanpop:
Fixes OPS-2342


packages/openops/src/lib/ai/providers.ts
```diff
@@ -167,7 +167,5 @@ function sanitizeProviderSettings(
 }
 
 export const providerOptions = {
-  openai: {
-    structuredOutputs: false,
-  },
+  openai: {},
 };
```
MarceloRGonc: If we are removing this flag, we can remove the object.

PR #1166, opened 2025-08-19
Add workflow job to add release label to Linear issues

ravikiranvm:
Fixes CI-23.

## Things to Note
- This action can only be run before creating a new release, because it would only find PRs made after the latest release available.
- This job requires a ticket prefix. I checked if we could run the job twice, once for OPS and once for CI. But the second job would always fail because the code tries to recreate a version label in Linear and then it would fail. So this job would only tag OPS related issues. 
- This job runs just before the "Create release" job. There may be instances when a draft release is postponed and then skipped for another draft release. Then there is a chance that PRs after last release to be re-tagged with latest version label. 

## Testing Checklist

Check all that apply:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,10 @@ on:
         type: boolean
         description: Deploy the release to the test environment
         default: true
+      label_linear_issues:
+        type: boolean
+        description: Label Linear issues with the release version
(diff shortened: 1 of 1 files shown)


ravikiranvm:
Creating an OpenOps workflow for this.

PR #1167, opened 2025-08-19
Use test data viewer for tools outputs display

alexandrudanpop:
Fixes OPS-2402

### DIFF
--- a/packages/ui-components/src/components/assistant-ui/tool-fallback/tool-fallback.tsx
+++ b/packages/ui-components/src/components/assistant-ui/tool-fallback/tool-fallback.tsx
@@ -32,18 +32,15 @@ export const ToolFallback = ({
 
   return (
     <BaseToolWrapper toolName={toolName} status={resultStatus}>
-      <div className="flex flex-col gap-4 pt-2">
-        <div className="px-4">
-          <div className="h-72">
-            <TestStepDataViewer
-              inputJson={formattedArgs}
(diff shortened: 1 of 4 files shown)


packages/ui-components/src/components/ai-chat-container/assistant-ui-chat-container.tsx
```diff
@@ -33,6 +33,9 @@ const AssistantUiResizableContainer = ({
         },
         className,
       )}
+      style={{
+        minWidth: `${minWidth}px`,
+      }}
```
alexandrudanpop, line 38: weird bug that makes the chatbox very small when switching from larger to smaller display

packages/ui-components/src/components/assistant-ui/tool-fallback/tool-fallback.tsx
```diff
@@ -1,35 +1,45 @@
...
   argsText,
   result,
   status,
-}) => {
+  theme,
+}: ToolFallbackProps) => {
   const resultStatus = extractResultStatus(result, status);
 
+  const formattedArgs = tryParseJson(argsText);
+  const formattedResult = formatToolResultForDisplay(result);
```
rmn-snpk: Shouldn't we add memorization here?

alexandrudanpop: done

PR #1168, opened 2025-08-19
Add type column to flow table

bigfluffycookie:
Fixes OPS-2417

### DIFF
--- a/packages/engine/test/resolve-variable.test.ts
+++ b/packages/engine/test/resolve-variable.test.ts
@@ -330,4 +330,4 @@ describe('resolveVariable', () => {
       expect(result.censoredValue).toBe(censoredValue);
     });
   });
-}); 
+});


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -74,6 +75,7 @@ export const flowService = {
       projectId,
       request: {
         displayName,
+        type: FlowType.FLOW,
```
bigfluffycookie, line 78: when created from a template we can just set the type to flow

packages/server/api/src/app/flows/flow-version/flow-version.service.ts
```diff
@@ -291,6 +292,7 @@ export const flowVersionService = {
     request: {
       displayName: string;
       description: string;
+      type?: FlowType;
```
bigfluffycookie: this and other type fields will be made mandatory in a later pr https://linear.app/openops/issue/OPS-2432/make-type-value-in-flow-creation-mandatory

bigfluffycookie:
Closing because we are slightly changing the implementation and moving it to a different repository.

PR #1170, opened 2025-08-19
Fix handling of empty `attributeName` in EC2 modify instance action

rSnapkoOpenOps:
Fixes OPS-2365.

https://www.loom.com/share/d8759e0eac214d8a895e931bf2455bd4?sid=40bf5d79-b242-454d-8ecd-7e4b4be1c47f


rita-gorokhod:
@rSnapkoOpenOps I think it's ok, but we should think about how to guard from it on the frontend, it's very likely that this will happen again.

rmn-snpk:
@rita-gorokhod https://linear.app/openops/issue/OPS-2441/investigate-dynamic-property-schema-protection

PR #1176, opened 2025-08-21
Add more schema descriptions

alexandrudanpop:
Part of OPS-2337, OPS-2442

### DIFF
--- a/packages/shared/src/lib/flows/actions/action.ts
+++ b/packages/shared/src/lib/flows/actions/action.ts
@@ -24,7 +24,7 @@ const commonActionProps = {
   }),
   name: Type.String({
     description:
-      'Name of the step (action), used for reference in other step configurations. The name should always be in the format: "step_{autoincrement}", example: "step_1", "step_2", etc.',
+      'Name of the step (action), used for reference in other step configurations.',
   }),
   valid: Type.Boolean({
     description:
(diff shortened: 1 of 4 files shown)


packages/shared/src/lib/flows/flow-operations.ts
```diff
@@ -129,6 +130,17 @@ export const AddActionRequest = Type.Object({
 });
 export type AddActionRequest = Static<typeof AddActionRequest>;
 
+export const AddActionRequestWithoutId = Type.Composite([
+  AddActionRequest,
+  Type.Object({
+    action: Type.Omit(UpdateActionRequest, ['id']),
+  }),
+]);
```
alexandrudanpop: needed for OPS-2442

packages/shared/src/lib/flows/actions/action.ts
```diff
@@ -24,7 +24,7 @@ const commonActionProps = {
   }),
   name: Type.String({
     description:
-      'Name of the step (action), used for reference in code or configurations. Usually the name is in the format step_index',
+      'Name of the step (action), used for reference in other step configurations. The name should always be in the format: "step_{autoincrement}", example: "step_1", "step_2", etc.',
```
rita-gorokhod: @alexandrudanpop do we need this? step_{autoincrement} ? it can absolutely be {{step_stepId}}. Can also be {{my_step}}

alexandrudanpop: This was how they were added from the UI. Shouldn't we keep it consistent for the user (because they will see it when it's referenced)?
We could add this logic in the tool instead of relying on the LLM.

https://github.com/openops-cloud/openops/blob/6ef1582448b2eb7631740babc814d25e6e9fd05e/packages/react-ui/src/app/features/builder/blocks-selector/block-selector-utils.ts#L36-L47

alexandrudanpop: fixed, the tool will do it.

packages/shared/src/lib/flows/flow-operations.ts
MarceloRGonc: I think this should be in a different file and not sure if in this repo. It will only be used on the new routes right?

MarceloRGonc (approved):

PR #1177, opened 2025-08-21
Fix scheduling triggers not respecting the defined time

MarceloRGonc:
Fixes OPS-2449.

### DIFF
--- a/packages/server/api/src/app/flows/trigger/hooks/enable-trigger-hook.ts
+++ b/packages/server/api/src/app/flows/trigger/hooks/enable-trigger-hook.ts
@@ -17,6 +17,7 @@ import {
   OpsEdition,
   ProjectId,
   RunEnvironment,
+  ScheduleOptions,
   TriggerHookType,
   TriggerType,
 } from '@openops/shared';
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/flows/trigger/hooks/enable-trigger-hook.ts
```diff
@@ -150,6 +149,29 @@ export const enableBlockTrigger = async (
       });
       break;
     }
+    case TriggerStrategy.SCHEDULED: {
+      if (isNil(engineHelperResponse.result.scheduleOptions)) {
+        throw new Error(
+          'No schedule options set for trigger of Schedule type.',
+        );
```
MarceloRGonc, line 150: As discussed offline, we should throw in this case. We don't want schedule types with default values.

packages/server/api/src/app/flows/trigger/hooks/enable-trigger-hook.ts
```diff
@@ -150,6 +149,29 @@ export const enableBlockTrigger = async (
...
+      }
+
+      await flowQueue.add({
```
rita-gorokhod: Extract to func please.
And I see the webhook is the same too, just needs jobType and scheduleOptions (and maybe environment?)

MarceloRGonc: Extracted. But the webhook uses a different data object

rita-gorokhod (approved):

PR #1175, opened 2025-08-20
Rename 'End workflow' action with 'Stop Execution"

ravikiranvm:
Fixes OPS-2389.

## Additional Notes
Screenshot 2025-08-20 at 4 53 02 PM

## Testing Checklist

Check all that apply:

- [ ] I tested the feature thoroughly, including edge cases

- [ ] I verified all affected areas still work as expected

- [ ] Automated tests were added/updated if necessary

- [ ] Changes are backwards compatible with any existing data, otherwise a migration script is provided

## Visual Changes (if applicable)

If there are UI/UX changes, include at least one of the following:

- Screenshots
- Loom video
- Preview deployment link

PR #1180, opened 2025-08-21
Revert "end workflow" action name

rita-gorokhod:
Fixes OPS-2389


packages/blocks/end-flow/src/lib/actions/end-workflow.ts
```diff
@@ -1,7 +1,7 @@
 import { createAction } from '@openops/blocks-framework';
 
 export const stopFlowAction = createAction({
-  name: 'stop_flow',
+  name: 'end_workflow',
```
rita-gorokhod, line 4: FYI this caused a regression https://github.com/openops-cloud/openops/pull/1175/files

You cannot rename names of properties/actions/blocks without adding a migration script.
The changes can only be made to display names.

PR #1188, opened 2025-08-24
Fix paginator cursor calculation issues for forward and backward navigation 

cezudas:
Fixes OPS-1227.

Demo of the cursor calculation issues (before): https://www.loom.com/share/b49df91ca94d4d7b957675d58b2d6c92

### DIFF
--- a/packages/server/api/src/app/flows/flow/flow.service.ts
+++ b/packages/server/api/src/app/flows/flow/flow.service.ts
@@ -129,6 +129,11 @@ export const flowService = {
         afterCursor: decodedCursor.nextCursor,
         beforeCursor: decodedCursor.previousCursor,
       },
+      customPaginationColumn: {
+        columnPath: 'versions[0].updated',
+        columnName: 'fv.updated',
+        columnType: 'timestamp with time zone',
+      },
     });
 
     const queryWhere: Record<string, unknown> = { projectId };
@@ -159,7 +164,7 @@ export const flowService = {
             .getQuery() +
           ')',
       )
-      .orderBy('fv.updated', 'DESC');
+      .addOrderBy('fv.updated', 'DESC');
 
     if (name) {
       query = query.andWhere('fv.displayName ILIKE :namePattern', {
--- a/packages/server/api/src/app/flows/flow/flow.service.ts
+++ b/packages/server/api/src/app/flows/flow/flow.service.ts
@@ -163,8 +163,7 @@ export const flowService = {
             .limit(1)
             .getQuery() +
           ')',
-      )
-      .addOrderBy('fv.updated', 'DESC');
+      );
 
     if (name) {
       query = query.andWhere('fv.displayName ILIKE :namePattern', {
--- a/packages/server/api/src/app/helper/pagination/pagination-utils.ts
+++ b/packages/server/api/src/app/helper/pagination/pagination-utils.ts
@@ -110,7 +110,7 @@ export function getValueByPath(entity: unknown, path: string): unknown {
       current = current?.[part];
     }
 
-    if (current === undefined || current === null) {
+    if (!current) {
       return null;
     }
   }
--- a/packages/server/api/src/app/helper/pagination/pagination-utils.ts
+++ b/packages/server/api/src/app/helper/pagination/pagination-utils.ts
@@ -2,6 +2,11 @@ import { SeekPage } from '@openops/shared';
 import dayjs from 'dayjs';
 import { CursorResult } from './paginator';
 
+const ARRAY_OPEN_BRACKET = '[';
+const ARRAY_CLOSE_BRACKET = ']';
+const PATH_SEPARATOR = '.';
+const PARSE_INT_RADIX = 10;
+
 export function atob(value: string): string {
   return Buffer.from(value, 'base64').toString();
 }
@@ -54,7 +59,7 @@ export function decodeByType(
     case 'timestamp with time zone':
     case 'datetime':
     case 'date': {
-      const timestamp = parseInt(value, 10);
+      const timestamp = parseInt(value, PARSE_INT_RADIX);
       if (Number.isNaN(timestamp)) {
         throw new Error('date column in cursor should be a valid timestamp');
       }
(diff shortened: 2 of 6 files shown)


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -129,6 +129,11 @@ export const flowService = {
         afterCursor: decodedCursor.nextCursor,
         beforeCursor: decodedCursor.previousCursor,
       },
+      customPaginationColumn: {
```
cezudas, line 132: Default paginator cursor is entity field`created`, but we have business requirements to order flows by "Last updated" column, so this PR adds support for aligning custom display order with pagination cursors so they both use the same sort key.

packages/server/api/src/app/helper/pagination/paginator.ts
```diff
@@ -133,10 +159,21 @@ export default class Paginator<Entity extends ObjectLiteral> {
...
+      ? this.paginationColumnName
+      : `${this.alias}.${PAGINATION_KEY}`;
+    const paramName = isCustomColumn ? CUSTOM_PAGINATION_KEY : PAGINATION_KEY;
+
     if (dbType === DatabaseType.SQLITE3) {
-      queryString = `strftime('%s', ${this.alias}.${PAGINATION_KEY}) ${operator} strftime('%s', :${PAGINATION_KEY})`;
+      queryString = `strftime('%s', ${columnName}) ${operator} strftime('%s', :${paramName})`;
     } else if (dbType === DatabaseType.POSTGRES) {
-      queryString = `DATE_TRUNC('second', ${this.alias}.${PAGINATION_KEY}) ${operator} DATE_TRUNC('second', :${PAGINATION_KEY}::timestamp)`;
+      if (this.hasBeforeCursor() && !this.hasAfterCursor()) {
```
cezudas, line 172: When going backward, the order gets flipped, but the `WHERE` clause still uses `DATE_TRUNC` while the `ORDER BY` uses the full timestamp.
This creates a mismatch:
WHERE clause: `DATE_TRUNC('second', created) > DATE_TRUNC('second', '2025-08-23T08:51:00.880Z')`
ORDER BY: `created DESC` (but order gets flipped to `ASC` for backward pagination)

packages/server/api/src/app/helper/pagination/paginator.ts
```diff
@@ -133,10 +159,21 @@ export default class Paginator<Entity extends ObjectLiteral> {
...
+        queryString = `${columnName} ${operator} (:${paramName}::timestamp + INTERVAL '1 millisecond')`;
```
cezudas, line 173: Without the 1ms offset, if you have multiple rows with the same timestamp at the second level:
```
2025-08-23T08:51:00.456Z (cursor)
2025-08-23T08:51:00.456Z (another row)
2025-08-23T08:51:00.456Z (another row)
```
The backward pagination `WHERE created > '2025-08-23T08:51:00.456Z'` would skip all rows because no timestamp is greater than itself.

packages/server/api/src/app/helper/pagination/paginator.ts
```diff
@@ -133,10 +159,21 @@ export default class Paginator<Entity extends ObjectLiteral> {
...
+    const isCustomColumn =
+      this.paginationColumnName && cursors[CUSTOM_PAGINATION_KEY];
+    const columnName = isCustomColumn
+      ? this.paginationColumnName
+      : `${this.alias}.${PAGINATION_KEY}`;
+    const paramName = isCustomColumn ? CUSTOM_PAGINATION_KEY : PAGINATION_KEY;
+
     if (dbType === DatabaseType.SQLITE3) {
-      queryString = `strftime('%s', ${this.alias}.${PAGINATION_KEY}) ${operator} strftime('%s', :${PAGINATION_KEY})`;
+      queryString = `${columnName} ${operator} :${paramName}`;
```
cezudas, line 170: Notice the query syntax is different between Postgres and SQLite.
My integration test uses SQLite. This change ensures that it is possible to test backward pagination when multiple rows have the same timestamp at the second level. Look for `1ms Offset Logic` in the tests

packages/server/api/src/app/helper/pagination/pagination-utils.ts
```diff
@@ -54,11 +59,11 @@ export function decodeByType(
     case 'timestamp with time zone':
     case 'datetime':
     case 'date': {
-      const timestamp = parseInt(value, 10);
+      const timestamp = parseInt(value, PARSE_INT_RADIX);
       if (Number.isNaN(timestamp)) {
         throw new Error('date column in cursor should be a valid timestamp');
       }
-      return new Date(timestamp).toISOString();
+      return new Date(timestamp);
```
MarceloRGonc, line 66: Why this change?

cezudas: The previous syntax only worked for Postgres**(read more at the end) but not for SQLite (used by the new tests).

By returning `new Date(timestamp)` instead of` new Date(timestamp).toISOString()`, we let TypeORM handle the database-specific formatting:
SQLite: TypeORM converts Date → "YYYY-MM-DD HH:MM:SS.mmm"
PostgreSQL: TypeORM converts Date → "YYYY-MM-DD HH:MM:SS.mmm+00"

Before:
```
// 1. Database stores timestamp as:
"2025-01-01 09:00:00.000"  // SQLite local datetime format

// 2. Cursor decodes to:
"2025-01-01T09:00:00.000Z"  // ISO UTC string format

// 3. When TypeORM compares:
WHERE created > "2025-01-01T09:00:00.000Z"
//  This is a STRING comparison, not a date comparison!
// SQLite sees: "2025-01-01 09:00:00.000" > "2025-01-01T09:00:00.000Z"
// Result: Incorrect because it's comparing strings character by character
```

After:
```
// 1. Database stores timestamp as:
"2025-01-01 09:00:00.000"  // SQLite local datetime format

// 2. Cursor decodes to:
new Date(timestamp)  // JavaScript Date object

// 3. When TypeORM compares:
WHERE created > ?  // TypeORM passes the Date object as a parameter
// TypeORM converts the Date object to the database's expected format
// SQLite gets: "2025-01-01 09:00:00.000" > "2025-01-01 09:00:00.000"
// Result: Correct date comparison!
```

------------------------------------

**
```
-- PostgreSQL is smart about type conversion
WHERE created > '2025-01-01T09:00:00.000Z'::timestamp
-- PostgreSQL automatically converts ISO string to timestamp type
-- The ::timestamp cast handles the format conversion properly
--> Correct date comparison
```
**

PR #1190, opened 2025-08-25
Get projectIds from organization ID

bigfluffycookie:
Fixes OPS-2429

### DIFF
--- a/packages/server/api/src/app/project/project-service.ts
+++ b/packages/server/api/src/app/project/project-service.ts
@@ -72,6 +72,21 @@ export const projectService = {
     );
     return organizationId;
   },
+
+  async getProjectIdsByOrganizationId(
+    organizationId: string,
+  ): Promise<ProjectId[]> {
+    const results = await projectRepo()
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/project/project-service.ts
```diff
@@ -72,6 +72,21 @@ export const projectService = {
...
+    const results = await projectRepo()
+      .createQueryBuilder('project')
+      .select('"id"')
+      .where({
+        organizationId,
+        deleted: IsNull(),
+      })
+      .getRawMany();
+    return results.map((result) => result.id);
+  },
```
MarceloRGonc:
```suggestion
  async getProjectIdsByOrganizationId(
    organizationId: string,
  ): Promise<ProjectId[]> {
    const projects = await projectRepo().find({
      select: ['id'],
      where: { organizationId, deleted: IsNull() },
    });

    return projects.map((project) => project.id);
  },
```

MarceloRGonc (approved):

PR #1191, opened 2025-08-25
Migrate step output table to new algorithm

MarceloRGonc:
Fixes OPS-2460.


packages/server/api/src/app/flows/step-run/step-run-service.ts
```diff
@@ -45,10 +45,10 @@ export const stepRunService = {
 
     const outputs = await flowStepTestOutputService.listEncrypted({
       flowVersionId: flowVersion.id,
-      stepIds,
+      stepIds: stepIds.filter((item) => item !== step.id),
```
MarceloRGonc, line 48: We don't need the current step output when we are testing a step

packages/server/api/src/app/flows/step-test-output/flow-step-test-output-entity.ts
```diff
@@ -18,11 +18,11 @@ export const FlowStepTestOutputEntity =
       stepId: OpenOpsIdSchema,
       flowVersionId: OpenOpsIdSchema,
       input: {
-        type: BLOB_COLUMN_TYPE,
+        type: JSONB_COLUMN_TYPE,
```
MarceloRGonc, line 21: The new object saved is a json with the encrypted data:

```
{
  'iv': ...
  'data': ...
}

File: packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts [hunk 3]
```diff
@@ -22,15 +23,17 @@ export const flowStepTestOutputService = {
...
-      compressedOutput = await encryptAndCompress(output);
+    if (input === undefined) {
+      input = Buffer.alloc(0);
     }
-    if (input !== undefined) {
-      compressedInput = await encryptAndCompress(input);
+
+    if (output === undefined) {
+      output = Buffer.alloc(0);
     }
```
MarceloRGonc, line 32: If what we want to save is undefined, we save it as an empty buffer. We do this because the output of a block can be undefined

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
```diff
@@ -153,15 +154,15 @@ async function getEngineToken(projectId: string): Promise<string> {
...
+): Promise<Record<OpenOpsId, EncryptedStepOutput>> {
   try {
     const stepIds = flowHelper.getAllStepIds(flow.version.trigger);
     const outputs = await flowStepTestOutputService.listEncrypted({
       flowVersionId,
       stepIds,
     });
 
-    return encodeStepOutputs(outputs);
+    return convertStepOutputs(outputs);
```
rita-gorokhod: I think this is not a good name, convert to what?

packages/server/api/src/app/database/postgres-connection.ts
```diff
@@ -68,6 +69,7 @@ const getMigrations = (): (new () => MigrationInterface)[] => {
     AddSuccessToFlowStepTestOutputTable1752487641303,
     UpdateUmbrellaRecommendationTypesAcrossFlowVersions1752758891771,
     AddTriggerSourceToFlowRun1754489349007,
+    MigrateStepTestOutputToNewAlgorithm1755942417309,
```
rita-gorokhod: nitpicking: toNewEncryptionAlgorithm

packages/server/api/src/app/flows/step-test-output/flow-step-test-output-entity.ts
rita-gorokhod: will this create a problem for templates that are on older versions?

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
rita-gorokhod: and if it's null?

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -116,6 +121,19 @@ async function decompressOutput(
   };
 }
 
+function isEmptyBuffer(val: unknown): boolean {
```
rita-gorokhod, line 124: don't we have a utility func for this?

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
```diff
@@ -116,6 +121,19 @@ async function decompressOutput(
...
+  return (
+    typeof val === 'object' &&
+    val !== null &&
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (val as any).type === 'Buffer' &&
```
rita-gorokhod, line 129: would the type be a buffer if it's empty?

packages/server/api/src/app/flows/step-test-output/flow-step-test-output-entity.ts
MarceloRGonc: We don't import the step output with the templates.

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
MarceloRGonc: If it's null is saved as null. The algorithm works with null

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
MarceloRGonc: WDYM? Empty string as output?

packages/server/api/src/app/database/postgres-connection.ts
MarceloRGonc: MigrateStepTestOutputToNewCompressionAlgorithm
What changed was the compression and the order

packages/server/api/src/app/ai/chat/context-enrichment.service.ts
MarceloRGonc: Changed to groupStepOutputsById

packages/server/api/src/app/flows/step-test-output/flow-step-test-output.service.ts
MarceloRGonc: I didn't find any

PR #1197, opened 2025-08-25
Send telemetry for executed workflows

MarceloRGonc:
Fixes OPS-2356.


packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -157,6 +160,17 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
     app.io
       .to(populatedRun.projectId)
       .emit(WebsocketClientEvent.FLOW_RUN_PROGRESS, runId);
+
+    if (
+      runDetails.status !== FlowRunStatus.RUNNING &&
```
rita-gorokhod: I think we should have a utility func next to FlowRunStatus enum, "IsFlowRunFinished", and use it.
The various statuses we have are way too error prone.

PR #1200, opened 2025-08-26
Fix trigger description

alexandrudanpop:
Fixes OPS-2337


packages/shared/src/lib/flows/triggers/trigger.ts
```diff
@@ -37,7 +37,7 @@ export const BlockTriggerSettings = Type.Object({
   packageType: Type.Enum(PackageType),
   triggerName: Type.Optional(
     Type.String({
-      description: 'The name of the trigger should always be "trigger".',
+      description: 'The name of the trigger',
     }),
   ),
```
alexandrudanpop, line 42: Incorrect because this is actually the trigger action name (eg: triggerName: "every_week")

cezudas (approved):

PR #1203, opened 2025-08-26
Reduce saved data in the loop-executor

MarceloRGonc:
Fixes OPS-2483.

## Additional notes:
- Remove clonedeep
- Do not save the iteration item
- Reduces the requests to storage to one get and one put

## Performed Tests
- Pause before the loop 🟢 
- Pause after the loop 🟢
- Pause inside the loop 🟢
- Pause before, inside and after the loop 🟢
- Pause before and inside the loop 🟢
- Pause inside and after the loop 🟢
- Pause before and after the loop 🟢
- Pause before, inside an inner loop and after the loop 🟢


packages/engine/src/lib/handler/loop-executor-old-resume.ts
MarceloRGonc, line 1: This is the old resume strategy. I think we can remove it in a few versions, but then we lined it up.

MarceloRGonc:
Here are the results of some performance tests. I used the workflow below, with **5 parallel runs**.

image


**First test:** 5 executions in parallel, and the loop has 20 iterations:

| 0.4.6    | This Branch |
|----------|-------------|
| 56456 ms | 14207 ms   |
| 60236 ms | 14187 ms    |
| 61078 ms | 15485 ms    |
| 53667 ms | 16387 ms   |
| 45740 ms | 14161 ms    |

**Second test:** 5 executions in parallel, and the loop has 40 iterations:

| 0.4.6    | This Branch |
|----------|-------------|
| OOM | 13515 ms   |
| OOM | 14244 ms    |
| OOM | 15571 ms    |
| OOM | 16441 ms   |
| OOM | 14524 ms    |

packages/engine/src/lib/handler/loop-executor-old-resume.ts
MarceloRGonc, line 1: This is the old resume strategy. I think we can remove it in a few versions, but then we lined it up.

packages/engine/src/lib/handler/loop-executor.ts
```diff
@@ -114,25 +122,46 @@ export const loopExecutor: BaseExecutor<LoopOnItemsAction> = {
...
+        iterationsMapping,
+        firstLoopAction,
+        resolvedInput,
+        constants,
+        payload,
+        store,
+      });
+    }
+
+    return resumePausedIterationOldStrategy(
```
MarceloRGonc, line 156: If not iterationsMapping when resuming apply the old strategy

bigfluffycookie (commented):
Looks good, did you test it with a very large list?

bigfluffycookie (approved):

PR #1207, opened 2025-08-27
Add new column to flow table

bigfluffycookie:
Fixes OPS-2488


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -660,6 +662,7 @@ type CreateParams = {
   userId: UserId;
   projectId: ProjectId;
   request: CreateEmptyFlowRequest;
+  isWorkflow?: boolean;
```
bigfluffycookie, line 669: made it optional, realistically we dont wanna change all the calls to this, so only set it when you want it to be false

PR #1213, opened 2025-08-28
Resume conversation after FE tool calls

alexandrudanpop:
Fixes OPS-2359


packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -379,7 +381,7 @@ async function getUserMessage(
       return null;
     }
 
-    return String(firstContentElement.text);
+    return String(firstContentElement.text ?? 'continue');
```
alexandrudanpop, line 384: If it's a frontend tool, we just say `continue` to the conversation (this will be shown as a user message if the user reloads).

cezudas (approved):

PR #1216, opened 2025-08-29
Fix not updating the run if it is already in a final state

MarceloRGonc:
Fixes #<issue_number>.

### DIFF
--- a/packages/engine/src/lib/operations.ts
+++ b/packages/engine/src/lib/operations.ts
@@ -34,26 +34,32 @@ import { flowExecutor } from './handler/flow-executor';
 import { blockHelper } from './helper/block-helper';
 import { triggerHelper } from './helper/trigger-helper';
 import { resolveVariable } from './resolve-variable';
+import { progressService } from './services/progress.service';
 import { utils } from './utils';
 
 const executeFlow = async (
   input: ExecuteFlowOperation,
   context: FlowExecutorContext,
 ): Promise<EngineResponse<Pick<FlowRunResponse, 'status' | 'error'>>> => {
-  const constants = EngineConstants.fromExecuteFlowInput(input);
+  try {
+    const constants = EngineConstants.fromExecuteFlowInput(input);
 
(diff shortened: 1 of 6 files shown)


packages/engine/src/lib/operations.ts
```diff
@@ -34,26 +34,32 @@ import { flowExecutor } from './handler/flow-executor';
...
+
+    return {
+      status: EngineResponseStatus.OK,
+      response: {
+        status: response.status,
+        error: response.error,
+      },
+    };
+  } finally {
+    await progressService.flushProgressUpdate(input.flowRunId);
```
MarceloRGonc: Moved because if any error occurs, we want to save the latest progress

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -26,13 +26,7 @@ function getRunDebouncer(runId: string): any {
   if (!debouncedFunc) {
     debouncedFunc = debounce(
       async (latestParams: UpdateStepProgressParams) => {
-        try {
-          await sendUpdateRunRequest(latestParams);
-        } finally {
-          if (runDebouncers.get(runId) === debouncedFunc) {
-            runDebouncers.delete(runId);
-          }
-        }
```
MarceloRGonc: Moved the delete. We always do the flush.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -278,25 +279,47 @@ export const flowRunService = {
...
+      logger.debug(
+        `Update for workflow run (${flowRunId}) skipped. The workflow run already in a final state.`,
+        {
+          flowRunId,
+          newStatus: status,
+        },
+      );
+
+      return undefined;
+    }
```
MarceloRGonc, line 321: The biggest change. If the run state in the DB is already final, we won't switch to another state.

bigfluffycookie (approved):

PR #1218, opened 2025-08-29
Avoid two progress updates for the same run in parallel

MarceloRGonc:
Fixes OPS-2508.


packages/engine/src/lib/operations.ts
```diff
@@ -34,27 +34,33 @@ import { flowExecutor } from './handler/flow-executor';
...
+    return {
+      status: EngineResponseStatus.OK,
+      response: {
+        status: response.status,
+        error: response.error,
+      },
+    };
+  } finally {
+    await progressService.flushProgressUpdate(input.flowRunId);
+  }
```
MarceloRGonc, line 63: Moved because if any error occurs, we want to save the latest progress

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -44,21 +50,35 @@ function getRunDebouncer(runId: string): any {
...
 
 export const progressService = {
-  sendUpdate: (params: UpdateStepProgressParams): void => {
+  sendUpdate: async (params: UpdateStepProgressParams): Promise<void> => {
     throwIfExecutionTimeExceeded();
 
-    void getRunDebouncer(params.engineConstants.flowRunId)(params);
+    const lock = await memoryLock.acquire(params.engineConstants.flowRunId);
+    updateParameters.set(params.engineConstants.flowRunId, params);
+    await lock.release();
```
MarceloRGonc, line 58: Using the lock to set/override the parameters

packages/engine/src/lib/services/progress.service.ts
```diff
@@ -17,21 +22,22 @@ import {
...
   let debouncedFunc = runDebouncers.get(runId);
 
   if (!debouncedFunc) {
     debouncedFunc = debounce(
-      async (latestParams: UpdateStepProgressParams) => {
+      async () => {
+        const lock = await memoryLock.acquire(runId);
         try {
-          await sendUpdateRunRequest(latestParams);
+          await sendUpdateRunRequest(runId);
```
MarceloRGonc, line 38: The update is performed using the acquired lock. It does not allow parallel requests or overtaking in the order.

packages/engine/test/services/progress.service.test.ts
```diff
@@ -263,9 +263,9 @@ describe('Progress Service', () => {
         },
       };
 
-      progressService.sendUpdate(request);
-      await new Promise(res => setTimeout(res, 801));
-      progressService.sendUpdate(request);
+      await progressService.sendUpdate(request);
+      await new Promise(res => setTimeout(res, 1500));
```
bigfluffycookie, line 267: any particular reason for adjusting the timeout values?

MarceloRGonc: Since we use locks, debouncing can cover a bit more. Otherwise, we'll make almost one request for each method call.

PR #1228, opened 2025-09-01
Add stop endpoint to allow a workflow execution to be stopped

MarceloRGonc:
Fixes OPS-2497

### DIFF
--- a/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
+++ b/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
@@ -25,6 +25,7 @@ import {
   RunEnvironment,
   SeekPage,
   spreadIfDefined,
+  TERMINAL_STATUSES,
 } from '@openops/shared';
 import { nanoid } from 'nanoid';
 import { In, LessThan } from 'typeorm';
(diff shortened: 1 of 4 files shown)


packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -253,6 +258,14 @@ export const flowRunService = {
         },
       });
     }
+
+    ensureFlowResumable({
+      flowRunId,
+      executionType,
+      flowRetryStrategy,
+      flowRunStatus: flowRunToResume.status,
+    });
```
MarceloRGonc: Check if the workflow is in a valid state to be resumed

packages/server/api/src/app/flows/flow-run/flow-run-controller.ts
```diff
@@ -90,6 +94,41 @@ export const flowRunController: FastifyPluginCallbackTypebox = (
...
+
+    if (flowRun.status === FlowRunStatus.PAUSED) {
+      await flowRunRepo().update(flowRunId, {
+        status: FlowRunStatus.ABORTED,
+        finishTime: new Date().toISOString(),
+      });
+
+      app.io
+        .to(flowRun.projectId)
+        .emit(WebsocketClientEvent.FLOW_RUN_PROGRESS, flowRunId);
```
MarceloRGonc: If the workflow is Paused we need to update the DB directly

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -253,6 +258,14 @@ export const flowRunService = {
         },
       });
     }
+
+    verifyResumeEligibility({
```
MarceloRGonc, line 263: Check if the workflow is in a valid state to be resumed.

packages/server/api/src/app/flows/flow-run/flow-run-controller.ts
```diff
@@ -90,6 +94,41 @@ export const flowRunController: FastifyPluginCallbackTypebox = (
...
+    }
+
+    await requestWorkflowCancellation(flowRunId);
+
+    return {
+      success: true,
+      flowRunId,
+    };
```
rmn-snpk, line 129: Could you update the return type to be similar to retry endpoint and return aborted run

MarceloRGonc: As we discussed, you don't need this :)

PR #1231, opened 2025-09-02
Refactor error handling when calling engine

MarceloRGonc:
Part of OPS-2497.

### DIFF
--- a/packages/server/worker/jest.config.ts
+++ b/packages/server/worker/jest.config.ts
@@ -1,0 +1,17 @@
+export default {
+  displayName: '@openops/worker',
+  preset: '../../../jest.preset.js',
+  setupFiles: ['../../../jest.env.js'],
+  globals: {},
+  testEnvironment: 'node',
+  transform: {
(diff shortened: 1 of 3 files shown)


packages/server/worker/src/lib/executors/flow-job-executor.ts
```diff
@@ -221,6 +209,50 @@ async function updateJobStatus(
...
+  | { engineSucceeded: true; failedRunStatus: undefined }
+  | {
+      engineSucceeded: false;
+      failedRunStatus:
+        | FlowRunStatus.TIMEOUT
+        | FlowRunStatus.ABORTED
+        | FlowRunStatus.INTERNAL_ERROR;
+    };
+
+function evaluateEngineStatus(
```
bigfluffycookie, line 222: please add a test for this

MarceloRGonc: I'll add tests to the exposed method if possible. I won't add tests to an internal method.

PR #1242, opened 2025-09-03
Support browser 'open in new tab' navigation for runs and flows

ravikiranvm:
Fixes OPS-2501.

### DIFF
--- a/packages/react-ui/src/app/routes/flows/index.tsx
+++ b/packages/react-ui/src/app/routes/flows/index.tsx
@@ -17,7 +17,6 @@ import { flowsApi } from '@/app/features/flows/lib/flows-api';
 import { FlowsFolderBreadcrumbsManager } from '@/app/features/folders/component/folder-breadcrumbs-manager';
 import { FLOWS_TABLE_FILTERS } from '@/app/features/folders/lib/flows-table-filters';
 import { FlowStatus, FlowVersionState } from '@openops/shared';
-import is from 'zod/v4/locales/is.cjs';
 
 const FlowsPage = () => {
   useDefaultSidebarState('expanded');
--- a/packages/react-ui/src/app/common/navigation/table-navigation-helper.ts
+++ b/packages/react-ui/src/app/common/navigation/table-navigation-helper.ts
@@ -1,4 +1,4 @@
-export const MiddleMouseButtonClickEvent = 1;
+const MiddleMouseButtonClickEvent = 1;
 
 export function isModifierOrMiddleClick(
   e: Pick<React.MouseEvent, 'ctrlKey' | 'metaKey' | 'button'>,


packages/ui-components/src/ui/data-table.tsx
```diff
@@ -347,24 +349,40 @@ export function DataTable<
...
+                    key={row.id}
+                    className={onRowClick ? 'cursor-pointer' : ''}
+                    data-state={row.getIsSelected() && 'selected'}
+                  >
+                    {row.getVisibleCells().map((cell) => (
+                      <TableCell key={cell.id}>
+                        {rowHref && cell.column.id !== 'actions' ? (
+                          <Link
+                            to={rowHref}
+                            target="_self"
```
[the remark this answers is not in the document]
ravikiranvm: Done

packages/react-ui/src/app/routes/flows/index.tsx
```diff
@@ -70,14 +70,19 @@ const FlowsPage = () => {
...
+            }
             onRowClick={(row, e) => {
               const route = `/flows/${row.id}?${qs.stringify({
                 folderId: searchParams.get(FOLDER_ID_PARAM_NAME),
                 viewOnly: true,
               })}`;
-
-              if (e.ctrlKey) {
-                window.open(route, '_blank');
+              if (e.ctrlKey || e.metaKey || e.button === 1) {
```
ravikiranvm: Modified this to respect browser's native navigation and thus avoid double navigation

packages/ui-components/src/ui/data-table.tsx
```diff
@@ -347,24 +349,36 @@ export function DataTable<
...
+                  <TableRow
+                    data-testid="data-table-row"
+                    onClick={(e) => onRowClick?.(row.original, e)}
+                    key={row.id}
+                    className={onRowClick ? 'cursor-pointer' : ''}
+                    data-state={row.getIsSelected() && 'selected'}
+                  >
+                    {row.getVisibleCells().map((cell) => (
+                      <TableCell key={cell.id}>
+                        {rowHref && cell.column.id !== 'actions' ? (
```
ravikiranvm, line 364: This condition helps render a table cell with `<a>` tag only for those tables with getRowHref prop, so that this change wouldn't affect other data-table components like Connections table.

packages/react-ui/src/app/features/home/flows-table/flows-table.tsx
```diff
@@ -53,9 +53,10 @@ const HomeFlowsTable = ({
           loading={loading}
           stickyHeader
           border={false}
+          getRowHref={(row) => `/flows/${row.id}`}
           onRowClick={(row, e) => {
-            if (e.ctrlKey) {
-              window.open(`/flows/${row.id}`, '_blank');
+            if (e.ctrlKey || e.metaKey || e.button === 1) {
```
alexandrudanpop: can extract this common logic to a helpers file (for example: `table-navigation-helper.ts`, also magic number can be extracted as constant, like:
```js
const MiddleMoustButtonClickEvent = 1;
```

ravikiranvm: Done

packages/react-ui/src/app/routes/flows/index.tsx
ravikiranvm: Later added a helper function for this logic

PR #1246, opened 2025-09-04
Block editing calls for internal flows

bigfluffycookie:
Fixes OPS-2535

### DIFF
--- a/packages/server/api/src/app/flows/flow/flow.controller.ts
+++ b/packages/server/api/src/app/flows/flow/flow.controller.ts
@@ -81,6 +81,7 @@ export const flowController: FastifyPluginAsyncTypebox = async (app) => {
       id: request.params.id,
       projectId: request.principal.projectId,
     });
+    await assertThatFlowIsNotInternal(flow);
     await assertThatFlowIsNotBeingUsed(flow, userId);
 
     const updatedFlow = await flowService.update({
@@ -158,6 +159,8 @@ export const flowController: FastifyPluginAsyncTypebox = async (app) => {
         id: request.params.id,
         projectId: request.principal.projectId,
       });
+
+      await assertThatFlowIsNotInternal(flow);
 
       if (!flow.publishedVersionId) {
         return await reply.status(StatusCodes.BAD_REQUEST).send({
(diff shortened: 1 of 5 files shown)


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -352,13 +352,6 @@ export const flowService = {
   }: UpdatePublishedVersionIdParams): Promise<PopulatedFlow> {
     const flowToUpdate = await this.getOneOrThrow({ id, projectId });
 
-    if (flowToUpdate.isInternal) {
```
bigfluffycookie, line 355: removed bc it gets blocked at the controller level now

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -409,7 +402,7 @@ export const flowService = {
           flowId: id,
         });
         throw new ApplicationError({
-          code: ErrorCode.FLOW_OPERATION_INVALID,
+          code: ErrorCode.FLOW_INTERNAL_FORBIDDEN,
```
bigfluffycookie: this still needs to be in here, because it is a hard deletion. at some point we will change this to be a soft delete but until then im blocking it here.

packages/server/api/src/app/flows/flow/flow.controller.ts
```diff
@@ -258,7 +261,7 @@ async function createFromTemplate(
   return updatedFlow;
 }
 
-async function assertThatFlowIsNotBeingUsed(
+export async function assertThatFlowIsNotBeingUsed(
```
bigfluffycookie, line 264: needed so we can reuse it in private templates

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -683,6 +670,21 @@ const getConnections = async (
   });
 };
 
+export async function assertThatFlowIsNotInternal(flow: Flow): Promise<void> {
+  if (flow.isInternal) {
+    const message =
+      'Flow is internal, cannot be manipulated through the flow API.';
+    logger.warn(message, {
```
bigfluffycookie: adding a logger for it because this should not happen, if we get here it means a user used the API with a flow id manually

packages/server/api/src/app/flows/flow/flow.controller.ts
bigfluffycookie: needed so we can reuse it in private templates

MarceloRGonc: If it's used on two controllers, move it to another file. In this case, does not make sense for one controller to import another.

packages/server/api/src/app/flows/flow/flow.service.ts
MarceloRGonc: Maybe create a file with the two assertions

PR #1251, opened 2025-09-05
Cezudas/ops 2469

cezudas:
Part of OPS-2469.

### DIFF
--- a/packages/server/api/tsconfig.app.json
+++ b/packages/server/api/tsconfig.app.json
@@ -3,7 +3,16 @@
   "compilerOptions": {
     "outDir": "../../../dist/out-tsc",
     "module": "CommonJS",
-    "types": ["node"]
+    "types": ["node"],
+    "paths": {
+      "@/mcp/tool-approval-wrapper": [
+        "packages/server/api/src/app/ai/mcp/tool-approval-wrapper"
+      ],
+      "server-worker": ["packages/server/worker/src/index"],
+      "@openops/server-shared": ["packages/server/shared/src/index"],
+      "@openops/shared": ["packages/shared/src/index"],
+      "@openops/common": ["packages/openops/src/index"]
+    }
   },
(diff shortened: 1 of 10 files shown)


packages/shared/src/lib/ai/tool-approval/index.ts
```diff
@@ -0,0 +1,5 @@
+export type ToolStepContext = {
```
cezudas, line 1: will be needed for ui-kit

packages/server/api/tsconfig.app.json
```diff
@@ -3,7 +3,16 @@
   "compilerOptions": {
     "outDir": "../../../dist/out-tsc",
     "module": "CommonJS",
-    "types": ["node"]
+    "types": ["node"],
+    "paths": {
+      "@/mcp/tool-approval-wrapper": [
+        "packages/server/api/src/app/ai/mcp/tool-approval-wrapper"
+      ],
+      "server-worker": ["packages/server/worker/src/index"],
```
cezudas: server-worker and the other "@openops/" paths below are already mapped in the root tsconfig.base.json, so paths object in packages/server/api/tsconfig.app.json completely overrides the parent's paths instead of merging them.

packages/server/api/tsconfig.json
```diff
@@ -16,6 +16,18 @@
...
+        "packages/server/api/src/app/ai/mcp/tool-approval-wrapper"
+      ],
+      "server-worker": ["packages/server/worker/src/index"],
+      "@openops/block-slack": ["packages/blocks/slack/src/index"],
+      "@openops/blocks-common": ["packages/blocks/common/src/index"],
+      "@openops/blocks-framework": ["packages/blocks/framework/src/index"],
+      "@openops/common": ["packages/openops/src/index"],
+      "@openops/server-shared": ["packages/server/shared/src/index"],
+      "@openops/shared": ["packages/shared/src/index"]
+    }
```
MarceloRGonc, line 31: Just a note for the future. We don't intend to stay like this for long.

PR #1255, opened 2025-09-05
Add support for modify tfvars files

MarceloRGonc:
Fixes OPS-2504.

image

image


packages/blocks/terraform/src/lib/modify/variable-updates-property.ts
```diff
@@ -0,0 +1,142 @@
...
+        await getVariables(fileParsed.content);
+
+      return {
+        updates: getUpdatesProperty(terraformVariables),
+      };
+    },
+  });
+}
+
+function getErrorMessageProperty(message?: string) {
```
bigfluffycookie, line 38: whats this for?

bigfluffycookie (approved):

packages/blocks/terraform/src/lib/modify/variable-updates-property.ts
MarceloRGonc: If the file is invalid, I want to show an error. I'm adding a dropdown with the error. Because we don't have a better way to do this.

PR #1258, opened 2025-09-08
Add new column to folder table, also add validation

bigfluffycookie:
Fixes OPS-2551


packages/server/api/src/app/flows/folder/uncategorized-flows.ts
```diff
@@ -0,0 +1,15 @@
+import { ApplicationError, ContentType, ErrorCode } from '@openops/shared';
+import { flowService } from '../flow/flow.service';
+
+export const getUncategorizedFlows = async (
```
bigfluffycookie, line 10: alright so this is a bit of an eyesore, im gonna explain the issue and why ive chosen this approach. For folders, we have the concept of uncategorized, which gets all flows that do not have a folderID, now for the templates, we can reuse everything inside the folder service except this, we need to somehow be able to getUncategorizedTemplates, my thinking was to then add this in here later, in another "if" based on the content type. In terms of merge conflicts i think this will be minimal since its in a seperate file here. if you have anothe approach in mind please let me know.

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -477,6 +484,32 @@ export const flowService = {
       isInternal: false,
     });
   },
+
+  async getUncategorizedWorkflows(projectId: string) {
```
bigfluffycookie: moved this over from folder.service, explanation below (anyways this is more flow related than folder related so i think its fine to move)

packages/server/api/src/app/flows/folder/folder.service.ts
```diff
@@ -212,6 +203,7 @@ export const flowFolderService = {
       })),
       subfolders: [],
       parentFolderId: undefined,
+      contentType: ContentType.WORKFLOW,
```
MarceloRGonc:
```suggestion
      contentType,
```

I think you want this.

packages/server/api/src/app/flows/folder/uncategorized-flows.ts
bigfluffycookie: agreed over slack to proceed with this

packages/server/api/src/app/database/migrations/1757331587268-AddContentTypeToFolder.ts
```diff
@@ -0,0 +1,39 @@
...
+      SET "contentType" = 'WORKFLOW'
+      WHERE "contentType" IS NULL;
+    `);
+
+    await queryRunner.query(`
+      ALTER TABLE "folder"
+      ALTER COLUMN "contentType" SET NOT NULL;
+    `);
+
+    await queryRunner.query(`
```
bigfluffycookie, line 24: adding a new index, you should be able to have the same folder name for different contenttypes

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -477,18 +484,98 @@ export const flowService = {
       isInternal: false,
     });
   },
+
+  async getWorkflowsByFolder({
```
bigfluffycookie: moved this from folder.service,  same as with the uncategorized query.

packages/server/api/src/app/flows/folder/folder.module.ts
```diff
@@ -51,6 +51,7 @@ const folderController: FastifyPluginAsyncTypebox = async (fastify) => {
       includeUncategorizedFolder: Boolean(
         !request.query.excludeUncategorizedFolder,
       ),
+      contentType: request.query.contentType,
```
bigfluffycookie, line 54: the reason why the only update is here is because in the other post operations the contentType is within the request body, so no change is visible here

packages/server/api/src/app/flows/folder/folder.service.ts
```diff
@@ -128,76 +143,57 @@ export const flowFolderService = {
...
+    const {
+      projectId,
+      includeUncategorizedFolder,
+      contentType = ContentType.WORKFLOW,
+    } = params;
     const query = folderRepo()
       .createQueryBuilder('folder')
       .loadRelationCountAndMap('folder.numberOfFlows', 'folder.flows')
       .leftJoinAndSelect('folder.parentFolder', 'parentFolder')
-      .leftJoinAndSelect(
```
bigfluffycookie, line 136: moved to flow.service

packages/server/api/src/app/flows/folder/folder.service.ts
```diff
@@ -128,76 +143,57 @@ export const flowFolderService = {
...
     );
   },
   async getUncategorizedFolder({
     projectId,
+    contentType = ContentType.WORKFLOW,
   }: {
     projectId: string;
+    contentType?: ContentType;
   }): Promise<FolderDto> {
-    const uncategorizedFlowsQuery = flowRepo()
```
bigfluffycookie, line 179: moved to uncategorized-flows.ts

packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -543,6 +631,16 @@ async function update({
...
     } else if (operation.type === FlowOperationType.CHANGE_FOLDER) {
+      if (operation.request.folderId) {
+        const folder = await flowFolderService.getOneOrThrow({
+          projectId,
+          folderId: operation.request.folderId,
+        });
+        await assertThatFlowIsInCorrectFolderContentType(
+          contentType,
+          folder.contentType,
+        );
```
MarceloRGonc: Extract to a method

packages/server/api/src/app/flows/folder/folder.entity.ts
```diff
@@ -20,11 +20,17 @@ export const FolderEntity = new EntitySchema<FolderSchema>({
       type: String,
     },
     projectId: OpenOpsIdSchema,
+    contentType: {
+      type: String,
+      enum: ContentType,
+      nullable: false,
+      default: ContentType.WORKFLOW,
```
MarceloRGonc: To do this you can put the default in the migration

packages/server/api/src/app/flows/folder/folder.service.ts
MarceloRGonc: Your change is worse in terms of performance. You changed from 2 queries to 2 plus one query per folder.

bigfluffycookie: Yea, i know :/ should i just copy the whole original query instead and just live with the duplication?

packages/server/api/src/app/flows/folder/folder.entity.ts
bigfluffycookie: leftover, dont want a default so removed it

packages/server/api/src/app/flows/folder/folder-flows.ts
```diff
@@ -0,0 +1,59 @@
+import { ApplicationError, ContentType, ErrorCode } from '@openops/shared';
```
bigfluffycookie, line 1: okay im pretty happy with the change now, it minimizes duplication and we can reuse getFilterCondition also for uncategorized (the only thing im not happy with the is name getFilterCondition)

packages/server/api/src/app/flows/folder/folder.service.ts
bigfluffycookie: changed it around quite a bit, its not in folder-flows directly

MarceloRGonc (commented):
Rename the method as we talked. And keep an eye on the Sonar errors

PR #1265, opened 2025-09-09
Change some logs to not trigger false production alerts

MarceloRGonc:
Fixes OPS-2564

### DIFF
--- a/packages/engine/src/engine-executor.ts
+++ b/packages/engine/src/engine-executor.ts
@@ -27,13 +27,10 @@ export async function executeEngine(
 
   const key = await saveRequestBody(requestId, result);
 
-  logger.info(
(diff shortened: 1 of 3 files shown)


packages/engine/src/engine-executor.ts
```diff
@@ -27,13 +27,10 @@ export async function executeEngine(
...
-    `Finished operation [${operationType}] with status [${result.status}] in ${duration}ms`,
-    {
-      engineStatus: result.status,
-      durationMs: duration,
-    },
-  );
+  logger.info(`Finished operation [${operationType}] in ${duration}ms`, {
+    engineStatus: result.status,
+    durationMs: duration,
+  });
```
MarceloRGonc, line 33: The status may be Error if the Engine has been stopped. Does not change alerts for execution failures.

PR #1270, opened 2025-09-10
Remove retry from failed step feature

MarceloRGonc:
Fixes OPS-2498.


packages/shared/src/lib/flow-run/flow-run.ts
```diff
@@ -13,7 +13,6 @@ export enum RunEnvironment {
 
 export enum FlowRetryStrategy {
   ON_LATEST_VERSION = 'ON_LATEST_VERSION',
-  FROM_FAILED_STEP = 'FROM_FAILED_STEP',
 }
```
rmn-snpk, line 17: Do we still need the enum here

bigfluffycookie (approved):

PR #1276, opened 2025-09-10
Save trigger status before execution

MarceloRGonc:
Fixes OPS-2500.

### DIFF
--- a/packages/server/worker/src/lib/executors/flow-job-executor.ts
+++ b/packages/server/worker/src/lib/executors/flow-job-executor.ts
@@ -19,6 +19,7 @@ import {
   isNil,
   ResumeExecuteFlowOperation,
   ResumePayload,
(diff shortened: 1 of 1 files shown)


packages/server/worker/src/lib/executors/flow-job-executor.ts
```diff
@@ -163,10 +169,19 @@ async function executeFlow(
...
   await engineApiService(engineToken).updateRunStatus({
     runDetails: {
-      steps: {},
+      steps: {
+        [triggerName]: {
+          status: StepOutputStatus.SUCCEEDED,
+          output: jobData.payload,
+          type: triggerType,
+          input: {},
+        },
```
MarceloRGonc, line 183: If the run fail we have the data to re-run. It's the same we do in the engine

PR #1281, opened 2025-09-11
Global app layout for AI chat

alexandrudanpop:
Fixes OPS-2502

### DIFF
--- a/packages/react-ui/src/app/features/builder/index.tsx
+++ b/packages/react-ui/src/app/features/builder/index.tsx
@@ -189,11 +189,14 @@ const BuilderPage = ({ children }: { children?: ReactNode }) => {
       memorizedSelectedStep?.type === TriggerType.BLOCK,
   });
 
-  const isRightSidebarVisible =
-    rightSidebar === RightSideBarType.BLOCK_SETTINGS &&
-    !!memorizedSelectedStep &&
-    memorizedSelectedStep.type !== TriggerType.EMPTY &&
-    !isBlockLoading;
+  const isRightSidebarVisible = useMemo(() => {
+    return (
+      rightSidebar === RightSideBarType.BLOCK_SETTINGS &&
+      !!memorizedSelectedStep &&
+      memorizedSelectedStep.type !== TriggerType.EMPTY &&
+      !isBlockLoading
+    );
+  }, [rightSidebar, memorizedSelectedStep, isBlockLoading]);
 
   const rightHandleRef = useRef<ImperativePanelHandle>(null);
   const leftHandleRef = useAnimateSidebar(


packages/react-ui/src/app/common/guards/allow-logged-in-user-only-guard.tsx
```diff
@@ -45,6 +45,15 @@ export const AllowOnlyLoggedInUserOnlyGuard = ({
...
   const expired = !token || isJwtExpired(token);
 
+  projectHooks.prefetchProject();
+  platformHooks.prefetchPlatform();
+  platformHooks.prefetchNewerVersionInfo(queryClient);
+
+  flagsHooks.useFlags();
+  userSettingsHooks.useUserSettings();
+  userHooks.useUserMeta();
+  appConnectionsHooks.useConnectionsMetadata();
```
alexandrudanpop, line 55: hooks were rendered conditionally resulting in a runtime error when token is expired

cezudas: cool 👍 is it related to this ticket/PR or can we do it in a standalone ticket?

packages/react-ui/src/app/features/ai/assistant/ai-chat-resizable-panel.tsx
```diff
@@ -39,60 +23,59 @@ const AiChatResizablePanel = ({ onDragging }: AiChatResizablePanelProps) => {
...
+          />
+        </div>
       </ResizablePanel>
-      <ResizableHandle
-        className={cn('w-0')}
-        onDragging={onDragging}
-        disabled={!showChat}
-      />
+      {showChat && (
+        <ResizableHandle className={cn('w-0')} onDragging={onDragging} />
```
rmn-snpk: [nitpick]
 
```suggestion
        <ResizableHandle className="w-0" onDragging={onDragging} />
```

packages/react-ui/src/app/features/builder/index.tsx
```diff
@@ -189,6 +189,18 @@ const BuilderPage = ({ children }: { children?: ReactNode }) => {
       memorizedSelectedStep?.type === TriggerType.BLOCK,
   });
 
+  const isRightSidebarVisible =
+    rightSidebar === RightSideBarType.BLOCK_SETTINGS &&
+    !!memorizedSelectedStep &&
+    memorizedSelectedStep.type !== TriggerType.EMPTY &&
+    !isBlockLoading;
```
rmn-snpk: Can we memorize the value as it's used as dependency for useEffect

alexandrudanpop: fixed

PR #1291, opened 2025-09-12
Update private templates handling

rmn-snpk:
Fixes OPS-2604.

image


packages/react-ui/src/app/features/templates/components/select-flow-template-dialog-content.tsx
```diff
@@ -174,7 +173,7 @@ const SelectFlowTemplateDialogContent = ({
         ) : (
           <FlowTemplateGallery
             selectionHeading={selectionHeading}
-            showPrivateTemplates={EDITION !== OpsEdition.COMMUNITY}
+            showPrivateTemplates={false}
```
rmn-snpk, line 176: Discussed with Maor, we don't want to depend on that flag

PR #1307, opened 2025-09-16
Fix Stop execution block to not change the verdict to SUCCEEDED

MarceloRGonc:
Fixes OPS-2632.

## Additional notes:
- The Stop execution step was marking the run as finished even though it was included inside a loop.
- Now we are marking the execution as finished only when we are returning from the engine to the server.

### DIFF
--- a/packages/engine/src/lib/handler/block-executor.ts
+++ b/packages/engine/src/lib/handler/block-executor.ts
@@ -203,7 +203,7 @@ const executeAction: ActionHandler<BlockAction> = async ({
           action.name,
           stepOutput.setOutput(output).setStatus(StepOutputStatus.STOPPED),
         )
-        .setVerdict(ExecutionVerdict.SUCCEEDED, {
+        .setVerdict(ExecutionVerdict.RUNNING, {
           reason: VerdictReason.STOPPED,
           stopResponse: hookResponse.stopResponse.response,
         })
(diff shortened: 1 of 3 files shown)


packages/engine/src/lib/handler/block-executor.ts
```diff
@@ -203,7 +203,7 @@ const executeAction: ActionHandler<BlockAction> = async ({
           action.name,
           stepOutput.setOutput(output).setStatus(StepOutputStatus.STOPPED),
         )
-        .setVerdict(ExecutionVerdict.SUCCEEDED, {
+        .setVerdict(ExecutionVerdict.RUNNING, {
```
MarceloRGonc, line 206: We shouldn't do this here. If we're inside a loop, the workflow will continue.

packages/engine/src/lib/handler/flow-executor.ts
```diff
@@ -107,7 +108,10 @@ export const flowExecutor = {
 
       await sendProgress(flowExecutionContext, constants);
 
-      if (flowExecutionContext.verdict !== ExecutionVerdict.RUNNING) {
+      if (
+        flowExecutionContext.verdict !== ExecutionVerdict.RUNNING ||
+        flowExecutionContext.verdictResponse?.reason === VerdictReason.STOPPED
```
MarceloRGonc, line 113: If the verdict response is Stopped we do not want to perform the following steps.

bigfluffycookie (approved):

PR #1308, opened 2025-09-16
Disable workflow retries

MarceloRGonc:
Fixes OPS-2633.

## Additional notes
- The locking time could be shorter than the workflow execution time.
- After the locking time had passed, the workflow would be retried.
- The retry overrides the previous execution.

### DIFF
--- a/packages/server/api/src/app/workers/redis/redis-consumer.ts
+++ b/packages/server/api/src/app/workers/redis/redis-consumer.ts
@@ -1,6 +1,7 @@
 import {
   createRedisClient,
   exceptionHandler,
+  flowTimeoutSandbox,
   JobStatus,
   logger,
   memoryLock,
@@ -97,7 +98,7 @@ async function ensureWorkerExists(queueName: QueueName): Promise<Worker> {
   consumer[queueName] = new Worker(queueName, null, {
     connection: createRedisClient(),
     lockDuration,
-    maxStalledCount: 5,
+    maxStalledCount: 0,
     drainDelay: 5,
     stalledInterval: 30000,
   });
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/workers/redis/redis-consumer.ts
```diff
@@ -112,7 +113,7 @@ function getLockDurationInMs(queueName: QueueName): number {
     case QueueName.WEBHOOK:
       return (60 + 60 * 3) * 1000;
     case QueueName.ONE_TIME:
-      return (600 + 60 * 3) * 1000;
+      return (flowTimeoutSandbox + 60 * 3) * 1000;
```
MarceloRGonc, line 116: Must be the same time defined for the execution of the workflow

packages/server/api/src/app/workers/redis/redis-consumer.ts
```diff
@@ -97,7 +98,7 @@ async function ensureWorkerExists(queueName: QueueName): Promise<Worker> {
   consumer[queueName] = new Worker(queueName, null, {
     connection: createRedisClient(),
     lockDuration,
-    maxStalledCount: 5,
+    maxStalledCount: 0,
```
MarceloRGonc, line 101: Let's disable retries for now. They should only happen if the workflow hasn't even started and we have no way to control that.

bigfluffycookie: the ticket is about a bug that a workflow was being retried, this pr stops this however do we know why it was being retried?

MarceloRGonc: The workflows timeout. But even if it succeeded if it's after the lock time the workflow would retry.

bigfluffycookie (approved):

PR #1309, opened 2025-09-16
Do not update logs file if the run is in final state

MarceloRGonc:
Part of OPS-2632.

## Additional notes:
- Even though the run was in a final state, the steps executed were updated
- We are updating only if the run is not in a final state

### DIFF
--- a/packages/engine/src/lib/handler/block-executor.ts
+++ b/packages/engine/src/lib/handler/block-executor.ts
@@ -203,7 +203,7 @@ const executeAction: ActionHandler<BlockAction> = async ({
           action.name,
           stepOutput.setOutput(output).setStatus(StepOutputStatus.STOPPED),
         )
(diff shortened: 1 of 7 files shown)


packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -294,13 +294,26 @@ export const flowRunService = {
     tags,
     duration,
   }: FinishParams): Promise<FlowRun | undefined> {
+    let flowRun = await flowRunRepo().findOneByOrFail({ id: flowRunId });
+    if (TERMINAL_STATUSES.includes(flowRun.status)) {
```
rita-gorokhod: use the helper method
please log an error

rita-gorokhod (approved):

PR #1319, opened 2025-09-17
Add duration and status to the step execution log

MarceloRGonc:
Part of CI-71

### DIFF
--- a/packages/engine/src/lib/handler/block-executor.ts
+++ b/packages/engine/src/lib/handler/block-executor.ts
@@ -76,10 +76,12 @@ export const blockExecutor: BaseExecutor<BlockAction> = {
     } finally {
       const duration = Math.floor(performance.now() - startTime);
       logger.info(
-        `The execution of block ${action.settings.blockName} took ${duration}ms`,
+        `Executed step [${action.name}] action [${action.settings.actionName}] in ${duration}ms`,
         {
           stepStatus,
+          stepId: action.id,
           durationMs: duration,
+          stepName: action.name,
           blockName: action.settings.blockName,
           actionName: action.settings.actionName,
           continueOnFailure:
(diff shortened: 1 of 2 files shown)


packages/engine/src/lib/handler/block-executor.ts
```diff
@@ -57,7 +57,7 @@ export const blockExecutor: BaseExecutor<BlockAction> = {
     constants: EngineConstants;
   }) {
     const startTime = performance.now();
-    let status: string | undefined;
+    let outputStatus: string | undefined;
```
maor-rozenfeld: How about stepStatus?

packages/engine/src/lib/handler/block-executor.ts
```diff
@@ -71,14 +71,15 @@ export const blockExecutor: BaseExecutor<BlockAction> = {
...
+      outputStatus = resultExecution.getStepOutput(action.name)?.status;
       return await continueIfFailureHandler(resultExecution, action, constants);
     } finally {
       const duration = Math.floor(performance.now() - startTime);
       logger.info(
         `The execution of block ${action.settings.blockName} took ${duration}ms`,
         {
-          status,
+          outputStatus,
+          durationInMS: duration,
```
maor-rozenfeld: We already have `durationMs` field elsewhere, so let's keep that "convention"

packages/engine/src/lib/handler/block-executor.ts
```diff
@@ -71,14 +71,15 @@ export const blockExecutor: BaseExecutor<BlockAction> = {
         executeAction,
       );
 
-      status = resultExecution.getStepOutput(action.name)?.status;
+      stepStatus = resultExecution.getStepOutput(action.name)?.status;
       return await continueIfFailureHandler(resultExecution, action, constants);
     } finally {
       const duration = Math.floor(performance.now() - startTime);
       logger.info(
         `The execution of block ${action.settings.blockName} took ${duration}ms`,
```
maor-rozenfeld: Is there a step ID we can use here?
```
`Executed step [${action.stepId}] using [${action.settings.blockName}] in ${duration}ms`
```

MarceloRGonc: image

maor-rozenfeld (approved):

PR #1320, opened 2025-09-17
Fix race condition in webhook simulation cleanup

cezudas:
Fixes OPS-2600.

The race condition manifests in this sequence:
- A published flow version exists with webhook simulations
- During import, the system creates a new empty draft version
- The IMPORT_FLOW operation triggers applySingleOperation
- This calls flowVersionSideEffects.preApplyOperation
- preApplyOperation calls deleteWebhookSimulation at the wrong time - when the new flow version is a draft with empty triggers
- This causes the webhook simulation deletion to fail and log an ApplicationError with ErrorCode.TRIGGER_DISABLE

### DIFF
--- a/packages/server/api/src/app/flows/flow/flow.service.ts
+++ b/packages/server/api/src/app/flows/flow/flow.service.ts
@@ -37,6 +37,7 @@ import {
   sendWorkflowExportedEvent,
   sendWorkflowUpdatedEvent,
 } from '../../telemetry/event-models';
+import { webhookSimulationService } from '../../webhooks/webhook-simulation/webhook-simulation-service';
 import {
   flowVersionRepo,
   flowVersionService,
@@ -610,6 +611,11 @@ async function update({
       });
 
       if (lastVersion.state === FlowVersionState.LOCKED) {
+        await webhookSimulationService.delete({
+          flowId: id,
+          projectId,
+        });
+
         const lastVersionWithArtifacts =
           await flowVersionService.getFlowVersionOrThrow({
             flowId: id,


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -610,6 +611,11 @@ async function update({
       });
 
       if (lastVersion.state === FlowVersionState.LOCKED) {
+        await webhookSimulationService.delete({
+          flowId: id,
+          projectId,
+        });
+
```
cezudas, line 618: The explicit webhook cleanup fixes a timing problem— the execution flow reaches `webhookSimulationService.delete` and calls `webhookSideEffects.preDelete`  [(see L127)](https://github.com/openops-cloud/openops/blob/cezudas/OPS-2600/packages/server/api/src/app/webhooks/webhook-simulation/webhook-simulation-service.ts#L127) with the flowVersionId parameter matching the original published version.

The `IMPORT_FLOW` operation still triggers a cleanup call; however, the execution flow inside `webhookSimulationService.delete` [returns early (L124-L126) ](https://github.com/openops-cloud/openops/blob/cezudas/OPS-2600/packages/server/api/src/app/webhooks/webhook-simulation/webhook-simulation-service.ts#L124-L126). That is because the webhook simulation no longer exists for the flow ID. So a second call to ` webhookSideEffects.preDelete` is avoided and no error is raised.

cezudas:
> ## Pull Request Overview
> Fixes a race condition in webhook simulation cleanup during flow import operations. The issue occurred when webhook simulations were deleted at the wrong time during the import process, causing failures when trying to delete simulations from empty draft flow versions.
> 
> * Moved webhook simulation deletion to occur before creating new draft flow versions
> * Added explicit webhook simulation cleanup in the flow update operation
> 
> **Tip:** Customize your code reviews with copilot-instructions.md. [Create the file](/openops-cloud/openops/new/main/.github?filename=copilot-instructions.md) or [learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).



> ## Pull Request Overview
> Fixes a race condition in webhook simulation cleanup during flow import operations. The issue occurred when webhook simulations were deleted at the wrong time during the import process, causing failures when trying to delete simulations from empty draft flow versions.
> 
> * Moved webhook simulation deletion to occur before creating new draft flow versions
> * Added explicit webhook simulation cleanup in the flow update operation
> 
> **Tip:** Customize your code reviews with copilot-instructions.md. [Create the file](/openops-cloud/openops/new/main/.github?filename=copilot-instructions.md) or [learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

The overview is not 100% accurate. The race condition happens in this sequence:
-Published workflow with Webhook
-Click test/retest to trigger a webhook simulation
-From the canvas, make an update operation to the workflow, such as deleting a step

PR #1323, opened 2025-09-18
Disable slack URL previews in message action

maor-rozenfeld:
Fixes OPS-2658.

### DIFF
--- a/packages/blocks/slack/src/lib/common/utils.ts
+++ b/packages/blocks/slack/src/lib/common/utils.ts
@@ -398,9 +398,9 @@ export function createMessageBlocks(
       text: {
         type: 'mrkdwn',
         text: `${messageText}`,
-      },
-      unfurl_links: false,
-      unfurl_media: false,
+        unfurl_links: false,
+        unfurl_media: false,
+      }
     },
   ];
 }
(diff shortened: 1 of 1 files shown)


packages/blocks/slack/src/lib/common/utils.ts
```diff
@@ -51,6 +51,8 @@ export const slackSendMessage = async ({
         event_type: 'slack-message',
         event_payload: eventPayload,
       },
+      unfurl_links: false,
```
rita-gorokhod, line 54: @maor-rozenfeld this should be configurable, we can't change this in our product for all users just because you don't like it.
You can add a new property.

rita-gorokhod (changes requested):
Add 2 non mandatory configurable properties, with default true as before.

packages/blocks/slack/src/lib/common/utils.ts
maor-rozenfeld: I don't think any user actually wants these previews, they're rather spammy. Let's discuss.

rita-gorokhod: I'm a user and I do want them, so let's go with 2 non mandatory configurable properties, with default true as before.

PR #1327, opened 2025-09-18
Add IMAP integration

rSnapkoOpenOps:
Fixes OPS-2640.


packages/blocks/imap/src/lib/common/build-search.ts
```diff
@@ -0,0 +1,71 @@
...
+  const recipientList = normalizeList(recipients);
+  const ccList = normalizeList(cc);
+  const senderList = normalizeList(senders);
+
+  const search: SearchObject = { since: sinceISO };
+
+  let clauses: SearchObject[] = [{}];
+  clauses = expandClauses(clauses, 'to', recipientList);
+  clauses = expandClauses(clauses, 'cc', ccList);
+  clauses = expandClauses(clauses, 'from', senderList);
```
rmn-snpk, line 52: We need to build or combinations for each filter, as their api supports only one `or` field

packages/blocks/imap/src/lib/common/build-client.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function buildClient(auth: BlockPropValueSchema<typeof imapAuth>) {
+  return new ImapFlow({
+    host: auth.host,
+    port: auth.port,
+    secure: auth.tls,
+    auth: { user: auth.username, pass: auth.password },
+    tls: { rejectUnauthorized: false },
+    logger: false,
+  });
```
rmn-snpk, line 13: Builds ImapFlow object which is used as IMAP client

packages/blocks/imap/src/lib/common/new-email-polling.ts
```diff
@@ -0,0 +1,32 @@
...
+  items: async ({ auth, lastFetchEpochMS, propsValue }) => {
+    const items = await fetchEmails({
+      auth,
+      lastEpochMilliSeconds: lastFetchEpochMS,
+      mailbox: propsValue.mailbox,
+      recipients: propsValue.recipients,
+      cc: propsValue.cc,
+      senders: propsValue.senders,
+      subject: propsValue.subject,
+    });
```
MarceloRGonc, line 28: What happens if an error is thrown here? Does BE support it?

packages/blocks/imap/src/lib/common/mailbox.ts
```diff
@@ -0,0 +1,32 @@
...
+import { buildClient } from './build-client';
+import { imapAuth } from './imap-auth';
+
+export const mailbox = Property.Dropdown({
+  displayName: 'Mailbox',
+  description: 'Select the mailbox to search',
+  required: true,
+  refreshers: ['auth'],
+  options: async ({ auth }) => {
+    const authValue = auth as BlockPropValueSchema<typeof imapAuth>;
```
MarceloRGonc, line 19: Check for no auth. Or invalid and add the proper error to the dropdown

rmn-snpk: Done

packages/blocks/imap/src/lib/common/new-email-polling.ts
rmn-snpk: The error will be catch on the executor level 
image

bigfluffycookie (approved):

PR #1330, opened 2025-09-18
Cancel the back-end stream when the user clicks on chat cancel button

cezudas:
Fixes OPS-2645.

### DIFF
--- a/packages/server/api/src/app/ai/chat/chat-request-router.ts
+++ b/packages/server/api/src/app/ai/chat/chat-request-router.ts
@@ -2,7 +2,7 @@ import { logger } from '@openops/server-shared';
 import { CODE_BLOCK_NAME, NewMessageRequest, Principal } from '@openops/shared';
 import { ModelMessage } from 'ai';
 import { FastifyInstance, FastifyReply } from 'fastify';
-import { ServerResponse } from 'node:http';
+import { IncomingMessage, ServerResponse } from 'node:http';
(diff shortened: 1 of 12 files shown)


packages/server/api/src/app/ai/chat/chat-request-router.ts
```diff
@@ -70,6 +79,7 @@ export async function routeChatRequest(
     'Cache-Control': 'no-cache',
     Connection: 'keep-alive',
   });
+  serverResponse.write(': connection established\n\n');
```
cezudas, line 82: It wakes up the connection monitoring system, and lowers the time between user clicks "cancel" and backend calls controller.abort(). The frontend ignores this message since it starts with ": " prefix.

packages/server/api/src/app/ai/chat/llm-stream-handler.ts
```diff
@@ -4,13 +4,18 @@ import {
...
   stepCountIs,
+  StepResult,
   streamText,
   StreamTextOnFinishCallback,
   StreamTextOnStepFinishCallback,
   TextStreamPart,
   ToolSet,
 } from 'ai';
 
+type StreamTextOnAbortCallback<TOOLS extends ToolSet> = (event: {
```
cezudas, line 15: Unfortunately AI SDK has StreamTextOnAbortCallback type but chose not to export it

PR #1332, opened 2025-09-19
Wait for Satoshi fontpack to be fully loaded before each Chromatic test run

cezudas:
Fixes OPS-
Uses [Option C](https://www.chromatic.com/docs/font-loading/#solution-c-check-fonts-have-loaded-in-a-loader) from Chromatic docs.

### DIFF
--- a/packages/ui-components/.storybook/preview.tsx
+++ b/packages/ui-components/.storybook/preview.tsx
@@ -1,3 +1,4 @@
+import isChromatic from 'chromatic/isChromatic';
 import React, { Suspense } from 'react';
 import { I18nextProvider } from 'react-i18next';
 import '../src/styles/global.css';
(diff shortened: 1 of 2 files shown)


packages/ui-components/.storybook/preview.tsx
```diff
@@ -49,3 +50,26 @@ export const parameters = {
     diffThreshold: 0.2,
   },
 };
+
+const fontLoader = async () => {
+  const satoshiVariants = [
+    '300 16px Satoshi',
```
cezudas: we don't use font-thin nor font-extralight but i can add them as well. wdyt @rSnapkoOpenOps @alexandrudanpop ?

rmn-snpk: Yes, please add, so we don't touch it later

PR #1334, opened 2025-09-22
Move getMicrosoftGraphClient to common folder

bigfluffycookie:
Part of OPS-2681


packages/blocks/microsoft-teams/src/lib/actions/send-channel-message.ts
```diff
@@ -1,6 +1,6 @@
 import { createAction, Property } from '@openops/blocks-framework';
+import { getMicrosoftGraphClient } from '@openops/common';
```
bigfluffycookie, line 2: moved so we can reuse this for microsoft outlook

packages/blocks/microsoft-teams/test/auth-validation.test.ts
```diff
@@ -2,7 +2,8 @@ import { microsoftTeamsAuth } from '../src/lib/common/microsoft-teams-auth';
 
 const mockGet = jest.fn();
 
-jest.mock('../src/lib/common/get-microsoft-graph-client', () => ({
+jest.mock('@openops/common', () => ({
+  ...jest.requireActual('@openops/common'),
```
MarceloRGonc: Did you validate that you need this line? If it is not necessary it is better not to add it. Same for the other tests.

```
...jest.requireActual('@openops/common'),

bigfluffycookie: we usually always had issues longterm withouth adding this, not adding it can have unforeseen consequences in other tests that need to mock this package

MarceloRGonc: This is a mock, if there is no more code used from the package this line is not necessary. What is being done with this is using the package 'real' code and just mocking the 'getMicrosoftGraphClient' method. Please check if the tests work without this line.

bigfluffycookie: I understand what this code does it was a deliberate choice to add it because I personally prefer mocking only the parts needed over mocking the entire package. I think it is cleaner and more future proof.

MarceloRGonc: We should only mock what the test uses. It's cleaner to look at the test and know what's being used. Therefore, if the test only uses one method from the common package, you shouldn't load the entire package and then mock a method.

PR #1339, opened 2025-09-22
Add support for dynamic values to add tags action

MarceloRGonc:
Fixes OPS-2685.

### DIFF
--- a/packages/blocks/aws/src/lib/actions/add-tags-action.ts
+++ b/packages/blocks/aws/src/lib/actions/add-tags-action.ts
@@ -7,8 +7,9 @@ import {
   getARNsProperty,
   getCredentialsForAccount,
   groupARNsByAccount,
+  tryParseJson,
 } from '@openops/common';
-import { RiskLevel } from '@openops/shared';
+import { isObject, RiskLevel } from '@openops/shared';
 
 export const addTagsAction = createAction({
   auth: amazonAuth,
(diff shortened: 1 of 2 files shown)


packages/blocks/aws/src/lib/actions/add-tags-action.ts
```diff
@@ -61,12 +62,19 @@ export const addTagsAction = createAction({
...
 });
 
-function convertToRecordString(
-  input: Record<string, unknown>,
-): Record<string, string> {
+function convertToRecordString(input: unknown): Record<string, string> {
   const output: Record<string, string> = {};
 
-  const entries = Object.entries(input);
+  let tags = input;
```
bigfluffycookie, line 68: i feel like at some point we should create a helper function. Everytime we allow dynamic input for a field thats not a string we need to do this.  You think we need to check other places in our codebase where this is an issue?

MarceloRGonc: I think we have more places to check. Regarding the helper method, I don't think one method will be enough. I think we need a method for Property type. (Array, Object). But I'll merge it as it is now.

bigfluffycookie: Okay then please add a ticket for checking the other places, with maybe a note on it for the helper methods.

PR #1341, opened 2025-09-22
Handle empty testResult output in trigger event service

rmn-snpk:
Part of OPS-2672

In this pr we save empty test runs of triggers

### DIFF
--- a/packages/server/api/test/unit/trigger-event.service.test.ts
+++ b/packages/server/api/test/unit/trigger-event.service.test.ts
@@ -1,4 +1,12 @@
-import { PopulatedFlow, TriggerType } from '@openops/shared';
+import {
+  PopulatedFlow,
+  SeekPage,
+  TriggerEvent,
+  TriggerType,
+} from '@openops/shared';
+import { engineRunner, webhookUtils } from 'server-worker';
+import { accessTokenManager } from '../../src/app/authentication/lib/access-token-manager';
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/flows/trigger-events/trigger-event.service.ts
```diff
@@ -110,13 +110,22 @@ export const triggerEventService = {
           });
         }
 
-        for (const output of testResult.output) {
+        if (!testResult.output?.length) {
```
bigfluffycookie: Please switch the conditional around so the positive one is in here

packages/server/api/src/app/flows/trigger-events/trigger-event.service.ts
```diff
@@ -110,13 +110,22 @@ export const triggerEventService = {
           });
         }
 
-        for (const output of testResult.output) {
```
bigfluffycookie, line 113: If this was previously a bug please add test cases to ensure the bug is fixed

packages/server/api/src/app/flows/trigger-events/trigger-event.service.ts
rmn-snpk: done

packages/server/api/src/app/flows/trigger-events/trigger-event.service.ts
rmn-snpk: done

PR #1345, opened 2025-09-23
Add new email outlook trigger

bigfluffycookie:
Fixes OPS-2641

image


packages/blocks/microsoft-outlook/src/lib/triggers/new-email.ts
```diff
@@ -0,0 +1,213 @@
...
+
+interface FilterProps {
+  senderEmail?: string;
+  recipientEmail?: string;
+  ccEmail?: string;
+  subjectContains?: string;
+  subjectIs?: string;
+}
+
+const polling: Polling<
```
bigfluffycookie, line 155: taken from AP but added the filters

packages/blocks/microsoft-outlook/src/lib/triggers/new-email.ts
```diff
@@ -0,0 +1,218 @@
...
+  );
+}
+
+function matchesAnyFilter(targets: string[], filters: string[]): boolean {
+  return filters.some((filter) =>
+    targets.some((target) => target.includes(filter)),
+  );
+}
+
+function applyClientSideFilters(
```
bigfluffycookie, line 63: i really wanted to do server side filtering but i just couldnt get it to work, mainly theres and issue because using the lastDate logic with filtering by email, just doesnt work easily, one uss $search while the other uses $filter, and you cannot use those two together in microsoft graph

packages/blocks/microsoft-outlook/src/lib/triggers/new-email.ts
```diff
@@ -0,0 +1,222 @@
...
+      message.from?.emailAddress?.address || '',
+    );
+    const messageSenderName = normalizeString(
+      message.from?.emailAddress?.name || '',
+    );
+
+    if (!matchesAnyFilter([messageSender, messageSenderName], sendersToMatch)) {
+      return false;
+    }
+  }
```
MarceloRGonc: I think you have a bug here. If we have an array (propsValue['senders'] = ['']) and for some reason the emailAddress is empty (''). It will be handled as a valid filter.

Also, these blocks look very similar. I think you can extract into a function and reuse.

bigfluffycookie: done

PR #1356, opened 2025-09-24
Persist successful steps result to message history when stream is cancelled via abort signal

cezudas:
Fixes OPS-2706.

### DIFF
--- a/packages/server/api/src/app/ai/chat/user-message-handler.ts
+++ b/packages/server/api/src/app/ai/chat/user-message-handler.ts
@@ -153,13 +153,19 @@ async function streamLLMResponse(
           );
         }
       },
-      onAbort: async () => {
+      onAbort: ({ steps }) => {
         sendAiChatAbortedEvent({
           projectId: params.projectId,
           userId: params.userId,
           chatId: params.chatId,
           provider: params.aiConfig.provider,
         });
+        return saveChatHistory(
+          params.chatId,
+          params.userId,
+          params.projectId,
+          steps.flatMap((step) => step.response.messages),
+        );
       },
       abortSignal: params.abortSignal,
     });


packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -153,13 +153,19 @@ async function streamLLMResponse(
...
           userId: params.userId,
           chatId: params.chatId,
           provider: params.aiConfig.provider,
         });
+        return saveChatHistory(
+          params.chatId,
+          params.userId,
+          params.projectId,
+          steps.flatMap((step) => step.response.messages),
+        );
```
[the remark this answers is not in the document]
cezudas: that's okay, we have a catch block in at line 177

packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -153,13 +153,19 @@ async function streamLLMResponse(
...
           projectId: params.projectId,
           userId: params.userId,
           chatId: params.chatId,
           provider: params.aiConfig.provider,
         });
+        return saveChatHistory(
+          params.chatId,
+          params.userId,
+          params.projectId,
+          steps.flatMap((step) => step.response.messages),
```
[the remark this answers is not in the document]
cezudas: no, step.response.messages is defined. 

Screenshot 2025-09-24 at 18 06 40

packages/server/api/src/app/ai/chat/user-message-handler.ts
```diff
@@ -153,13 +153,19 @@ async function streamLLMResponse(
           );
         }
       },
-      onAbort: async () => {
+      onAbort: ({ steps }) => {
```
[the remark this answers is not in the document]
cezudas: that's okay, we have a catch block in at line 177

cezudas, line 156: Unfortunately, `steps` do not include the tool call that was intrerrupted upon abort event

MarceloRGonc: Will it work if we abort with the tool call without a tool response?

alexandrudanpop: good point, this could corrupt the calls after that

cezudas: https://ai-sdk.dev/docs/advanced/stopping-streams#handling-stream-abort-cleanup
I think the docs aren't very explicit, but it should be generally be safe to persist and resume from.


I looked into [their tests](https://github.com/vercel/ai/blob/main/packages/ai/src/generate-text/stream-text.test.ts#L12431-L12973) and my interpretation is:

- Steps are only "recorded" AFTER they're truly complete - meaning after onStepFinish is called AND the step is fully processed with all tool results
- If abort happens during tool execution, that step is NOT included in recordedSteps
- Only genuinely complete steps with matching tool calls and results are passed to onAbort

MarceloRGonc (approved):

PR #1359, opened 2025-09-25
Add trigger strategy for repeatable jobs

MarceloRGonc:
Fixes OPS-2696.

## Additional notes:
- If the trigger is scheduled, we continue using the flowVersionId as correlationId because we don't want to have runs in parallel. We want the second want to be ignored.

### DIFF
--- a/packages/server/api/src/app/workers/redis/redis-migration.ts
+++ b/packages/server/api/src/app/workers/redis/redis-migration.ts
@@ -13,6 +13,8 @@ import {
   ScheduleType,
 } from '@openops/shared';
 import { Job } from 'bullmq';
+import { blockMetadataService } from '../../blocks/block-metadata-service';
+import { flowVersionService } from '../../flows/flow-version/flow-version.service';
 import { flowRepo } from '../../flows/flow/flow.repo';
 import { bullMqGroups } from './redis-queue';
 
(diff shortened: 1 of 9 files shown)


packages/server/api/src/app/workers/redis/redis-migration.ts
```diff
@@ -91,6 +93,16 @@ async function migrateJob(job: Job<ScheduledJobData>): Promise<void> {
     modifiedJobData.executionType = undefined;
     await job.updateData(modifiedJobData);
   }
+
+  if (modifiedJobData.schemaVersion === 4) {
+    modifiedJobData.schemaVersion = 5;
```
MarceloRGonc, line 98: This is the way we have to run redis migrations

packages/server/api/src/app/workers/worker-controller.ts
```diff
@@ -155,19 +191,13 @@ export const flowWorkerController: FastifyPluginAsyncTypebox = async (app) => {
...
-        }
-
         return flowRunService.start({
           environment: RunEnvironment.PRODUCTION,
           flowVersionId,
           payload,
           synchronousHandlerId,
           projectId,
-          executionCorrelationId: correlationId,
+          executionCorrelationId: openOpsId(),
```
MarceloRGonc, line 200: If we have multiple payloads we want to generate a executionCorrelationId for each payload.

packages/server/api/src/app/workers/worker-controller.ts
```diff
@@ -127,6 +128,42 @@ export const flowWorkerController: FastifyPluginAsyncTypebox = async (app) => {
     },
   );
 
+  app.post(
+    '/submit-payload',
+    {
```
MarceloRGonc, line 133: Adding this endpoint for the cases we want to use a specific executionCorrelationId

packages/server/shared/src/lib/job/job-data.ts
```diff
@@ -6,7 +7,7 @@ import {
 } from '@openops/shared';
 import { Static, Type } from '@sinclair/typebox';
 
-export const LATEST_JOB_DATA_SCHEMA_VERSION = 4;
+export const LATEST_JOB_DATA_SCHEMA_VERSION = 5;
```
MarceloRGonc, line 10: We need to increment for the migration to run

PR #1362, opened 2025-09-25
Update IMAP error handling 

rSnapkoOpenOps:
Fixes OPS-2719.

### DIFF
--- a/packages/blocks/imap/src/lib/common/fetch-emails.ts
+++ b/packages/blocks/imap/src/lib/common/fetch-emails.ts
@@ -1,5 +1,6 @@
 import { BlockPropValueSchema } from '@openops/blocks-framework';
 import dayjs from 'dayjs';
+import { MailboxLockObject } from 'imapflow';
 import { ParsedMail } from 'mailparser';
 import { buildClient } from './build-client';
 import { buildImapSearch } from './build-search';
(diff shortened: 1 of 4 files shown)


packages/blocks/imap/src/lib/common/fetch-emails.ts
```diff
@@ -54,7 +56,9 @@ export async function fetchEmails({
     }
     return messages;
   } finally {
-    lock.release();
+    if (lock) {
+      lock.release();
+    }
```
rmn-snpk, line 61: We don't have the catch here to pass the error to fe

packages/blocks/imap/src/lib/common/mailbox.ts
```diff
@@ -29,6 +30,15 @@ export const mailbox = Property.Dropdown({
           value: mailbox.path,
         };
       });
+    } catch (error) {
+      logger.warn(`Failed to connect to IMAP`, {
+        error,
+      });
+      return {
+        disabled: true,
+        placeholder: 'The connection failed. Please try again later.',
```
MarceloRGonc: Please change this message.  I don't think we should put 'try again later'

packages/blocks/imap/src/lib/common/imap-auth.ts
```diff
@@ -36,10 +37,13 @@ export const imapAuth = BlockAuth.CustomAuth({
...
       return { valid: true };
-    } catch (e) {
+    } catch (error) {
+      logger.warn(`Failed to connect to IMAP`, {
+        error,
+      });
       return {
         valid: false,
-        error: JSON.stringify(e),
+        error: JSON.stringify(error),
```
MarceloRGonc: What happens to this JSON? Are we showing it on the UI?

rmn-snpk: Yes, it should be shown on the ui, but validation is broken https://linear.app/openops/issue/OPS-2319/auth-validation-does-not-work

packages/blocks/imap/src/lib/common/mailbox.ts
rmn-snpk: done

PR #1366, opened 2025-09-26
Add ai auth and analyze with ai empty block

bigfluffycookie:
Fixes OPS-2713 & Part of OPS-2712


packages/blocks/ai/.eslintrc.json
```diff
@@ -0,0 +1,33 @@
+{
```
bigfluffycookie, line 1: added an empty ai block as its needed to see the new auth

PR #1370, opened 2025-09-29
Update connections select to work outside step settings

rSnapkoOpenOps:
Part of OPS-2715.

### DIFF
--- a/packages/react-ui/src/app/features/builder/block-properties/auto-form-field-wrapper.tsx
+++ b/packages/react-ui/src/app/features/builder/block-properties/auto-form-field-wrapper.tsx
@@ -7,7 +7,7 @@ import {
   ToggleSwitch,
   ToggleSwitchOption,
 } from '@openops/components/ui';
-import { Action, isNil, Trigger } from '@openops/shared';
+import { isNil } from '@openops/shared';
 import { t } from 'i18next';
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/app/features/builder/block-properties/auto-form-field-wrapper.tsx
```diff
@@ -140,7 +140,7 @@ const AutoFormFieldWrapper = ({
   disabled,
   field,
 }: AutoFormFieldWrapperProps) => {
-  const form = useFormContext<Action | Trigger>();
+  const form = useFormContext();
```
rmn-snpk, line 143: The field can be used in different forms

packages/react-ui/src/app/features/builder/builder-types.ts
```diff
@@ -48,7 +48,7 @@ export type MidpanelAction =
   | {
       type: 'GENERATE_WITH_AI_CLICK';
       property?: BlockProperty & {
-        inputName: `settings.input.${string}` | `settings.${string}`;
+        inputName: `settings.input.${string}` | `settings.${string}` | string;
```
rmn-snpk, line 51: I think it's nice to have it this way so we can easily see some of 'templates'

alexandrudanpop (approved):

PR #1377, opened 2025-09-29
Mg/ops 2695

MarceloRGonc:
Part of OPS-2695.


packages/engine/src/lib/variables/processors/dropdown.ts
```diff
@@ -0,0 +1,55 @@
...
+  InputProperty,
+  PropertyType,
+  StaticPropsValue,
+} from '@openops/blocks-framework';
+import { tryParseJson } from '@openops/common';
+import { PropsValidationError } from '../props-processor';
+
+export const isDropdownProperty = (propertyType: PropertyType): boolean => {
+  return (
+    // propertyType === PropertyType.DROPDOWN ||
```
MarceloRGonc, line 12: Not sure that DROPDOWN should be included. We can't know what the expected type is. It can be a string or something else. We can always try to parse, but in that case, we assume that a string like this "{\"a\": \"b\"}" is expected as object.

packages/shared/src/lib/common/utils/object-utils.ts
```diff
@@ -47,49 +47,54 @@ export function sanitizeObjectForPostgresql<T>(input: T): T {
...
 ): Promise<T> {
   if (isNil(obj)) {
     return obj as T;
   } else if (isString(obj)) {
     return (await apply(obj)) as T;
   } else if (Array.isArray(obj)) {
-    for (let i = 0; i < obj.length; ++i) {
-      obj[i] = await applyFunctionToValues(obj[i], apply);
-    }
+    // Create a new array and map over it with Promise.all
```
bigfluffycookie: remove comments.

packages/engine/src/lib/variables/props-resolver.ts
```diff
@@ -0,0 +1,301 @@
...
+    return cp;
+  }
+  return `connection${cp}`;
+}
+
+function parseConnectionNameOnly(variableName: string): string | null {
+  const connectionWithNewFormatSquareBrackets = variableName.includes('[');
+  if (connectionWithNewFormatSquareBrackets) {
+    return parseSquareBracketConnectionPath(variableName);
+  }
```
bigfluffycookie, line 221: why is this commented out

packages/engine/src/lib/variables/props-processor.ts
```diff
@@ -0,0 +1,223 @@
...
+      }
+
+      const shouldValidate =
+        key !== AUTHENTICATION_PROPERTY_NAME &&
+        property.type !== PropertyType.MARKDOWN;
+      if (!shouldValidate) {
+        continue;
+      }
+
+      // If the value is required, we don't allow it to be undefined or null
```
bigfluffycookie: remove comment.

packages/blocks/framework/src/lib/trigger/trigger.ts
```diff
@@ -62,6 +62,7 @@ type BaseTriggerParams<
   name: string;
   displayName: string;
   description: string;
+  requireAuth?: boolean;
```
bigfluffycookie, line 65: why is the change with requireAuth necessary?

packages/engine/src/lib/variables/processors/dropdown.ts
bigfluffycookie: how is dropdown different to multi select dropdown in this sense?

packages/engine/src/lib/variables/processors/object.ts
```diff
@@ -0,0 +1,10 @@
+import { tryParseJson } from '@openops/common';
+import { ProcessorFn } from './types';
+
+export const objectProcessor: ProcessorFn = (_property, value) => {
+  if (typeof value === 'string') {
```
bigfluffycookie: i dont think you need this if, tryParseJson already does 
`if(typeof value !== string){
return value
}`

packages/engine/src/lib/variables/props-processor.ts
```diff
@@ -0,0 +1,223 @@
...
+      if (isNil(value) && property.required) {
+        errors[property.displayName] = [
+          formatErrorMessage(ErrorMessages.REQUIRED, { userInput: value }),
+        ];
+      }
+    }
+
+    for (const [key, value] of Object.entries(processedInput)) {
+      const property = props[key];
+      if (isNil(property)) {
```
bigfluffycookie: why not merge the two ifs

if (isNil(property) || (isNil(value) && !property.required)) {
  continue;
}

packages/engine/src/lib/variables/props-resolver.ts
```diff
@@ -0,0 +1,301 @@
...
+async function handleConnection(
+  params: ResolveSingleTokenParams,
+): Promise<unknown> {
+  const { variableName, engineToken, projectId, apiUrl, censoredInput } =
+    params;
+  const connectionName = parseConnectionNameOnly(variableName);
+  if (isNil(connectionName)) {
+    return '';
+  }
+  if (censoredInput) {
```
bigfluffycookie: move this if above the const connectionName = parseConnectionNameOnly(variableName);

no need to do any addtional logic in that case

packages/engine/src/lib/variables/props-resolver.ts
```diff
@@ -0,0 +1,301 @@
...
+  const connectionWithNewFormatSquareBrackets = variableName.includes('[');
+  if (connectionWithNewFormatSquareBrackets) {
+    return parseSquareBracketConnectionPath(variableName);
+  }
+  // {{connections.connectionName.path}}
+  // This does not work If connectionName contains .
+  return variableName.split('.')?.[1];
+}
+
+function parseSquareBracketConnectionPath(variableName: string): string | null {
```
bigfluffycookie, line 227: please remove comments

packages/engine/src/lib/variables/processors/dropdown.ts
MarceloRGonc: The dropdown, you don't know the expected type. Multi-selects, you know it's an array. Static dropdown you have the list of values, you can compare the types

packages/engine/src/lib/variables/props-resolver.ts
MarceloRGonc: You know that I merged from upstream right? But related to this comment, I think it makes sense since it looks like we have a limitation with the connection name.

bigfluffycookie: no i dont know, which changes are yours then? none of them?

packages/blocks/framework/src/lib/trigger/trigger.ts
MarceloRGonc: It allows us to do this :D 

image

image

bigfluffycookie (approved):

PR #1383, opened 2025-09-30
Add help method to do schema validator

MarceloRGonc:
Part of OPS-2695.


rita-gorokhod:
@MarceloRGonc I think it will be too complex to use in the blocks, I'm not sure we need it.

PR #1384, opened 2025-10-01
Fix error object parsing for invalid connection

ravikiranvm:
Fixes OPS-2742.

## Additional Notes

We have a few blocks which could send the error message as an object instead of a string. Example blocks include Jira, SMTP, Ternary. So this PR converts the error object to string if it is not already a string (in the frontend).

### DIFF
--- a/packages/react-ui/src/app/features/connections/components/create-edit-connection-dialog-content.tsx
+++ b/packages/react-ui/src/app/features/connections/components/create-edit-connection-dialog-content.tsx
@@ -184,6 +184,10 @@ const CreateEditConnectionDialogContent = ({
               msg:
                 typeof apError.params.error === 'string'
                   ? apError.params.error
+                  : apError.params.error &&
+                    typeof apError.params.error === 'object' &&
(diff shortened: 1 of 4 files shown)


packages/react-ui/src/app/features/connections/components/create-edit-connection-dialog-content.tsx
```diff
@@ -181,7 +181,14 @@ const CreateEditConnectionDialogContent = ({
         } else if (apError.code === ErrorCode.INVALID_APP_CONNECTION) {
           setErrorMessage(
             t('Connection failed with error {msg}', {
-              msg: apError.params.error,
+              msg:
+                typeof apError.params.error === 'string'
+                  ? apError.params.error
```
alexandrudanpop: Hi @ravikiranvm as we have multiple ternary expressions I would extract this to a function (maybe in a helpers file and add unit tests to this function, as the logic as I see becomes more complex.

alexandrudanpop (approved):

PR #1390, opened 2025-10-01
Add Sonnet 4.5 to anthropic and vertex providers

ravikiranvm:
Fixes OPS-2740.


packages/openops/src/lib/ai/providers/anthropic.ts
```diff
@@ -3,6 +3,7 @@ import { LanguageModel } from 'ai';
 import { AiProvider } from '../providers';
 
 const anthropicModels = [
+  'claude-sonnet-4-5',
```
bigfluffycookie: hm are you sure this model name is right? it doesnt match the others

PR #1394, opened 2025-10-01
Add nOps.io block integration

jordanatnops:
This pull request is to add an nOps.io integration to OpenOps. nOps.io is a Multi-cloud visibility platform with numerous SaaS integrations, a robust FinOps Agent, and automated optimization for AWS commitments and EKS.

Fixes OPS-2770

### DIFF
--- a/packages/blocks/nops/src/index.ts
+++ b/packages/blocks/nops/src/index.ts
@@ -7,7 +7,7 @@ import { BASE_NOPS_URL } from './lib/common/base-url';
 
 export const nops = createBlock({
   displayName: 'nOps',
-  description: 'Cloud cost optimization and FinOps platform',
+  description: 'Multi-cloud cost optimization and FinOps platform',
   auth: nopsAuth,
   minimumSupportedRelease: '0.20.0',
   logoUrl: 'https://static.openops.com/blocks/nops.png',


packages/blocks/nops/src/lib/auth.ts
```diff
@@ -0,0 +1,25 @@
+import { BlockAuth } from '@openops/blocks-framework';
+
+const markdown = `
```
rita-gorokhod, line 3: @jordanatnops Alternatively, you can also link to the documentation, if there is one (to have one source of truth)

packages/blocks/nops/src/lib/actions/get-cost-summary-action.ts
```diff
@@ -0,0 +1,174 @@
...
+      required: false,
+      options: {
+        options: [
+          { label: 'All (Resources and Non-Resources)', value: '' },
+          { label: 'Resources Only', value: 'false' },
+          { label: 'Non-Resources Only', value: 'true' },
+        ],
+      },
+    }),
+    accountIds: Property.Array({
```
rita-gorokhod, line 54: proposal: if there is an API endpoint to get accounts, this can be a multi-select dropdown property, that shows friendly values to the user.
For examples you can search for "Property.MultiSelectDropdown" or see for example https://github.com/openops-cloud/openops/blob/3b8564ec7629e661d55d67276512c8f063b7dd2d/packages/blocks/anodot/src/lib/account-property.ts#L6

packages/blocks/nops/src/lib/actions/get-cost-summary-action.ts
```diff
@@ -0,0 +1,174 @@
...
+      displayName: 'Account IDs',
+      description: 'Filter by account IDs (supports multiple values)',
+      required: false,
+    }),
+    regions: Property.Array({
```
rita-gorokhod, line 59: are those AWS regions? if so, you can reuse maybe https://github.com/openops-cloud/openops/blob/3b8564ec7629e661d55d67276512c8f063b7dd2d/packages/openops/src/lib/aws/regions.ts#L50

or if there is an API on your side to pull available regions, then that can be dynamic too.

packages/blocks/nops/src/lib/actions/get-cost-summary-action.ts
```diff
@@ -0,0 +1,174 @@
...
+      displayName: 'Regions',
+      description: 'Filter by regions (supports multiple values)',
+      required: false,
+    }),
+    products: Property.Array({
```
rita-gorokhod, line 64: Same comment here that if you'd like, you can change this to be dynamic and pull from the API the possible values.

rita-gorokhod:
Superseded by #1403

PR #1421, opened 2025-10-07
Add gcp and azure custom api actions

bigfluffycookie:
Fixes OPS-2730

### DIFF
--- a/packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
+++ b/packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
@@ -1,0 +1,17 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
+import { authenticateUserWithAzure, azureAuth } from '@openops/common';
+
+export const customAzureApiCallAction = createCustomApiCallAction({
+  auth: azureAuth,
+  name: 'custom_azure_api_call',
+  description: 'Make a custom REST API call to Azure',
+  displayName: 'Custom Azure API Call',
+  baseUrl: () => 'https://management.azure.com?api-version=2024-10-21',
+  authMapping: async (context: any) => {
+    const { access_token } = await authenticateUserWithAzure(context.auth);
+    return {
+      Authorization: `Bearer ${access_token}`,
+      'Content-Type': 'application/json',
+    };
+  },
+});
--- a/packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
+++ b/packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
@@ -1,0 +1,33 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
+import { googleCloudAuth } from '@openops/common';
+import { runCommands } from '../google-cloud-cli';
+
+async function getAccessToken(auth: any): Promise<string> {
+  const outputs = await runCommands(
+    ['auth print-access-token'],
+    auth,
+    false,
+    undefined,
+    'gcloud',
+  );
+  const token = outputs[0]?.trim();
+  if (!token) {
+    throw new Error('Failed to obtain Google Cloud access token');
+  }
+  return token;
+}
+
+export const customGoogleCloudApiCallAction = createCustomApiCallAction({
+  auth: googleCloudAuth,
+  name: 'google_custom_api_call',
+  description: 'Make a custom REST API call to Google Cloud',
+  displayName: 'Custom Google Cloud API Call',
+  baseUrl: () => 'https://www.googleapis.com',
+  authMapping: async (context: any) => {
+    const token = await getAccessToken(context.auth);
+    return {
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'application/json',
+    };
+  },
+});
--- a/packages/blocks/azure/src/index.ts
+++ b/packages/blocks/azure/src/index.ts
@@ -3,6 +3,7 @@ import { azureAuth } from '@openops/common';
 import { BlockCategory } from '@openops/shared';
 import { advisorAction } from './lib/actions/azure-advisor-action';
 import { azureCliAction } from './lib/actions/azure-cli-action';
+import { customAzureApiCallAction } from './lib/actions/custom-azure-api-action';
 
 export const azure = createBlock({
   displayName: 'Azure',
@@ -11,6 +12,6 @@ export const azure = createBlock({
   logoUrl: 'https://static.openops.com/blocks/azure.svg',
   authors: [],
   categories: [BlockCategory.CLOUD],
-  actions: [azureCliAction, advisorAction],
+  actions: [azureCliAction, advisorAction, customAzureApiCallAction],
   triggers: [],
 });
--- a/packages/blocks/azure/test/index.test.ts
+++ b/packages/blocks/azure/test/index.test.ts
@@ -12,7 +12,7 @@ import { azure } from '../src/index';
 
 describe('block declaration tests', () => {
   test('should return block with correct actions', () => {
-    expect(Object.keys(azure.actions()).length).toBe(2);
+    expect(Object.keys(azure.actions()).length).toBe(3);
     expect(azure.actions()).toMatchObject({
       azure_cli: {
         name: 'azure_cli',
@@ -20,6 +20,10 @@ describe('block declaration tests', () => {
       },
       advisor: {
         name: 'advisor',
+        requireAuth: true,
+      },
+      custom_azure_api_call: {
+        name: 'custom_azure_api_call',
         requireAuth: true,
       },
     });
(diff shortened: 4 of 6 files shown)


packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -0,0 +1,17 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
+import { authenticateUserWithAzure, azureAuth } from '@openops/common';
+
+export const customAzureApiCallAction = createCustomApiCallAction({
+  auth: azureAuth,
+  name: 'custom_azure_api_call',
+  description: 'Make a custom REST API call to Azure',
+  displayName: 'Custom Azure API Call',
+  baseUrl: () => 'https://management.azure.com?api-version=2024-10-21',
```
bigfluffycookie: it doesnt work without version, i think it makes sense to add it in the baseurl, alternatively we can add it to the description

packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
```diff
@@ -0,0 +1,33 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
+import { googleCloudAuth } from '@openops/common';
+import { runCommands } from '../google-cloud-cli';
+
+async function getAccessToken(auth: any): Promise<string> {
```
bigfluffycookie: there are other ways to do this, but this is the easiest that does not require a change to our current auth, while also using our existing structure

packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
```diff
@@ -0,0 +1,33 @@
...
+    'gcloud',
+  );
+  const token = outputs[0]?.trim();
+  if (!token) {
+    throw new Error('Failed to obtain Google Cloud access token');
+  }
+  return token;
+}
+
+export const customGoogleCloudApiCallAction = createCustomApiCallAction({
```
bigfluffycookie, line 26: you can test it with a get to (https://cloudresourcemanager.googleapis.com/v1/projects

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
rita-gorokhod: I think base url is fine yep, maybe should add a "/" at the end? otherwise the user has to do it manually

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -0,0 +1,17 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
+import { authenticateUserWithAzure, azureAuth } from '@openops/common';
+
+export const customAzureApiCallAction = createCustomApiCallAction({
```
rita-gorokhod, line 29: is there a link to the docs we can add?

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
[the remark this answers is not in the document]
bigfluffycookie: hmmm maybe i do this?

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -0,0 +1,17 @@
...
+  auth: azureAuth,
+  name: 'custom_azure_api_call',
+  description: 'Make a custom REST API call to Azure',
+  displayName: 'Custom Azure API Call',
+  baseUrl: () => 'https://management.azure.com?api-version=2024-10-21',
+  authMapping: async (context: any) => {
+    const { access_token } = await authenticateUserWithAzure(context.auth);
```
rita-gorokhod: what if we use host credentials?

packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
```diff
@@ -0,0 +1,33 @@
...
+  auth: googleCloudAuth,
+  name: 'google_custom_api_call',
+  description: 'Make a custom REST API call to Google Cloud',
+  displayName: 'Custom Google Cloud API Call',
+  baseUrl: () => 'https://www.googleapis.com',
```
rita-gorokhod: same comments as azure with / and docs

packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
```diff
@@ -0,0 +1,33 @@
...
+  authMapping: async (context: any) => {
```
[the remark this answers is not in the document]
bigfluffycookie: its always any in this context

packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
rita-gorokhod: does it work for both enableHostSession and without?

bigfluffycookie: nope, also forgot to add projects dropdown, same with azure

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
bigfluffycookie: we dont normally add the / in other custom apis, should i still do it? usually it comes along when copying the endpoint to be honest

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
bigfluffycookie: done

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
rita-gorokhod: I think I'd add it, yeah, most APIs can handle two //

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -0,0 +1,62 @@
...
+  authenticateUserWithAzure,
+  azureAuth,
+  getUseHostSessionProperty,
+} from '@openops/common';
+import { runCommand } from '../azure-cli';
+import { subDropdown } from '../common-properties';
+
+const getHostAccessToken = async (
+  auth: unknown,
+  subscription?: string,
```
rita-gorokhod: why do you want the user to pass the subscription? 
does it matter for http calls token?

packages/blocks/google-cloud/src/lib/actions/custom-google-cloud-api-action.ts
```diff
@@ -0,0 +1,58 @@
+import { createCustomApiCallAction } from '@openops/blocks-common';
+import { Property } from '@openops/blocks-framework';
+import { getUseHostSessionProperty, googleCloudAuth } from '@openops/common';
+import { projectCliDropdown } from '../common-properties';
+import { runCommands } from '../google-cloud-cli';
+
+async function getAccessToken(
+  auth: unknown,
+  useHostSession: boolean,
+  project?: string,
```
rita-gorokhod, line 10: same comment, why do we need project for the authentication?

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
bigfluffycookie: hmmm, i get what you mean, i do think we need it for the host, from what i understand when you get the token from the cli, the subscription you select changes the tennatId, but i should put the subscription together with the useHostCredentials

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -0,0 +1,93 @@
...
+              options: {
+                disabled: true,
+                options: [],
+                placeholder: `Something went wrong fetching subscriptions`,
+                error: `${error}`,
+              },
+            }),
+          } as any;
+        }
+        return {};
```
rita-gorokhod, line 58: as a best practice for readability, it's best to short circuit and return this condition first

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -0,0 +1,93 @@
...
+        useHostSession,
+      }): Promise<{ [key: string]: any }> => {
+        try {
+          if (useHostSession?.['useHostSessionCheckbox'] as boolean) {
+            return {
+              subDropdown: await getSubscriptionsDropdownForHostSession(auth),
+            } as any;
+          }
+        } catch (error) {
+          return {
```
rita-gorokhod, line 54: this should be inside getSubscriptionsDropdownForHostSession

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
bigfluffycookie: personally i disagree, because we then have an inverted if, i think its harder to read that way. Now with the try catch moved i think this way is fine

PR #1426, opened 2025-10-07
Allow object keys to accept and resolve dynamic values

ravikiranvm:
Fixes OPS-2683.

## Additional Notes

Fix:

https://www.loom.com/share/787d73a96c754267b1eeacc319570b18?sid=55d217cb-ce18-4891-8ff0-11a91aca7aff

### DIFF
--- a/packages/react-ui/src/app/features/blocks/lib/blocks-api.ts
+++ b/packages/react-ui/src/app/features/blocks/lib/blocks-api.ts
@@ -11,6 +11,7 @@ import {
   AddBlockRequestBody,
   BlockOptionRequest,
   BlockScope,
+  EngineErrorResponse,
   GetBlockRequestParams,
   GetBlockRequestQuery,
   ListBlocksRequestQuery,
@@ -36,9 +37,9 @@ export const blocksApi = {
       version: request.version ?? undefined,
     });
   },
-  options<T extends DropdownState<unknown> | BlockPropertyMap>(
-    request: BlockOptionRequest,
-  ): Promise<T> {
+  options<
+    T extends DropdownState<unknown> | BlockPropertyMap | EngineErrorResponse,
+  >(request: BlockOptionRequest): Promise<T> {
     return api.post<T>(`/v1/blocks/options`, request);
   },
   mapToMetadata(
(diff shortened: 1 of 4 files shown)


rmn-snpk:
It doesn't work correctly in case key is an object. I see [Object object] 
image

rSnapkoOpenOps:
When switching to dynamic some of the pairs are lost https://www.loom.com/share/ba8b4d8c81134fcb9afb02be05ff19bf?sid=5912592e-e243-40d0-8bbd-3e5b28904c1e

rSnapkoOpenOps (changes requested):
See comment

rmn-snpk:
> It doesn't work correctly in case key is an object. I see [Object object] image

We don't expect user to use object as a key. They always can stringify the value themselves. This comment is not relevant anymore

rSnapkoOpenOps (approved):

MarceloRGonc:
> > It doesn't work correctly in case key is an object. I see [Object object] image
> 
> We don't expect user to use object as a key. They always can stringify the value themselves. This comment is not relevant anymore

Has this been discussed? I think we should return an error in these situations. We're hiding an error from the user. Is this expected? @rita-gorokhod Other than that, the PR is ready. But I think we should throw an error.

MarceloRGonc (changes requested):
I spoke with Rita and, if the key obtained is not valid (e.g., [Object Object]), we should throw an exception.

Therefore, the behaviour should be as follows:
- If the key is a primitive type (boolean, string, number), we should stringify it.
- If the key is an object, we should throw an exception.

packages/shared/src/lib/common/utils/object-utils.ts
```diff
@@ -80,14 +80,14 @@ export async function applyFunctionToValues<T>(
...
     );
     return newArray as unknown as T;
   } else if (isObject(obj)) {
     const newEntries = await Promise.all(
       Object.entries(obj).map(async ([key, value]) => [
-        key,
-        await applyFunctionToValues(value, apply),
+        await apply(key),
+        await applyFunctionToKeysAndValues(value, apply),
       ]),
```
MarceloRGonc: Something like this.
```
      Object.entries(obj).map(async ([key, value]) => {
        const key2 = await apply(key);

        if (isObject(key2) || Array.isArray(key2)) {
          throw new Error('Error');
        }

        return [key2, await applyFunctionToKeysAndValues(value, apply)];
      }),

ravikiranvm: Done

PR #1427, opened 2025-10-07
Add CommunityBadge to menu header and general settings page

rmn-snpk:
Fixes OPS-2707.

https://www.loom.com/share/71b5f7da7d6149268b3cae23e34847e3?sid=ce23a536-208d-4b86-bb41-2cde9f5f6a95

### DIFF
--- a/packages/react-ui/src/app/features/navigation/side-menu/dashboard/dashboard-side-menu-header.tsx
+++ b/packages/react-ui/src/app/features/navigation/side-menu/dashboard/dashboard-side-menu-header.tsx
@@ -1,11 +1,14 @@
 import { AppLogo } from '@/app/common/components/app-logo';
+import { flagsHooks } from '@/app/common/hooks/flags-hooks';
 import { useAppStore } from '@/app/store/app-store';
 import {
   Button,
   cn,
+  CommunityBadge,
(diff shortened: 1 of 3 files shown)


packages/server/api/src/app/flags/flag.service.ts
```diff
@@ -157,7 +157,7 @@ export const flagService = {
       },
       {
         id: FlagId.SHOW_COMMUNITY,
-        value: false,
+        value: true,
```
rmn-snpk, line 160: Will be disabled in the EE

packages/react-ui/src/app/features/navigation/side-menu/dashboard/dashboard-side-menu-header.tsx
rmn-snpk, line 1: Just added the badge and slightly adapted styles

packages/server/api/src/app/flags/flag.service.ts
alexandrudanpop: Is there a PR for it?

rmn-snpk: Not yet, will add right after this one is merged

PR #1429, opened 2025-10-07
Update dependabot to group all the PRs

rmn-snpk:
Fixes OPS-2805.

In theory it should work, let's observe for a bit


maor-rozenfeld:
Why? It's better to have the two separate.

rmn-snpk:
> Why? It's better to have the two separate.

PR #1431, opened 2025-10-07
Add telemetry support in Ask AI action

rSnapkoOpenOps:
Fixes OPS-2800.

### DIFF
--- a/packages/blocks/ai/src/lib/actions/askAi.ts
+++ b/packages/blocks/ai/src/lib/actions/askAi.ts
@@ -8,6 +8,7 @@ import {
   getAiModelFromConnection,
   getAiProvider,
   getAiProviderLanguageModel,
+  isLLMTelemetryEnabled,
 } from '@openops/common';
 import { AiProviderEnum, analysisLLMSchema } from '@openops/shared';
 import { generateObject } from 'ai';
@@ -112,6 +113,7 @@ export const askAi = createAction({
       prompt: composedPrompt,
       schema: analysisLLMSchema,
       ...((modelSettings as Record<string, unknown>) ?? {}),
+      experimental_telemetry: { isEnabled: isLLMTelemetryEnabled() },
     });
     return result.object;
   },


packages/blocks/ai/test/askAi-action.test.ts
```diff
@@ -59,12 +60,15 @@ describe('analyze action', () => {
...
-      schema: analysisLLMSchema,
-      maxRetries: 2,
-    });
+    expect(generateObject).toHaveBeenCalledWith(
+      expect.objectContaining({
+        model: 'languageModel',
+        prompt: 'Hello',
+        schema: analysisLLMSchema,
+        maxRetries: 2,
+        experimental_telemetry: { isEnabled: false },
```
rita-gorokhod, line 69: Use test cases that it passes the correct param + verify that it called the method to fetch the param

packages/engine/src/main.ts
```diff
@@ -29,6 +34,25 @@ function installCodeBlockDependencies(): void {
   );
 }
 
+let telemetrySDK: NodeSDK | undefined;
+
+function initTelemetry(): void {
+  telemetrySDK = isLLMTelemetryEnabled()
+    ? new NodeSDK({
```
rita-gorokhod: This is the same code as in app.ts, right? Can you reuse?

rmn-snpk: Done

packages/blocks/ai/test/askAi-action.test.ts
rmn-snpk: I added separate test focused on telemetry

jest.setup.js
```diff
@@ -0,0 +1,7 @@
+// Mock langfuse-vercel to prevent dynamic import errors in Jest
+jest.mock('langfuse-vercel', () => ({
+  LangfuseExporter: jest.fn().mockImplementation(() => ({
+    export: jest.fn(),
+    shutdown: jest.fn(),
+  })),
+}));
```
rmn-snpk, line 7: Added it globally to avoid errors in all our packages

packages/blocks/ai/test/askAi-action.test.ts
```diff
@@ -188,6 +197,40 @@ describe('analyze action', () => {
 
     expect(result).toBe('ok');
   });
+  test('should set experimental_telemetry based on isLLMTelemetryEnabled', async () => {
```
rita-gorokhod:
```
test.each([true, false])('should set experimental_telemetry based on isLLMTelemetryEnabled=%p', async (isLLMTelemetryEnabled) => {
    isLLMTelemetryEnabled.mockReturnValueOnce(isLLMTelemetryEnabled);

...
    await askAi.run(context as any);

    expect(generateObject).toHaveBeenCalledWith(
      expect.objectContaining({ experimental_telemetry: { isEnabled: isLLMTelemetryEnabled } }),
    );
  });
});
```

packages/blocks/ai/test/askAi-action.test.ts
```diff
@@ -59,12 +60,20 @@ describe('analyze action', () => {
...
+      expect.objectContaining({
+        model: 'languageModel',
+        prompt: 'Hello',
+        schema: analysisLLMSchema,
+        maxRetries: 2,
+        experimental_telemetry: { isEnabled: false },
+      }),
+    );
+
+    const { isLLMTelemetryEnabled } = jest.requireMock('@openops/common') as {
```
rita-gorokhod: this part is not needed

packages/openops/src/lib/ai/providers.ts
```diff
@@ -135,6 +139,20 @@ export const isLLMTelemetryEnabled = () =>
   !!system.get(SharedSystemProp.LANGFUSE_SECRET_KEY) &&
   !!system.get(SharedSystemProp.LANGFUSE_PUBLIC_KEY);
 
+export const getAiTelemetrySDK = () => {
```
rita-gorokhod, line 142: minor: why not expose startAiTelemeryIfEnabled() ?
(getAiTelemetrySDK can stay private)

packages/server/api/src/app/app.ts
```diff
@@ -73,17 +69,7 @@ import { workerModule } from './workers/worker-module';
...
-        traceExporter: new LangfuseExporter({
-          secretKey: system.get(SharedSystemProp.LANGFUSE_SECRET_KEY),
-          publicKey: system.get(SharedSystemProp.LANGFUSE_PUBLIC_KEY),
-          baseUrl: system.get(SharedSystemProp.LANGFUSE_HOST),
-          environment: system.get(SharedSystemProp.ENVIRONMENT_NAME),
-        }) as SpanExporter,
-        instrumentations: [getNodeAutoInstrumentations()],
-      })
-    : undefined;
+  const otelSDK = getAiTelemetrySDK();
```
rita-gorokhod, line 72: I see that we also shut it down, should we do it for the engine too or is it not needed?

packages/openops/src/lib/ai/providers.ts
rmn-snpk: We need to do a shutdown, so I would keep it separate so the consumer has full control

packages/server/api/src/app/app.ts
rmn-snpk: I added cleanup to the engine

PR #1432, opened 2025-10-08
Get full body for email

bigfluffycookie:
Fixes OPS-2817


packages/blocks/microsoft-outlook/src/lib/triggers/new-email.ts
```diff
@@ -132,9 +132,15 @@ async function fetchMessages(
       ? '$top=10'
       : `$filter=receivedDateTime gt ${dayjs(lastFetchEpochMS).toISOString()}`;
 
+  const headers: Record<string, string> = {
+    ConsistencyLevel: 'eventual',
+    Prefer: 'outlook.body-content-type="html"',
```
bigfluffycookie, line 137: can be text or html, discussed with marco to use html to avoid data loss

PR #1444, opened 2025-10-10
Prevent Date objects from being converted to empty objects during PostgreSQL sanitization

cezudas:
Fixes OPS-2838.

Object.entries(dateObject) → returns [] (Date objects have no enumerable properties)
Object.fromEntries([]) → returns {}

### DIFF
--- a/packages/shared/src/lib/common/utils/object-utils.ts
+++ b/packages/shared/src/lib/common/utils/object-utils.ts
@@ -59,8 +59,6 @@ export function applyFunctionToValuesSync<T>(
     return obj.map((item) =>
       applyFunctionToValuesSync(item, apply),
     ) as unknown as T;
-  } else if (obj instanceof Date) {
-    return obj as T;
   } else if (isObject(obj)) {
     return Object.fromEntries(
       Object.entries(obj).map(([key, value]) => [


packages/shared/src/lib/common/utils/object-utils.ts
```diff
@@ -59,6 +59,8 @@ export function applyFunctionToValuesSync<T>(
     return obj.map((item) =>
       applyFunctionToValuesSync(item, apply),
     ) as unknown as T;
+  } else if (obj instanceof Date) {
+    return obj as T;
```
MarceloRGonc: I don't think you should make this change. I think the isObject method is wrong.

packages/shared/src/lib/common/utils/object-utils.ts
```diff
@@ -83,6 +85,8 @@ export async function applyFunctionToValues<T>(
       obj.map((item) => applyFunctionToValues(item, apply)),
     );
     return newArray as unknown as T;
+  } else if (obj instanceof Date) {
+    return obj as T;
   } else if (isObject(obj)) {
```
MarceloRGonc, line 86:
```
export const isObject = (obj: unknown): obj is Record<string, unknown> => {
  if (typeof obj !== 'object' || obj === null || Array.isArray(obj)) {
    return false;
  }
  
  const prototype = Object.getPrototypeOf(obj);
  return prototype === Object.prototype || prototype === null;
};
```

cezudas: yes, agree. ty.
Done

PR #1450, opened 2025-10-13
Rename requireToolApproval to isWriteAction

bigfluffycookie:
Part of OPS-2807


packages/blocks/framework/src/lib/action/action.ts
```diff
@@ -41,7 +41,7 @@ type CreateActionParams<
   test?: ActionRunner<BlockAuth, ActionProps>;
   requireAuth?: boolean;
   errorHandlingOptions?: ErrorHandlingOptionsParam;
-  requireToolApproval: boolean;
+  IsWriteAction: boolean;
```
cezudas: I think the convention is to use camelCase for type properties instead of pascal case.
`IsWriteAction` -> `isWriteAction`

PR #1454, opened 2025-10-13
Switch tabs on triggering a test with keyboard shortcut

ravikiranvm:
Fixes OPS-2636

## Additional Notes

Tested scenarios:
- Tested *CMD + G* in action/trigger toggles the tab to test and triggers the test.
- Triggers tests from test tab.
- Prevent tab switch when form is invalid or in read only mode.
- Prevents test triggering in read only mode.
- Disables test/retest button in read only mode. 

Deployed to UX

### DIFF
--- a/packages/react-ui/src/app/features/builder/test-step/index.tsx
+++ b/packages/react-ui/src/app/features/builder/test-step/index.tsx
@@ -114,7 +114,7 @@ const TestStepContainer = React.memo(
             <TestStepDataViewer
               outputJson={selectedStep?.settings?.inputUiInfo?.sampleData ?? ''}
               onChange={useSaveSelectedStepSampleData}
-              readonly={false}
+              readonly={readOnly}
               theme={theme}
               containerClassName="h-full"
             />


packages/react-ui/src/app/features/builder/test-step/index.tsx
```diff
@@ -48,12 +57,14 @@ const TestStepContainer = React.memo(
         <TabsList className="grid grid-cols-2 bg-transparent border-none rounded-none w-full">
           <TabsTrigger
             value={TabListEnum.STEP_OUTPUT}
+            disabled={readOnly}
```
ravikiranvm: I observed that these tabs are active when in read only mode. So now we disable these tabs in read only mode.

packages/react-ui/src/app/features/builder/test-step/test-action-section.tsx
```diff
@@ -142,9 +150,9 @@ const TestActionSection = React.memo(
               size="sm"
               onClick={handleTest}
               keyboardShortcut="G"
-              onKeyboardShortcut={mutate}
+              onKeyboardShortcut={readOnly ? undefined : mutate}
```
ravikiranvm, line 153: Since now we added an option to exit the edit mode, it is possible that users can exit this mode while test tab is active. Now we disable this button in read only mode and also disable the keyboard shortcut in read only mode.

packages/react-ui/src/app/features/builder/test-step/index.tsx
rmn-snpk: In case I exited edit mode on test tab I'm not able to switch to configuration. I think we need to enable all the tabs
image

packages/react-ui/src/app/features/builder/step-settings/index.tsx
```diff
@@ -225,6 +231,12 @@ const StepSettingsContainer = React.memo(() => {
   const sidebarHeaderContainerRef = useRef<HTMLDivElement>(null);
   const modifiedStep = form.getValues();
 
+  const [activeTab, setActiveTab] = useState('configure');
+  const onTestCallback = useCallback(() => {
+    if (readonly) return;
+    setActiveTab('test');
+  }, [readonly]);
+
```
rmn-snpk: Please extract tab names into enum

packages/react-ui/src/app/features/builder/test-step/index.tsx
```diff
@@ -101,7 +114,7 @@ const TestStepContainer = React.memo(
             <TestStepDataViewer
               outputJson={selectedStep?.settings?.inputUiInfo?.sampleData ?? ''}
               onChange={useSaveSelectedStepSampleData}
-              readonly={false}
+              readonly={readOnly}
```
ravikiranvm, line 117: We're supposed to pass the readonly value to this component.

packages/react-ui/src/app/features/builder/step-settings/index.tsx
ravikiranvm: Done

packages/react-ui/src/app/features/builder/test-step/index.tsx
ravikiranvm: Done

PR #1456, opened 2025-10-13
Generate an id every time we save a request body in Redis

MarceloRGonc:
Fixes OPS-2833.

## Additional notes:
- When the LLM ​​is generating the workflow, calls to the engine are made in parallel, which means that one call to the server generates multiple calls to the engine.

### DIFF
--- a/packages/blocks/slack/src/lib/actions/send-message-action.ts
+++ b/packages/blocks/slack/src/lib/actions/send-message-action.ts
@@ -46,7 +46,7 @@ export const slackSendMessageAction = createAction({
     assertNotNullOrUndefined(conversationId, 'conversationId');
     if (!isBlock) {
       text = context.propsValue['text']['text'];
-      const headerText = context.propsValue['headerText']['headerText'];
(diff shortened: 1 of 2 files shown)


packages/engine/src/engine-executor.ts
```diff
@@ -25,7 +25,7 @@ export async function executeEngine(
 
   const duration = Math.floor(performance.now() - startTime);
 
-  const key = await saveRequestBody(requestId, result);
```
bigfluffycookie, line 28: hmmm, im only a bit concerned about this change, was this requestId also a random openOpsId() ?

MarceloRGonc: This is the key used on redis. It was always a random ID, but before this change, it was synced with the requestId generated by Fastify.

bigfluffycookie: gotcha, and its okay for us that its no longer synced? we werent using it for anything else?

MarceloRGonc: We can identify the keys through the logs. But we haven't used the sync for anything to date.

PR #1457, opened 2025-10-14
Make possible to configure fixed height in pixels from within markdown code snippets

cezudas:
Fixes OPS-2806.
Notice the Execute GraphQL query action renders a markdown with two graphql code snippets.
Under the hood, the Markdown rendered uses Monaco editor, which makes it impossible to set up graceful sizing policies with CSS alone.

It would look much nicer if we could adjust the height of each snippet individually.
Screenshot 2025-10-14 at 10 26 24

### DIFF
--- a/packages/ui-components/src/components/custom/markdown.tsx
+++ b/packages/ui-components/src/components/custom/markdown.tsx
@@ -124,7 +124,7 @@ const Markdown = React.memo(
 
               const codeContent = String(props.children).trim();
 
-              const meta = (props as any).node?.data?.meta || '';
+              const meta = props.node?.data?.meta || '';
               const fullClassName = meta
                 ? `${props.className} ${meta}`
                 : props.className;


packages/ui-components/src/components/code-editor/types.ts
```diff
@@ -26,4 +26,5 @@ export type MonacoLanguage =
   | 'powershell'
   | 'bat'
   | 'ini'
+  | 'graphql'
```
cezudas, line 29: Enable colored syntax for graphql snippets as well, not exactly related to this ticket.

packages/ui-components/src/components/code-editor/code-utils.ts
```diff
@@ -19,6 +19,7 @@ export const getLanguageExtensionForCode = (
     return 'typescript';
   if (className.includes('language-jsx')) return 'javascript';
   if (className.includes('language-tsx')) return 'typescript';
+  if (className.includes('language-graphql')) return 'graphql';
```
cezudas, line 22: Enable colored syntax for graphql snippets as well, not exactly related to this ticket.

PR #1460, opened 2025-10-14
Add test run limits dialog

rmn-snpk:
Part of OPS-2808 and Fixes OPS-2831.
https://www.loom.com/share/d09cffb1518d41d2a55da0c5b6af5268?sid=c60c649e-6ebf-4fec-9608-aae6ddcba511

### DIFF
--- a/packages/react-ui/src/app/features/builder/flow-canvas/widgets/test-flow-widget.tsx
+++ b/packages/react-ui/src/app/features/builder/flow-canvas/widgets/test-flow-widget.tsx
@@ -1,5 +1,6 @@
 import {
   Button,
+  cn,
   INTERNAL_ERROR_TOAST,
   isMacUserAgent,
   toast,
@@ -95,12 +96,12 @@ const TestFlowWidget = ({ flowVersion, setRun }: TestFlowWidgetProps) => {
   return (
     flowVersion.valid && (
       <>
-        <div className="h-8 flex rounded-full text-base font-medium border-collapse">
+        <div className="h-8 flex rounded-full text-base font-medium">
           <Tooltip>
             <TooltipTrigger asChild>
               <Button
                 variant="ghost"
-                className="h-8 !bg-primary-200/20 dark:!bg-primary-200 text-primary-300 disabled:pointer-events-auto enabled:hover:!border-primary-200 enabled:hover:!text-primary-300 border-primary-200/50 border-[2px] border-r-[1px] rounded-l-full animate-fade z-10"
+                className="h-8 !bg-primary-200/20 dark:!bg-primary-200 text-primary-300 disabled:pointer-events-auto enabled:hover:!border-primary-200 enabled:hover:!text-primary-300 border-primary-200/70 border-r-primary-200/35 border-[1px] rounded-l-full animate-fade z-10"
                 disabled={!triggerHasSampleData}
                 loading={isPending || isLoading}
                 onClick={() => handleExecuting()}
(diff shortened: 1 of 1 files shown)


packages/react-ui/src/app/features/builder/test-run-limits/test-run-limits-dialog-hook.ts
```diff
@@ -0,0 +1,84 @@
...
+      const actionMeta = (blockMeta?.suggestedActions || []).find(
+        (a) => a?.name === actionName,
+      );
+
+      map[blockName].actions[actionName] =
+        actionMeta?.displayName ?? actionName;
+    }
+
+    return map;
+  }, [testRunActionLimits, metadata]);
```
rmn-snpk, line 63: Build a lookup map for the limits dialog.

We do this by going through the configured limits and grouping them based on their blockName. Each block and action is then enriched with human-friendly display names from metadata when available.

packages/react-ui/src/app/features/builder/test-run-limits/test-run-limits-dialog-hook.ts
```diff
@@ -0,0 +1,84 @@
...
+  );
+
+  const { metadata } = blocksHooks.useAllStepsMetadata({
+    type: 'action',
+  });
+
+  const blockActionMetaMap: BlockActionLimitMetadata = useMemo(() => {
+    const map: BlockActionLimitMetadata = {};
+
+    const limits = testRunActionLimits?.limits;
```
cezudas: I think you can remove optional chaining. If `limits` was possibly undefined then you would gen an error at line 33.

```suggestion
    const limits = testRunActionLimits.limits;
```

packages/react-ui/src/app/features/builder/test-run-limits/test-run-limits-dialog-hook.ts
```diff
@@ -0,0 +1,84 @@
...
+    for (const limit of limits) {
+      const blockName = limit.blockName;
+      const actionName = limit.actionName;
+
+      const blockMeta = metadata?.find(
+        (metadata) =>
+          metadata.type === ActionType.BLOCK &&
+          'blockName' in metadata &&
+          metadata.blockName === blockName,
+      ) as BlockStepMetadataWithSuggestions;
```
cezudas: The cast assumes the found metadata is always BlockStepMetadataWithSuggestions, but find() can return undefined. This could cause runtime errors when accessing blockMeta?.suggestedActions.

packages/react-ui/src/app/features/builder/test-run-limits/test-run-limits-dialog.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <DialogHeader>
+          <DialogTitle className="text-primary text-[22px] font-bold">
+            {t('Test run limits')}
+          </DialogTitle>
+        </DialogHeader>
+        <TestRunLimitsForm
+          value={testRunActionLimits}
+          onSave={(v) => {
+            setIsOpen(false);
+            onSave(v);
```
cezudas, line 34: I think the dialog should close only if the save operation successfully completes.

rmn-snpk: It goes via apply operation, so that means that the only loader we can do is subscribe to saving changes and close the dialog when it's done. But this is ugly and we have a loader on publish button

packages/react-ui/src/app/features/builder/test-run-limits/test-run-limits-dialog-hook.ts
rmn-snpk: updated

PR #1465, opened 2025-10-14
Prevent duplicate workspace creation when updating admin user

MarceloRGonc:
Fixes OPS-2845

## Summary
Previously, changing the admin password triggered the creation of a new workspace in OpenOps Tables. This caused the operation to fail because the workspace ID didn’t match the one already stored in the organisation table.

Issue:
- When the admin password was changed, a new workspace was mistakenly created even though the organisation already existed.
- The first attempt failed because of mismatched organisation IDs.
- The second attempt succeeded after the password change, but it left two workspaces associated with the same organisation.

This update adds a safeguard to check whether the organisation already exists before attempting to create a new workspace.

### DIFF
--- a/packages/server/api/src/main.ts
+++ b/packages/server/api/src/main.ts
@@ -90,7 +90,7 @@ const main = async (): Promise<void> => {
     await createOpenOpsTablesMcpEndpoint();
     await updateOpenopsTablesDatabase();
     await deleteOldOpportunitiesTable();
-    // await seedDevData();
+    await seedDevData();
 
     await seedTemplateTables();
     await seedOpportunitesTemplateTable();


maor-rozenfeld:
Test it in various scenarios please

MarceloRGonc:
> Test it in various scenarios please

So I tested:
- Change password
- Change email
- Remove the organisation without removing the workspace in Tables -> workspace reused

packages/server/api/src/app/database/seeds/seed-admin.ts
```diff
@@ -20,14 +20,23 @@ export const upsertAdminUser = async (): Promise<void> => {
     logger.debug(`Failed to sign in as admin [${email}]`, e);
 
     const user = await ensureUserExists(email, password);
+
+    const organizationAlreadyCreated = await hasExistingOrganization(user);
+    const projectAlreadyCreated = await hasExistingProject(user);
+    if (organizationAlreadyCreated && projectAlreadyCreated) {
```
rita-gorokhod: what if the org was created but project wasn't?
I think we should have new wrappers createOrganizationIfNotExists and createProjectIfNotExists

MarceloRGonc: I reverted some changes. Behavior:
- We try to create the workspace and database in tables or get the existing one.
- We try to create the organisation or get the existing one.
- We try to create the project or get the existing one.

MarceloRGonc:
> looks good but whats the testing plan for this?

It's only change the admin password and see the result.

PR #1469, opened 2025-10-15
Force dynamic properties to be mandatory

rita-gorokhod:
Fixes OPS-2861
The problem is that if the property is not mandatory, the LLM (rightfully) doesn't pass it when building the workflow, which can lead to bugs such as https://github.com/openops-cloud/openops/pull/1438.

In practice, dynamic properties will always return either a sub property, or {}. So they can be marked as required=true.
This will serve as a guardrail for any future properties that will be added.

### DIFF
--- a/packages/blocks/azure/test/azure-advisor-action.test.ts
+++ b/packages/blocks/azure/test/azure-advisor-action.test.ts
@@ -14,7 +14,6 @@ const azureCliMock = {
 
 jest.mock('../src/lib/azure-cli', () => azureCliMock);
 
-import { tr } from 'zod/v4/locales';
 import { advisorAction } from '../src/lib/actions/azure-advisor-action';
 
 const auth = {


packages/blocks/framework/src/lib/property/input/common.ts
```diff
@@ -16,6 +16,26 @@ export const SupportsAISchema = Type.Object({
...
+    type: Type.Literal(propertyType),
+    required: Type.Literal(true),
+    defaultValue: Type.Optional(Type.Any()),
+  });
+
+export type TMandatoryPropertyValue<T, U extends PropertyType> = TPropertyValue<
+  T,
+  U,
+  ValidationInputType.ANY,
+  true
```
rita-gorokhod: re-uses the TPropertyValue defined below and passes the final boolean TRUE for required

packages/blocks/framework/src/lib/property/input/dynamic-prop.ts
```diff
@@ -38,9 +37,4 @@ export type DynamicProperties<R extends boolean> = BasePropertySchema & {
     ctx: PropertyContext,
   ) => Promise<Record<string, DynamicProp>>;
   refreshers: string[];
-} & TPropertyValue<
```
rita-gorokhod, line 41: this previously allowed setting required:false (the <R> generics)

packages/blocks/framework/src/lib/property/input/common.ts
```diff
@@ -16,6 +16,26 @@ export const SupportsAISchema = Type.Object({
...
+export const TMandatoryPropertyValue = <
+  T extends TSchema,
+  U extends PropertyType,
+>(
+  T: T,
+  propertyType: U,
+): TObject =>
+  Type.Object({
+    type: Type.Literal(propertyType),
+    required: Type.Literal(true),
```
rita-gorokhod: new schema with required always set to true

packages/blocks/framework/src/lib/property/input/dynamic-prop.ts
```diff
@@ -38,9 +37,4 @@ export type DynamicProperties<R extends boolean> = BasePropertySchema & {
...
-    DynamicPropsValue,
-    PropertyType.DYNAMIC,
-    ValidationInputType.ANY,
-    R
```
MarceloRGonc, line 45: I think you just need to change the 'R' to 'true'.

packages/blocks/framework/src/lib/property/input/common.ts
MarceloRGonc: I don't think we need a new schema.

MarceloRGonc (commented):
If the goal is for all dynamic properties to be required, we don't need a new schema. We just need to change the existing one to only allow 'true'

packages/blocks/framework/src/lib/property/input/dynamic-prop.ts
```diff
@@ -38,9 +37,4 @@ export type DynamicProperties<R extends boolean> = BasePropertySchema & {
...
-  >;
+} & TMandatoryPropertyValue<DynamicPropsValue, PropertyType.DYNAMIC>;
```
MarceloRGonc:
```suggestion
} & TPropertyValue<
    DynamicPropsValue,
    PropertyType.DYNAMIC,
    ValidationInputType.ANY,
    true
  >;
```

packages/blocks/framework/src/lib/property/input/common.ts
rita-gorokhod: how would you do it without and still force a compilation error if anyone tries to set it to false?

packages/blocks/framework/src/lib/property/input/dynamic-prop.ts
rita-gorokhod: this only passes the param as default, it doesn't create a compilation error if you set it to false.

MarceloRGonc: I see a compilation error if I try to put false.

image

image

PR #1471, opened 2025-10-15
Refactor chat expiration handling to use configurable system property

alexandrudanpop:
Fixes OPS-2823


packages/server/shared/src/lib/system/system.ts
```diff
@@ -95,6 +95,7 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
   [AppSystemProp.REQUEST_BODY_LIMIT]: '10',
   [SharedSystemProp.LANGFUSE_HOST]: 'https://us.cloud.langfuse.com',
   [AppSystemProp.MAX_LLM_CALLS_WITHOUT_INTERACTION]: '100',
+  [AppSystemProp.LLM_CHAT_EXPIRE_TIME]: '86400', // 24 hours
```
cezudas: Consider making the metric explicit: e.g. `LLM_CHAT_EXPIRE_TIME_SECONDS`

rita-gorokhod: Agreed

alexandrudanpop: fixed

PR #1473, opened 2025-10-15
Fix emails not containing thread in reply outlook action

bigfluffycookie:
Fixes OPS-2854

Screenshot 2025-10-15 at 18 20 15

### DIFF
--- a/packages/blocks/microsoft-outlook/src/lib/actions/reply-email.ts
+++ b/packages/blocks/microsoft-outlook/src/lib/actions/reply-email.ts
@@ -62,18 +62,6 @@ export const replyEmailAction = createAction({
         }
       },
     }),
-    bodyFormat: Property.StaticDropdown({
-      displayName: 'Body Format',
-      required: true,
-      defaultValue: 'text',
-      options: {
-        disabled: false,
-        options: [
-          { label: 'HTML', value: 'html' },
-          { label: 'Text', value: 'text' },
-        ],
-      },
-    }),
     replyBody: Property.LongText({
       displayName: 'Reply Body',
       required: true,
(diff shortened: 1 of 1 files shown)


packages/blocks/microsoft-outlook/src/lib/actions/reply-email.ts
```diff
@@ -116,11 +116,15 @@ export const replyEmailAction = createAction({
       file: WorkflowFile;
       fileName: string;
     }>;
+
+    const formattedReplyBody =
```
bigfluffycookie: so, the thing is outlook either allows body OR comment, its not possible to have both see -> https://learn.microsoft.com/en-us/graph/api/message-reply?view=graph-rest-1.0&tabs=http 

"Specify either a comment or the body property of the message parameter. Specifying both will return an HTTP 400 Bad Request error."

so this is my solution to still keep body format in logically, wdyt?

packages/blocks/microsoft-outlook/src/lib/actions/reply-email.ts
```diff
@@ -116,11 +116,16 @@ export const replyEmailAction = createAction({
...
```
bigfluffycookie: ok so, we dont actually need this i just tested it again, technically i dont think bodyFormat is necessary at all, but we cannot just remove it. This just adds some save escaping and all that. 

we could alternatively also just ignore the body format parameter entirely tbh

bigfluffycookie: we could also just leave it and ignore it, and maybe we use it again at a later time

packages/blocks/microsoft-outlook/src/lib/actions/reply-email.ts
```diff
@@ -62,18 +62,6 @@ export const replyEmailAction = createAction({
         }
       },
     }),
-    bodyFormat: Property.StaticDropdown({
```
bigfluffycookie, line 65: we dont need it anymore in comment, it accepts both text/html, doesnt care about the format

PR #1480, opened 2025-10-17
Allow auth to be null when authentication is not required

MarceloRGonc:
Fixes OPS-2856.

## Additional Notes

image

image

### DIFF
--- a/packages/engine/src/lib/variables/props-resolver.ts
+++ b/packages/engine/src/lib/variables/props-resolver.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { applyFunctionToValues, isNil, isString } from '@openops/shared';
+import { applyFunctionToKeysAndValues, isNil, isString } from '@openops/shared';
 import { initCodeSandbox } from '../core/code/code-sandbox';
 import { FlowExecutorContext } from '../handler/context/flow-execution-context';
 import { createConnectionService } from '../services/connections.service';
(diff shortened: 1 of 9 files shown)


packages/engine/src/lib/variables/props-processor.ts
```diff
@@ -41,7 +41,7 @@ export const propsProcessor = {
     if (isAuthenticationProperty) {
       const { processedInput: authProcessedInput, errors: authErrors } =
         await propsProcessor.applyProcessorsAndValidators(
-          resolvedInput[AUTHENTICATION_PROPERTY_NAME],
+          resolvedInput[AUTHENTICATION_PROPERTY_NAME] ?? {},
```
MarceloRGonc, line 44: It was already like this in the variable service.

rita-gorokhod: @MarceloRGonc why did it get here? requireAuth should've been false and it never should've entered this IF

MarceloRGonc: RequireAuth is true because we want to present the property for authentication. The property itself is not required. If we set requireAuth to true the property is not displayed.

See this: https://github.com/openops-cloud/openops/pull/1377#discussion_r2394871252

rita-gorokhod: as we disscussed offline, you added more details:

> 'requireAuth' is a property that AP added that enables or disables the authentication property. If it's disabled you don't see the Auth property in the UI
> 'required' is the property that says if a property is required or not.
> In this case 'requireAuth' is true, because the property should be on the UI. But it shouldn't be required.

PR #1482, opened 2025-10-17
Add termination reason back

bigfluffycookie:
Part of OPS-2814


packages/engine/src/lib/handler/context/flow-execution-context.ts
```diff
@@ -257,12 +258,14 @@ export class FlowExecutorContext {
             ...baseExecutionOutput,
             error: this.error,
             status: FlowRunStatus.INTERNAL_ERROR,
+            terminationReason: 'Flow execution encountered an internal error',
```
bigfluffycookie, line 260: idk if i was just supposed to leave them undefined or not

packages/server/worker/src/lib/executors/flow-job-executor.ts
```diff
@@ -202,13 +202,24 @@ async function updateRunWithError(
     | FlowRunStatus.STOPPED
     | FlowRunStatus.INTERNAL_ERROR,
 ): Promise<void> {
+  let terminationReason: string | undefined;
+
+  if (status === FlowRunStatus.TIMEOUT) {
```
bigfluffycookie: same here, i guess i can set them to undefined

MarceloRGonc: I don't think you need to make this change. I think you only need to set this values on the 'packages/engine/src/lib/handler/context/flow-execution-context.ts' file

bigfluffycookie: oki, removed

MarceloRGonc: I was wrong sorry :)

MarceloRGonc (approved):

PR #1485, opened 2025-10-20
Add  Servicenow get record and delete record actions

cezudas:
Part of OPS-2366.

### DIFF
--- a/packages/blocks/servicenow/src/actions/delete-record-action.ts
+++ b/packages/blocks/servicenow/src/actions/delete-record-action.ts
@@ -1,6 +1,7 @@
 import { httpClient, HttpMethod } from '@openops/blocks-common';
 import { createAction, Property } from '@openops/blocks-framework';
 import { servicenowAuth, ServiceNowAuth } from '../lib/auth';
+import { buildServiceNowApiUrl } from '../lib/build-api-url';
 import { generateAuthHeader } from '../lib/generate-auth-header';
 import { servicenowTableDropdownProperty } from '../lib/table-dropdown-property';
 
@@ -24,7 +25,7 @@ export const deleteRecordAction = createAction({
 
     await httpClient.sendRequest({
       method: HttpMethod.DELETE,
-      url: `https://${auth.instanceName}.service-now.com/api/now/table/${tableName}/${sysId}`,
+      url: buildServiceNowApiUrl(auth, `${tableName}/${sysId}`),
       headers: {
         ...generateAuthHeader({
           username: auth.username,
--- a/packages/blocks/servicenow/src/actions/get-record-action.ts
+++ b/packages/blocks/servicenow/src/actions/get-record-action.ts
@@ -6,6 +6,7 @@ import {
 } from '@openops/blocks-framework';
 import { logger } from '@openops/server-shared';
 import { servicenowAuth, ServiceNowAuth } from '../lib/auth';
+import { buildServiceNowApiUrl } from '../lib/build-api-url';
 import { generateAuthHeader } from '../lib/generate-auth-header';
 import { getServiceNowTableFields } from '../lib/get-table-fields';
 import { servicenowTableDropdownProperty } from '../lib/table-dropdown-property';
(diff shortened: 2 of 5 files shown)


packages/blocks/servicenow/src/actions/delete-record-action.ts
```diff
@@ -0,0 +1,39 @@
...
+      headers: {
+        ...generateAuthHeader({
+          username: auth.username,
+          password: auth.password,
+        }),
+        Accept: 'application/json',
+      },
+    });
+
+    return response.body;
```
rmn-snpk: The API response is empty, let's add some message there, like

"The record with id "123" was deleted"

packages/blocks/servicenow/src/actions/delete-record-action.ts
```diff
@@ -0,0 +1,42 @@
...
+      required: true,
+    }),
+  },
+  async run(context) {
+    const auth = context.auth as ServiceNowAuth;
+    const { tableName, sysId } = context.propsValue;
+
+    await httpClient.sendRequest({
+      method: HttpMethod.DELETE,
+      url: `https://${auth.instanceName}.service-now.com/api/now/table/${tableName}/${sysId}`,
```
bigfluffycookie: please extract the base url to somewhere, maybe 'getBaseUrl' or something

packages/blocks/servicenow/src/actions/delete-record-action.ts
```diff
@@ -0,0 +1,42 @@
...
+          username: auth.username,
+          password: auth.password,
+        }),
+        Accept: 'application/json',
+      },
+    });
+
+    return {
+      success: true,
+      message: `The record with id "${sysId}" was deleted`,
```
bigfluffycookie, line 40: does the request itself not return anything?

packages/blocks/servicenow/src/actions/get-record-action.ts
```diff
@@ -0,0 +1,97 @@
...
+      },
+    }),
+  },
+  async run(context) {
+    const auth = context.auth as ServiceNowAuth;
+    const { tableName, sysId, fields } = context.propsValue;
+
+    const queryParams: Record<string, string> = {};
+
+    if (fields && Array.isArray(fields) && fields.length > 0) {
```
bigfluffycookie:
```suggestion
    if (Array.isArray(fields) && fields.length) {
```

packages/blocks/servicenow/src/lib/get-tables.ts
```diff
@@ -0,0 +1,33 @@
...
+    method: HttpMethod.GET,
+    url: `https://${auth.instanceName}.service-now.com/api/now/table/sys_db_object`,
+    headers: {
+      ...generateAuthHeader({
+        username: auth.username,
+        password: auth.password,
+      }),
+      Accept: 'application/json',
+    },
+    queryParams: {
```
bigfluffycookie, line 26: what was the reason for these default values? is there any chance users would want to adjust them? just curious

packages/blocks/servicenow/src/actions/delete-record-action.ts
cezudas: the body is {} , status 200.  Roman suggested I improve this.

packages/blocks/servicenow/src/lib/get-tables.ts
cezudas:
```
queryParams: {
      sysparm_fields: 'name,label', // only retrieve what we need, optimizes the response payload size
      sysparm_limit: '10000', // limit how many records can be returned in the payload.  this is just for the dropdown options, so no need for user to adjust them.
      sysparm_query: 'nameSTARTSWITHx_', // this is essential because each ServiceNow instance will bring 6000+ sys tables. Our dropdowns cannot handle so many records. 
    },
```

packages/blocks/servicenow/src/actions/delete-record-action.ts
cezudas: ✅

packages/blocks/servicenow/src/actions/get-record-action.ts
cezudas: ✅

PR #1489, opened 2025-10-20
Use the same trigger payload structure for workflow run and test 

MarceloRGonc:
Fixes OPS-2866.

### DIFF
--- a/packages/server/api/src/app/flows/flow-version/flow-version.service.ts
+++ b/packages/server/api/src/app/flows/flow-version/flow-version.service.ts
@@ -42,6 +42,7 @@ import { FlowVersionEntity } from './flow-version-entity';
 import { flowVersionSideEffects } from './flow-version-side-effects';
 import {
   calculateTestRunActionLimits,
+  mergeTestRunLimits,
   shouldRecalculateTestRunActionLimits,
   tryIncrementalUpdate,
 } from './test-run-action-limits-calculator';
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -409,10 +409,14 @@ export const flowRunService = {
...
 
+    const triggerPayload = payload.find(
+      (o) => o.stepId === flowVersion.trigger.id,
+    )?.output;
+
     return this.start({
       projectId,
       flowVersionId,
-      payload,
+      payload: triggerPayload,
```
MarceloRGonc: Before we were wrapping the trigger payload in an array

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -409,10 +409,13 @@ export const flowRunService = {
       stepIds: [flowVersion.trigger.id],
     });
 
+    const triggerPayload =
```
rita-gorokhod: shouldn't it be payload[0] ?
what other items are in the array?

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
rita-gorokhod: why is this a problem only in the new email trigger? why does it work in other triggers like jira?

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
MarceloRGonc: I expect the array to have only one element. So, payload[0] works, but i think the way I have right now it's safer.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
MarceloRGonc: It's for all the triggers.

Before my change:
image

After:
image

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
MarceloRGonc: Changed to:
```
payload[0]?.output || {}
```

PR #1491, opened 2025-10-20
Add additional Microsoft Teams scopes for enhanced permissions 

rmn-snpk:
Fixes OPS-2898.

tested without admin permission, works fine

### DIFF
--- a/packages/blocks/microsoft-teams/src/lib/common/microsoft-teams-auth.ts
+++ b/packages/blocks/microsoft-teams/src/lib/common/microsoft-teams-auth.ts
@@ -10,7 +10,13 @@ export const microsoftTeamsAuth = BlockAuth.OAuth2({
   scope: [
     'User.Read',
     'ChannelMessage.Send',
+    'Channel.ReadBasic.All',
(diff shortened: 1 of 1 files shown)


packages/blocks/microsoft-teams/src/lib/common/microsoft-teams-auth.ts
```diff
@@ -10,7 +10,13 @@ export const microsoftTeamsAuth = BlockAuth.OAuth2({
   scope: [
```
rita-gorokhod, line 10: how did it work before this?
and will this require users to re-authenticate or re-create the connection?

rmn-snpk: We have different ways to login. 

For basic login flow  ("Connect" button with our app credentials) previous list of permissions worked fine, but it was not enough for app credentials. I don't have an answer why it works like that, just keep in mind it's microsoft. 

Existing connections will continue to work as before. Reconnection will only be required for user app credentials, which I assume were not used before.

PR #1492, opened 2025-10-20
Add add/update record, get records actions for ServiceNow

cezudas:
Part of OPS-2366.

### DIFF
--- a/packages/blocks/servicenow/src/actions/get-records-action.ts
+++ b/packages/blocks/servicenow/src/actions/get-records-action.ts
@@ -93,7 +93,7 @@ export const getRecordsAction = createAction({
 
     const selectedFields = (fields as { selected?: string[] })?.selected;
 
-    if (Array.isArray(selectedFields) && selectedFields) {
+    if (Array.isArray(selectedFields)) {
       queryParams['sysparm_fields'] = selectedFields.join(',');
     }
 
--- a/packages/blocks/servicenow/test/actions/get-records-action.test.ts
+++ b/packages/blocks/servicenow/test/actions/get-records-action.test.ts
@@ -95,7 +95,6 @@ describe('get_records action', () => {
           filters: [],
         },
         limit: 100,
-        offset: 0,
       },
     };
 
@@ -130,7 +129,6 @@ describe('get_records action', () => {
           filters: [],
         },
         limit: 100,
-        offset: 0,
       },
     };
 
@@ -171,7 +169,6 @@ describe('get_records action', () => {
           ],
         },
         limit: 100,
-        offset: 0,
       },
     };
 
@@ -210,7 +207,6 @@ describe('get_records action', () => {
           ],
         },
         limit: 100,
-        offset: 0,
       },
     };
 
@@ -249,7 +245,6 @@ describe('get_records action', () => {
           ],
         },
         limit: 100,
-        offset: 0,
       },
     };
 


packages/blocks/servicenow/src/actions/create-fields-properties.ts
```diff
@@ -0,0 +1,127 @@
+import { Property } from '@openops/blocks-framework';
+import { ServiceNowAuth } from '../lib/auth';
+import { createFieldValueProperty } from '../lib/create-field-value-property';
+import { getServiceNowTableFields } from '../lib/get-table-fields';
+
+/**
```
bigfluffycookie: please remove the comments, unless you think they are absolutely necessary to understanding the code (this goes for all the files)

packages/blocks/servicenow/src/actions/create-fields-properties.ts
```diff
@@ -0,0 +1,127 @@
...
+ * Filters table fields to only include writable, non-primary, non-collection fields
+ */
+function filterWritableFields(
+  tableFields: Awaited<ReturnType<typeof getServiceNowTableFields>>,
```
bigfluffycookie: why  `tableFields: Awaited<ReturnType<typeof getServiceNowTableFields>>`

instead of just ` tableFields: ServiceNowTableField[]` ? seems like a strange pattern to me, same for all the ones below

packages/blocks/servicenow/src/actions/create-fields-properties.ts
```diff
@@ -0,0 +1,127 @@
...
+          },
+        }),
+        [valuePropertyName]: Property.DynamicProperties({
+          displayName: valueDisplayName,
+          required: true,
+          refreshers: ['fieldName'],
+          props: async ({ fieldName }) => {
+            const currentField = fieldName as unknown as string;
+
+            if (!currentField) {
```
bigfluffycookie, line 54: maybe in here we could instead return a property with a placeholder text? (or is that not a thing within array properties?)

packages/blocks/servicenow/src/actions/create-filters-properties.ts
```diff
@@ -0,0 +1,108 @@
...
+import { isSingleValueFilter, ViewFilterTypesEnum } from '@openops/common';
+import { ServiceNowAuth } from '../lib/auth';
+import { createFieldValueProperty } from '../lib/create-field-value-property';
+import { getServiceNowTableFields } from '../lib/get-table-fields';
+
+export async function createFiltersProperties(
+  auth: ServiceNowAuth,
+  tableName: string,
+) {
+  let tableFields: Awaited<ReturnType<typeof getServiceNowTableFields>> = [];
```
bigfluffycookie: again here, weird typing

packages/blocks/servicenow/src/actions/create-filters-properties.ts
```diff
@@ -0,0 +1,108 @@
...
+  let hasError = false;
+
+  try {
+    tableFields = await getServiceNowTableFields(auth, tableName);
+  } catch (error) {
+    console.error(
```
bigfluffycookie: logger

packages/blocks/servicenow/test/actions/list-records-action.test.ts
```diff
@@ -0,0 +1,267 @@
...
+            },
+            {
+              fieldName: 'priority',
+              filterType: ViewFilterTypesEnum.lower_than,
+              value: { value: '3' },
+            },
+          ],
+        },
+        limit: 50,
+        offset: 0,
```
bigfluffycookie: hm where did this come from? its not in any of the actions

packages/blocks/servicenow/src/actions/list-records-action.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    const queryParams: Record<string, string> = {};
+
+    const filtersProps = context.propsValue.filters['filters'] as unknown as {
+      fieldName: string;
+      value: { value: unknown };
+      filterType: ViewFilterTypesEnum;
+    }[];
+
+    if (filtersProps && filtersProps.length > 0) {
```
bigfluffycookie: (filtersProps?.length > 0)

packages/blocks/servicenow/src/actions/create-fields-properties.ts
cezudas: this same pattern is in openops-tables,  I would keep it consistent if you don't mind.

packages/blocks/servicenow/src/actions/create-filters-properties.ts
```diff
@@ -0,0 +1,112 @@
...
+  let tableFields: ServiceNowTableField[] = [];
+  let hasError = false;
+
+  try {
+    tableFields = await getServiceNowTableFields(auth, tableName);
+  } catch (error) {
+    logger.error(
+      `Failed to fetch ServiceNow table fields for ${tableName}:`,
+      error,
+    );
```
rmn-snpk: Maybe use warn instead so we don't get notifications

packages/blocks/servicenow/src/actions/get-records-action.ts
```diff
@@ -0,0 +1,114 @@
...
+      const query = buildServiceNowQuery(filters, filterType);
+
+      if (query) {
+        queryParams['sysparm_query'] = query;
+      }
+    }
+
+    if (limit !== undefined && limit !== null) {
+      queryParams['sysparm_limit'] = String(limit);
+    }
```
rmn-snpk, line 92: I think that setting the limit to 0 has no sense, should we include it?

packages/blocks/servicenow/src/actions/create-filters-properties.ts
cezudas: done

packages/blocks/servicenow/src/actions/get-records-action.ts
cezudas: I used 

```
    limit: Property.Number({
      displayName: 'Limit',
      description: 'Maximum number of records to return. Default is 10000.',
      required: true,
      defaultValue: 10000,
      validators: [Validators.minValue(1)],
    }),
```

servicenow will use a limit of 10000 by default
Screenshot 2025-10-21 at 19 09 10

packages/blocks/servicenow/src/actions/get-records-action.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const selectedFields = (fields as { selected?: string[] })?.selected;
+
+    if (Array.isArray(selectedFields) && selectedFields) {
```
bigfluffycookie: the second check is redundant

rSnapkoOpenOps (approved):

packages/blocks/servicenow/test/actions/list-records-action.test.ts
bigfluffycookie: idk if youve accidentally resolved this without removing or if you forgot to reply to it so i am adding another comment. this param makes me think that there is pagination or something like it that we are ignoring, i would like to know why its

cezudas: Probably I forgot to check in those changes.
Servicenow tables APIs support this optional offset parameter, I added it initially but Roman talked be out of using it.
I removed it in the meantime from the test mock as well.

PR #1495, opened 2025-10-20
Add ServiceNow requests specific actions

cezudas:
Part of OPS-2366.


packages/blocks/servicenow/src/actions/requests/add-request-action.ts
```diff
@@ -0,0 +1,63 @@
...
+      }[]) ?? [];
+
+    const body: Record<string, any> = {};
+    for (const { fieldName, fieldValue } of fields) {
+      body[fieldName] = fieldValue['fieldValue'];
+    }
+
+    const response = await httpClient.sendRequest({
+      method: HttpMethod.POST,
+      url: buildServiceNowApiUrl(auth, TABLE_NAME),
```
bigfluffycookie: im a bit confused here, the code in this action is almost identical to packages/blocks/servicenow/src/actions/create-record-action.ts, in fact i even added it to a diff comparison and the only difference it showed was that in the other action this like is 
      `url: buildServiceNowApiUrl(auth, tableName as string),`

can it really be the same endpoint? i dont get it

rmn-snpk: Why not to use basic "Add record" action?  Can't we select the table there?

rmn-snpk: As discussed in Slack, we keep it separate as logically it's different action

packages/blocks/servicenow/src/actions/requests/get-requests-action.ts
```diff
@@ -0,0 +1,68 @@
...
+            value: FilterType.AND,
+          },
+          {
+            label: 'Match any filter',
+            value: FilterType.OR,
+          },
+        ],
+      },
+      defaultValue: FilterType.AND,
+    }),
```
rmn-snpk, line 36: Why not to make it required? We have default value in the run

packages/blocks/servicenow/src/actions/action-runners.ts
```diff
@@ -0,0 +1,231 @@
...
+    const query = buildServiceNowQuery(filters, filterType || FilterType.AND);
+
+    if (query) {
+      queryParams['sysparm_query'] = query;
+    }
+  }
+
+  if (limit !== undefined && limit !== null) {
+    queryParams['sysparm_limit'] = String(limit);
+  }
```
rmn-snpk: limit is required prop, do we still need this check?

bigfluffycookie (approved):

packages/blocks/servicenow/src/actions/requests/get-requests-action.ts
cezudas: It's the same as in Openops tables 🤷
```
    filterType: Property.StaticDropdown({
      displayName: 'Filter type',
      required: false,
      options: {
        options: [
          {
            label: 'Match all filters',
            value: FilterType.AND,
          },
          {
            label: 'Match any filter',
            value: FilterType.OR,
          },
        ],
      },
      defaultValue: FilterType.AND,
    }),
```

rmn-snpk: Got it

PR #1503, opened 2025-10-21
Do not save trigger polling state when timeout is reached

MarceloRGonc:
Fixes OPS-2900.

## Additional Notes
- If the trigger execution times out, we don't lose the data. It will be fetched in the subsequent execution

### DIFF
--- a/packages/server/worker/src/lib/engine/call-engine.ts
+++ b/packages/server/worker/src/lib/engine/call-engine.ts
@@ -183,7 +183,7 @@ function logEngineError(
     status = EngineResponseStatus.TIMEOUT;
     errorMessage = 'Engine execution timed out.';
 
-    logger.debug(errorMessage, {
+    logger.info(errorMessage, {
       error,
       requestId,
       errorTimestamp,


packages/blocks/common/src/lib/polling/index.ts
```diff
@@ -71,10 +73,15 @@ export const pollingHelper = {
...
         );
-        await store.put('lastPoll', newLastEpochMilliSeconds);
-        return items
+
+        const itemsToPoll = items
           .filter((f) => f.epochMilliSeconds > lastEpochMilliSeconds)
           .map((item) => item.data);
+
+        throwIfExecutionTimeExceeded();
+        await store.put('lastPoll', newLastEpochMilliSeconds);
```
MarceloRGonc, line 82: We check the timeout before updating the key. Ideally, this wouldn't be done this way. However, polling triggers depend on the value stored in storage.

packages/blocks/kion/jest.config.ts
```diff
@@ -2,6 +2,7 @@
 export default {
   displayName: 'blocks-kion',
   preset: '../../../jest.preset.js',
+  setupFiles: ['../../../jest.env.js'],
```
MarceloRGonc, line 5: To add the default env vars for the tests. After the updates one of the tests was failing because of missing environment variable

packages/blocks/microsoft-outlook/src/lib/triggers/new-email.ts
```diff
@@ -127,24 +127,25 @@ async function fetchMessages(
...
 
-  let response: PageCollection = await client
-    .api(`${baseUrl}?${filter}&$select=${SELECT_FIELDS}`)
+  let req = client
+    .api(baseUrl)
+    .select(SELECT_FIELDS)
     .orderby('receivedDateTime desc')
-    .headers(headers)
-    .get();
+    .headers(headers);
```
MarceloRGonc, line 139: The default page size is 10

PR #1518, opened 2025-10-22
Add Slack configuration to disable link unfurling

rita-gorokhod:
Fixes OPS-2658

image


packages/blocks/slack/src/lib/common/utils.ts
```diff
@@ -51,6 +52,8 @@ export const slackSendMessage = async ({
         event_type: 'slack-message',
         event_payload: eventPayload,
       },
+      unfurl_links: unfurlLinksAndMedia ?? true,
```
rita-gorokhod, line 55: backwards compatibility. true is the default slack behavior.

rSnapkoOpenOps (approved):

PR #1522, opened 2025-10-23
Add connection property to http block

bigfluffycookie:
Fixes OPS-2887

### DIFF
--- a/packages/blocks/http/src/lib/common/auth.ts
+++ b/packages/blocks/http/src/lib/common/auth.ts
@@ -24,12 +24,12 @@ export const httpAuth = BlockAuth.CustomAuth({
       properties: {
         key: Property.ShortText({
           displayName: 'Header Key',
-          required: false,
+          required: true,
           defaultValue: 'Authorization',
         }),
         value: Property.SecretText({
           displayName: 'Header Value',
-          required: false,
+          required: true,
         }),
       },
     }),


packages/blocks/http/src/lib/actions/send-http-request-action.ts
```diff
@@ -168,11 +170,33 @@ export const httpSendRequestAction = createAction({
     await validateHost(url);
     await validateHost(context.propsValue.proxy_settings?.proxy_host);
 
+    const toLowerCaseKeys = (obj: HttpHeaders) =>
```
MarceloRGonc: Could you move this function outside of this block?

packages/blocks/http/src/lib/common/auth.ts
```diff
@@ -0,0 +1,37 @@
...
+          displayName: 'Header Key',
+          required: false,
+          defaultValue: 'Authorization',
+        }),
+        value: Property.SecretText({
+          displayName: 'Header Value',
+          required: false,
+        }),
+      },
+    }),
```
rmn-snpk, line 35: Not sure, but maybe we need another field to determine whether the value is secret or not

packages/blocks/http/src/lib/actions/send-http-request-action.ts
```diff
@@ -168,11 +170,33 @@ export const httpSendRequestAction = createAction({
...
+      (
+        context.auth?.headers as
+          | Array<{ key?: string; value?: string }>
+          | undefined
+      )?.reduce<HttpHeaders>((acc, item) => {
+        if (item.key && item.value) {
+          acc[item.key] = item.value;
+        }
+        return acc;
+      }, {}) ?? {};
```
MarceloRGonc: Try to simplify this code a bit.

```
const authHeaders: HttpHeaders = {};
const headersArray = (
  context.auth?.headers as
    | Array<{ key?: string; value?: string }>
    | undefined
) ?? [];

for (const item of headersArray) {
  if (item.key && item.value) {
    authHeaders[item.key] = item.value;
  }
}
```

or

```
const headersArray = (
  context.auth?.headers as
    | Array<{ key?: string; value?: string }>
    | undefined
) ?? [];

const authHeaders = Object.fromEntries(
  headersArray
    .filter(item => item.key && item.value)
    .map(item => [item.key!, item.value!])
);
```

Even if you keep the reduce you can split it into two and it will be more readable.

packages/blocks/http/src/lib/common/auth.ts
bigfluffycookie: no rita said its okay that all the values are treated as secret, but i guess we could add a checkbox for it, not sure its super necessary though

packages/blocks/http/src/lib/common/auth.ts
```diff
@@ -0,0 +1,37 @@
...
+  required: false,
+  description,
+  props: {
+    headers: Property.Array({
+      displayName: 'Headers',
+      required: false,
+      properties: {
+        key: Property.ShortText({
+          displayName: 'Header Key',
+          required: false,
```
rita-gorokhod: why required false? the array itself is not required, so if you do add an item, you need to specify the key and value

PR #1524, opened 2025-10-23
Ensure proper webhook response when triggered via /sync

MarceloRGonc:
Fixes OPS-2780.

### DIFF
--- a/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
+++ b/packages/server/api/src/app/flows/flow-run/flow-run-service.ts
@@ -115,18 +115,19 @@ function returnHandlerId(
   }
 }
 
-function returnProgressUpdateType(
+function getEffectiveProgressUpdateType(
   pauseMetadata: PauseMetadata | undefined,
+  progressUpdateType: ProgressUpdateType,
   executionCorrelationId: string,
-): ProgressUpdateType | undefined {
+): ProgressUpdateType {
   if (
     pauseMetadata &&
     pauseMetadata.executionCorrelationId === executionCorrelationId
   ) {
-    return pauseMetadata.progressUpdateType;
+    return pauseMetadata.progressUpdateType || progressUpdateType;
   }
 
-  return undefined;
+  return progressUpdateType;
 }
 
 export const flowRunService = {
@@ -284,9 +285,11 @@ export const flowRunService = {
 
     const pauseMetadata = flowRunToResume.pauseMetadata;
 
-    const previousProgressUpdateType =
-      returnProgressUpdateType(pauseMetadata, executionCorrelationId) ||
-      progressUpdateType;
+    const progressUpdateTypeToUse = getEffectiveProgressUpdateType(
+      pauseMetadata,
+      progressUpdateType,
+      executionCorrelationId,
+    );
 
     return flowRunService.start({
       payload,
@@ -298,7 +301,7 @@ export const flowRunService = {
         executionCorrelationId,
       ),
       executionCorrelationId,
-      progressUpdateType: previousProgressUpdateType,
+      progressUpdateType: progressUpdateTypeToUse,
       executionType,
       environment: RunEnvironment.PRODUCTION,
       triggerSource: flowRunToResume.triggerSource,


packages/server/api/src/app/webhooks/webhook-controller.ts
```diff
@@ -73,132 +50,18 @@ export const webhookController: FastifyPluginAsyncTypebox = async (app) => {
...
     await reply
       .status(response.status)
       .headers(response.headers)
       .send(response.body);
   });
 };
 
-async function handleWebhook({
-  request,
-  flowId,
```
MarceloRGonc, line 91: Moved to webhook-handler

packages/server/api/src/app/webhooks/webhook-handler.ts
MarceloRGonc, line 1: Moved from webhook-controller. And split in two methods:
- handleWebhookSimulation: simulates the call it's used in the test button
- handleWebhook: the webhook call logic it can be sync or async

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -121,10 +121,11 @@ export const flowEngineWorker: FastifyPluginAsyncTypebox = async (app) => {
     if (
       progressUpdateType === ProgressUpdateType.WEBHOOK_RESPONSE &&
       workerHandlerId &&
-      executionCorrelationId
+      executionCorrelationId &&
+      isFlowStateTerminal(runDetails.status)
     ) {
       await webhookResponseWatcher.publish(
-        executionCorrelationId,
+        runId,
```
MarceloRGonc, line 128: The executionCorrelationId changes if the workflow pauses/resume. We need to send the notification using the flowRunId

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -288,19 +289,6 @@ async function getFlowResponse(
...
-          status: StatusCodes.OK,
-          body: result.pauseMetadata.response,
-          headers: {},
-        };
-      }
-      return {
-        status: StatusCodes.NO_CONTENT,
-        body: {},
-        headers: {},
-      };
```
MarceloRGonc, line 303: We don't want to notify in this case. We wait for the final workflow status

packages/server/api/src/app/webhooks/webhook-handler.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  if (async) {
+    return {
+      status: StatusCodes.OK,
+      body: {},
+      headers: {},
+    };
+  }
+
+  return webhookResponseWatcher.oneTimeListener(flowRunId.id, false);
```
MarceloRGonc, line 134: Changed to false. Because we are disabling the timeout for sync calls.

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -115,6 +115,24 @@ function returnHandlerId(
...
 
+function returnProgressUpdateType(
+  pauseMetadata: PauseMetadata | undefined,
+  executionCorrelationId: string,
+): ProgressUpdateType | undefined {
+  if (isNil(pauseMetadata)) {
+    return undefined;
+  }
+
+  if (
```
bigfluffycookie, line 123: i feel like the 2 ifs in this function can be merged

packages/server/api/src/app/workers/engine-controller.ts
```diff
@@ -338,6 +325,9 @@ async function getFlowResponse(
         body: {},
         headers: {},
       };
+    // Case that should be handled before
```
bigfluffycookie: remove

packages/server/api/src/app/flows/flow-run/flow-run-service.ts
```diff
@@ -279,7 +297,9 @@ export const flowRunService = {
         executionCorrelationId,
       ),
       executionCorrelationId,
-      progressUpdateType,
+      progressUpdateType:
```
bigfluffycookie: i feel like this reads very confusing, i'd prefer if you pass in the progressUpdateType to the function, and then figure it out internally

MarceloRGonc: I'm passing progressUpdateType to the function, but in some cases, if it's a resume, we have to get the value from pauseMetadata.

I changed the function that decides and I'm using a variable.

PR #1527, opened 2025-10-23
Fix text wrap on condition fields

rmn-snpk:
Fixes OPS-2919.

image
image

### DIFF
--- a/packages/react-ui/src/app/features/builder/step-settings/split-settings/single-condition.tsx
+++ b/packages/react-ui/src/app/features/builder/step-settings/split-settings/single-condition.tsx
@@ -112,7 +112,7 @@ const SingleCondition = ({
                   disabled={readonly}
                   initialValue={field.value ?? ''}
                   onChange={field.onChange}
-                  className={'px-1'}
+                  className="px-1"
                 ></TextInputWithMentions>
                 <FormMessage />
               </FormItem>


packages/react-ui/src/app/features/builder/block-properties/text-input-with-mentions/index.tsx
```diff
@@ -114,12 +114,12 @@ export const TextInputWithMentions = ({
...
         attributes: {
           class: cn(
-            className ??
-              ' w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring focus-visible:ring-offset-1 disabled:cursor-not-allowed disabled:opacity-50',
+            'w-full rounded-md border border-input bg-background px-3 py-2 text-sm ring-offset-background file:border-0 file:bg-transparent file:text-sm file:font-medium placeholder:text-muted-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring focus-visible:ring-offset-1 disabled:cursor-not-allowed disabled:opacity-50',
             textMentionUtils.inputThatUsesMentionClass,
             {
               'cursor-not-allowed opacity-50': disabled,
             },
+            className,
```
rmn-snpk, line 122: Custom class name was not used, so it's safe to change

PR #1529, opened 2025-10-24
Change timeout log level from debug to info

MarceloRGonc:
Part of OPS-2921.


packages/server/worker/src/lib/engine/call-engine.ts
```diff
@@ -183,7 +183,7 @@ function logEngineError(
     status = EngineResponseStatus.TIMEOUT;
     errorMessage = 'Engine execution timed out.';
 
-    logger.debug(errorMessage, {
+    logger.info(errorMessage, {
```
MarceloRGonc, line 186: It will facilitate investigation in cases where the timeout occurs earlier than expected.

maor-rozenfeld (approved):

PR #1534, opened 2025-10-27
Update schedule triggers to return start time

rmn-snpk:
Fixes OPS-2735.

image

### DIFF
--- a/packages/blocks/schedule/src/lib/triggers/every-hour.trigger.ts
+++ b/packages/blocks/schedule/src/lib/triggers/every-hour.trigger.ts
@@ -9,7 +9,7 @@ function calculateEveryHourCron(runOnWeekends: boolean) {
 }
 
 function getEveryHourData(runOnWeekends: boolean) {
-  const { cronExpression } = calculateEveryHourCron(runOnWeekends);
+  const cronExpression = calculateEveryHourCron(runOnWeekends);
   return Promise.resolve([
     {
       cron_expression: cronExpression,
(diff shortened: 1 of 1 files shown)


packages/blocks/schedule/src/lib/triggers/every-hour.trigger.ts
```diff
@@ -4,6 +4,22 @@ import {
   TriggerStrategy,
 } from '@openops/blocks-framework';
 
+function calculateEveryHourCron(runOnWeekends: boolean) {
+  const cronExpression = runOnWeekends ? `0 * * * *` : `0 * * * 1-5`;
+  return { cronExpression };
```
MarceloRGonc:
```suggestion
  return runOnWeekends ? `0 * * * *` : `0 * * * 1-5`;
```

packages/blocks/schedule/src/lib/triggers/every-x-minutes.trigger.ts
```diff
@@ -26,20 +41,19 @@ export const everyXMinutesTrigger = createTrigger({
...
   },
+  test(ctx) {
+    return getEveryXMinutesData(ctx.propsValue.minutes);
+  },
   run(ctx) {
-    const cronExpression = `*/${ctx.propsValue.minutes} * * * *`;
-    return Promise.resolve([
-      {
-        cron_expression: cronExpression,
-        timezone: 'UTC',
```
MarceloRGonc, line 40: Isn't this value being lost?

rmn-snpk: You are right, fixed

PR #1539, opened 2025-10-29
Add step index in flow service output

alexandrudanpop:
Fixes OPS-2869

Same index as shown in the UI when viewing a flow.

### DIFF
--- a/packages/shared/src/lib/flows/flow-helper.ts
+++ b/packages/shared/src/lib/flows/flow-helper.ts
@@ -1166,8 +1166,7 @@ function addStepIndices(trigger: Trigger): Trigger {
   const stepIndexMap = new Map<string, number>();
 
   const getStepKey = (step: TriggerWithOptionalId | Action): string => {
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    return `${step.id!}-${step.name}`;
+    return `${step.id}-${step.name}`;
   };
 
   allSteps.forEach((step, index) => {


alexandrudanpop:
> Why don't we save the value in the database?

I don't think we need it in the DB.

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -1161,6 +1161,32 @@ function getUsedConnections(step: Trigger | Action) {
     }, {});
 }
 
+function addStepIndices(trigger: Trigger): Trigger {
+  const allSteps = getAllSteps(trigger);
+  const stepIndexMap = new Map<string, number>();
+
+  const getStepKey = (step: TriggerWithOptionalId | Action): string => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return `${step.id!}-${step.name}`;
```
cezudas:
```suggestion
    return `${step.id}-${step.name}`;
```

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -1161,6 +1161,32 @@ function getUsedConnections(step: Trigger | Action) {
     }, {});
 }
 
+function addStepIndices(trigger: Trigger): Trigger {
+  const allSteps = getAllSteps(trigger);
+  const stepIndexMap = new Map<string, number>();
+
+  const getStepKey = (step: TriggerWithOptionalId | Action): string => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
cezudas:
```suggestion
```

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -1161,6 +1161,32 @@ function getUsedConnections(step: Trigger | Action) {
...
+    return `${step.id!}-${step.name}`;
+  };
+
+  allSteps.forEach((step, index) => {
+    stepIndexMap.set(getStepKey(step), index + 1);
+  });
+
+  return transferStep(trigger, (step: Step) => {
+    const stepIndex = stepIndexMap.get(getStepKey(step));
+    if (stepIndex) {
```
cezudas, line 1178: So this truthiness check is equivalent of
`if (stepIndex !== undefined && stepIndex > 0)`

Can we simply use `if (stepIndex !== undefined)` to keep logic simple instead?

alexandrudanpop: we don't have 0 stepIndex. The stepIndex is shown to the user in the flow builder.

PR #1541, opened 2025-10-29
Display query router reasoning in the chat

cezudas:
Part of OPS-2935.

Screenshot 2025-10-29 at 15 18 11
Screenshot 2025-10-29 at 15 18 35


packages/ui-components/src/components/assistant-ui/thread/reasoning-part.tsx
```diff
@@ -0,0 +1,26 @@
...
+type ReasoningPartProps = {
+  text: string;
+};
+
+export const ReasoningPart: FC<ReasoningPartProps> = ({ text }) => {
+  return (
+    <div className="my-2">
+      <ExpandableContent fullContent={text}>
+        {(content) => (
+          <div className="flex items-start gap-2 pt-2">
```
alexandrudanpop:
```suggestion
          <div className="flex items-center gap-2 pt-2">
```

looks better I think:
image

PR #1548, opened 2025-10-31
Add connection timeout for AI assistant chat to prevent hanging on poor network connections

cezudas:
Fixes OPS-2896.

Screenshot 2025-11-03 at 09 00 23

505117582-e9018058-4dc3-4090-b857-6b14830b0587
505117575-85fe894d-8971-4c1e-8231-f2d7149cae3b


packages/server/api/src/app/ai/chat/chat-request-router.ts
```diff
@@ -130,12 +136,12 @@ function isResponseOpen(res: ServerResponse): boolean {
...
 function startSSEHeartbeat(
   res: ServerResponse,
-  intervalMs = 15000,
+  intervalMs = SSE_HEARTBEAT_INTERVAL_MS,
 ): NodeJS.Timeout {
   const heartbeat = setInterval(() => {
     if (isResponseOpen(res)) {
       try {
-        res.write(`: heartbeat\n\n`);
+        res.write(createStreamMessage({ type: 'data-heartbeat' }));
```
cezudas: SSE comments are not triggering an update in the frontend, and the connection monitoring feature on FE should not disconnect while heartbeat messages are being sent.

alexandrudanpop:
@cezudas should we also add this issue in ai-sdk? looks like something that should be resolved in the framework

cezudas:
> @cezudas should we also add this issue in ai-sdk? looks like something that should be resolved in the framework

I just rolled back to ```res.write(`: heartbeat\n\n`)``` and added `wrapResponseForSSETracking` to the frontend, which intercepts any SSE message, including:
: heartbeat comments (every 15 seconds from server)
: connection-established comments
data: {...} actual data messages

packages/react-ui/src/app/features/ai/lib/use-connection-monitoring.ts
```diff
@@ -0,0 +1,389 @@
...
+  Timeout,
+  UseConnectionMonitoringProps,
+  UseConnectionMonitoringReturn,
+} from './types';
+
+const RESPONSE_WARNING_MS = SSE_HEARTBEAT_INTERVAL_MS * 2;
+
+const SSE_MESSAGE_GAP_TIMEOUT_MS = SSE_HEARTBEAT_INTERVAL_MS * 4;
+
+export const useConnectionMonitoring = ({
```
alexandrudanpop, line 20: @cezudas is there anything we can do to make this cleaner and easier to reason about? I see a lot of ref and useEffect, but I wonder how we can make it more maintainable in the long run. Right now it's pretty hard to read  and reason about.

packages/react-ui/src/app/features/ai/lib/use-connection-monitoring.ts
```diff
@@ -0,0 +1,389 @@
...
+  const [connectionError, setConnectionError] = useState<string | null>(null);
+
+  const lastMessageTimeRef = useRef<number>(Date.now());
+  const previousStatusRef = useRef<ChatStatusType>(chatStatus);
+  const previousMessageCountRef = useRef<number>(messages.length);
+  const warningTimerRef = useRef<Timeout | null>(null);
+  const gapMonitorTimerRef = useRef<Timeout | null>(null);
+
+  const isShowingSlowWarningRef = useRef<boolean>(isShowingSlowWarning);
+  const connectionErrorRef = useRef<string | null>(connectionError);
```
alexandrudanpop, line 35: why do we need both ref and useState here?

packages/react-ui/src/app/features/ai/lib/use-connection-monitoring.ts
alexandrudanpop: i know there might be caveats I might be missing, but maybe the code could be something like:

```jsx
export const useConnectionMonitoring = ({ chatStatus, messages, stopChat }) => {
  const [warning, setWarning] = useState<'slow' | 'error' | null>(null);
  const lastActivityRef = useRef(Date.now());
  
  // Update activity timestamp when messages arrive
  useEffect(() => {
    if (isActivelyLoading(chatStatus)) {
      lastActivityRef.current = Date.now();
      setWarning(null);
    }
  }, [chatStatus, messages.length]);
  
  // Single monitoring loop
  useEffect(() => {
    if (!isActivelyLoading(chatStatus)) {
      setWarning(null);
      return;
    }
    
    const interval = setInterval(() => {
      const elapsed = Date.now() - lastActivityRef.current;
      
      if (elapsed > SSE_MESSAGE_GAP_TIMEOUT_MS) {
        stopChat();
        setWarning('error');
        clearInterval(interval);
      } else if (elapsed > RESPONSE_WARNING_MS) {
        setWarning('slow');
      }
    }, 1000);
    
    return () => clearInterval(interval);
  }, [chatStatus, stopChat]);
  
  return { warning, clearWarning: () => setWarning(null) };
};
```

cezudas: Your proposed solution is much cleaner, but there are some caveats:

Important:

1. **SSE activity tracking**: Your version relies only on messages.length changing, but the current implementation also tracks raw SSE activity (heartbeats, partial chunks) via wrapResponseForSSETracking. This is important because:
- Server might send heartbeat comments that don't create new messages
- Large messages arrive in chunks

2. Precision timing: The interval approach can drift and be less precise. If you need to show a warning at exactly 10s, the interval might check at 9s, then 10s, then 11s.

Less important:
1.  Polling overhead: Checking every 1000ms adds unnecessary work when the connection is healthy. The current implementation only schedules checks when needed.

2. Tab visibility: Current implementation pauses monitoring when tab is hidden (battery/performance optimization).

packages/react-ui/src/app/features/ai/lib/use-connection-monitoring.ts
cezudas: We need both because of stale closures in async callbacks:
1. useState is needed for React to re-render the UI when warnings/errors change
2. useRef is needed because:
* onSSEActivity is registered globally and called from the fetch wrapper
* Timer callbacks (setTimeout) capture values at creation time
* Adding these states to dependencies would cause constant re-registration and timer recreation
The refs act as a "current value pointer" that callbacks can read without being recreated.

PR #1552, opened 2025-10-31
Query router should only append tools

alexandrudanpop:
Fixes OPS-2891


packages/server/api/src/app/ai/mcp/llm-query-router.ts
```diff
@@ -112,16 +127,21 @@ export async function routeQuery({
...
     }
 
+    const mergedToolNames = Array.from(
+      new Set([...previousTools, ...selectedToolNames]),
+    ).filter((name) => validToolNames.includes(name));
+
     const selectedTools = Object.fromEntries(
       Object.entries(tools)
-        .filter(([name]) => selectedToolNames.includes(name))
+        .filter(([name]) => mergedToolNames.includes(name))
```
cezudas, line 136: before tools would reset each turn, now odds of reaching or exceeding 128 is higher.

alexandrudanpop: We have hard limit at 128 below this line.

PR #1553, opened 2025-11-03
Fix ask AI sometimes throwing schema generation errors

bigfluffycookie:
Fixes OPS-2884

### DIFF
--- a/packages/blocks/ai/src/lib/actions/askAi.ts
+++ b/packages/blocks/ai/src/lib/actions/askAi.ts
@@ -113,6 +113,7 @@ export const askAi = createAction({
         model: languageModel,
         prompt: composedPrompt,
         schema: analysisLLMSchema,
+        maxRetries: 0,
         ...((modelSettings as Record<string, unknown>) ?? {}),
         experimental_telemetry: { isEnabled: isLLMTelemetryEnabled() },
       });


packages/blocks/ai/src/lib/actions/askAi.ts
```diff
@@ -108,13 +108,23 @@ export const askAi = createAction({
...
-    const result = await generateObject({
-      model: languageModel,
-      prompt: composedPrompt,
-      schema: analysisLLMSchema,
-      ...((modelSettings as Record<string, unknown>) ?? {}),
-      experimental_telemetry: { isEnabled: isLLMTelemetryEnabled() },
-    });
-    return result.object;
+    try {
+      const result = await generateObject({
```
rita-gorokhod, line 112: why not always use generateText?

bigfluffycookie: hmm i can, i just thought this way we could preserve the schema logic for whenever it does work

rita-gorokhod: @alexandrudanpop wdyt?

alexandrudanpop: For now that we have such a simple schema, I think it's fine.

alexandrudanpop: I think in this case we should also put `maxRetries` = 0 in `generateObject`. 
A more sophisticated way to do it would be using [repair_text](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-object#experimental_repair-text) and passing the error & the current generated text in the prompt of the next LLM telling it to fix it.

bigfluffycookie: hm but if we do that dont we risk the llm getting it wrong again? im not sure we should do it

PR #1555, opened 2025-11-03
Fix chat name generation by filtering user messages

ravikiranvm:
Fixes OPS-2963.

## Additional Notes

- We were sending the whole chat history (including user, assistant and reasoning) messages which made the LLMs misbehave in several ways like schema errors, hallucinations and outputting part of reasoning messages as chat name
- Now we only send user messages to the LLM to generate a name and it works. I've tested with Vertex (Claude and Gemini), GPT (o1, 4o, 5, 5-mini, 5-nano)

### Before and After the change with Vertex provider:

- LLM responded with a sentence from reasoning messages.

Screenshot 2025-11-03 at 6 14 23 PM
Screenshot 2025-11-03 at 6 15 05 PM

### With GPT-5 Nano (Before and after the change):

Screenshot 2025-11-03 at 6 17 11 PM
Screenshot 2025-11-03 at 6 17 55 PM

### DIFF
--- a/packages/react-ui/public/locales/en/translation.json
+++ b/packages/react-ui/public/locales/en/translation.json
@@ -414,6 +414,7 @@
   "Close sidebar": "Close sidebar",
   "Settings": "Settings",
   "Change": "Change",
+  "No Connection (Optional)": "No Connection (Optional)",
   "By OpenOps": "By OpenOps",
   "Your new workflow requires the following connections. We recommend setting them up now.": "Your new workflow requires the following connections. We recommend setting them up now.",
   "You will not be able to run the workflow until all required connections are added": "You will not be able to run the workflow until all required connections are added",
--- a/packages/react-ui/src/app/features/ai/lib/assistant-ui-chat-hook.ts
+++ b/packages/react-ui/src/app/features/ai/lib/assistant-ui-chat-hook.ts
@@ -129,10 +129,14 @@ export const useAssistantChat = ({
           stepDetails &&
           flowId
         ) {
+          const stepId =
+            flowHelper.getStep(context.flowVersion, context.selectedStep)?.id ??
+            context.selectedStep;
+
           return await aiChatApi.open(
             flowId,
             getBlockName(stepDetails),
-            context.selectedStep,
+            stepId,
             getActionName(stepDetails),
           );
         }
@@ -292,7 +296,11 @@ export const useAssistantChat = ({
 
       if (oldChatId) {
         const context = getBuilderState();
-        if (context?.selectedStep && context?.flowVersion) {
+        if (
+          context?.selectedStep &&
+          context?.flowVersion &&
+          chatMode === ChatMode.StepSettings
+        ) {
           await aiChatApi.delete(oldChatId);
           chat.setMessages([]);
         } else {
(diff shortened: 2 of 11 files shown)


packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -45,7 +45,7 @@ import { parseUserMessage } from './message-parser';
 import { createUserMessage } from './model-message-factory';
 import { getBlockSystemPrompt } from './prompts.service';
 
-const DEFAULT_CHAT_NAME = 'New Chat';
+const DEFAULT_CHAT_NAME = 'Undefined Chat';
```
ravikiranvm: I think "Undefined Chat" would be a better name when AI couldn't generate a name for any reason

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -243,7 +243,19 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
...
       }
 
-      const rawChatName = await generateChatName(chatHistory, projectId);
+      const userMessages = chatHistory.filter(
+        (msg): msg is ModelMessage =>
+          msg &&
+          typeof msg === 'object' &&
+          'role' in msg &&
+          msg.role === 'user',
+      );
```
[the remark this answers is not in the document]
ravikiranvm: Added a more specific type guard

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
bigfluffycookie: Did you check with marco? or anyone? i wouldnt rename something just out of preference usually, for all we know this was decided upon by multiple people

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -243,7 +243,19 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
...
-      const rawChatName = await generateChatName(chatHistory, projectId);
+      const userMessages = chatHistory.filter(
+        (msg): msg is UserModelMessage =>
+          msg &&
+          typeof msg === 'object' &&
+          'role' in msg &&
+          msg.role === 'user',
+      );
+
+      if (userMessages.length === 0) {
```
bigfluffycookie, line 254: Wont this throw if userMessages is undefined? Or is that the desired behavior

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
ravikiranvm: I think we should rename it, but I will check with Marco about this.

ravikiranvm: I talked to Marco and we thought to use either 'New Chat' or 'Untitled Chat'; so I reverted this change to use the 'New Chat'

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
ravikiranvm: We could throw an error in this case but I wanted to keep it consistent with the other condition where the chatHistory is empty.

`if (chatHistory.length === 0) {
        return await reply.code(200).send({ chatName: DEFAULT_CHAT_NAME });
      }`

PR #1559, opened 2025-11-04
Add AWS SSM document utils

rmn-snpk:
Part of OPS-2931.

### DIFF
--- a/packages/openops/src/lib/aws/ssm/get-ssm-describe-document-info.ts
+++ b/packages/openops/src/lib/aws/ssm/get-ssm-describe-document-info.ts
@@ -4,19 +4,16 @@ import {
   DocumentParameter,
   SSMClient,
 } from '@aws-sdk/client-ssm';
-import { BlockPropValueSchema } from '@openops/blocks-framework';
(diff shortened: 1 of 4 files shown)


packages/openops/src/lib/aws/ssm/get-ssm-describe-document-info.ts
```diff
@@ -0,0 +1,34 @@
...
+export const getSsmDescribeDocumentInfo = async ({
+  auth,
+  region,
+  runbookName,
+  version,
+}: {
+  auth: BlockPropValueSchema<typeof amazonAuth>;
+  region: string;
+  runbookName: string;
+  version?: string;
```
MarceloRGonc: Extract to a type. It's used in 3 places. Have you checked if AWS exports these types?

rmn-snpk: Types are extracted. RunbookName and version are just a string in AWS

MarceloRGonc (approved):

PR #1560, opened 2025-11-04
Add runbook name field

rmn-snpk:
Part of OPS-2931

### DIFF
--- a/packages/openops/src/lib/aws/ssm/get-ssm-describe-document-info.ts
+++ b/packages/openops/src/lib/aws/ssm/get-ssm-describe-document-info.ts
@@ -4,19 +4,16 @@ import {
   DocumentParameter,
   SSMClient,
 } from '@aws-sdk/client-ssm';
-import { BlockPropValueSchema } from '@openops/blocks-framework';
-import { amazonAuth, getCredentialsForAccount } from '../auth';
+import { getCredentialsForAccount } from '../auth';
 import { getAwsClient } from '../get-client';
+import { SsmRunbookParams } from './types';
 
 export const getSsmDescribeDocumentInfo = async ({
   auth,
   region,
   runbookName,
   version,
-}: {
-  auth: BlockPropValueSchema<typeof amazonAuth>;
-  region: string;
-  runbookName: string;
+}: SsmRunbookParams & {
   version?: string;
 }): Promise<DocumentParameter[]> => {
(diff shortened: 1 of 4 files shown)


packages/openops/src/lib/aws/ssm/runbook-name-property.ts
```diff
@@ -0,0 +1,54 @@
...
+          label: d.Name || 'Unknown',
+          value: d.Name,
+        })),
+      };
+    } catch (error) {
+      return {
+        disabled: true,
+        options: [],
+        placeholder: 'Failed to load runbooks',
+        error: String(error),
```
MarceloRGonc, line 52: I think this will be [Object Object]. Try:

```
extractPropertyString(error, ['message']) || 'Some default error',

File: packages/openops/src/lib/aws/ssm/runbook-name-property.ts [hunk 2]
```diff
@@ -0,0 +1,54 @@
...
+      const docs = await getSsmDocuments({
+        auth: awsAuth,
+        region: awsRegion,
+        owner: owner as DocumentOwner,
+      });
+
+      return {
+        disabled: false,
+        options: docs.map((d: DocumentIdentifier) => ({
+          label: d.Name || 'Unknown',
```
bigfluffycookie: if its unknown wouldnt we not want it in the list? we could potentially have multiple unknowns in the list else

packages/openops/src/lib/aws/ssm/runbook-name-property.ts
rmn-snpk: No, it's fine

image

packages/openops/src/lib/aws/ssm/runbook-name-property.ts
rmn-snpk: Fixed

packages/openops/src/lib/aws/ssm/runbook-name-property.ts
MarceloRGonc: Here’s why:
- The global String() function calls the object’s toString() method if it exists.
- For a regular Error instance, Error.prototype.toString() returns something like "Error: message".

I think it's ok to assume it will always be an Error instance

MarceloRGonc (approved):

PR #1562, opened 2025-11-04
Add document version field

rmn-snpk:
Part of OPS-2931.

### DIFF
--- a/packages/openops/src/lib/aws/ssm/runbook-version-property.ts
+++ b/packages/openops/src/lib/aws/ssm/runbook-version-property.ts
@@ -29,12 +29,14 @@ export const runbookVersionProperty = Property.Dropdown({
         runbookName,
       });
 
-      const opts = versions.map((v) => {
-        const ver = v.DocumentVersion || '';
-        const name = v.VersionName ? ` - ${v.VersionName}` : '';
-        const def = v.IsDefaultVersion ? ' (default)' : '';
-        return { label: `${ver}${name}${def}`, value: ver };
-      });
+      const opts = versions
+        .map((v) => {
+          const ver = v.DocumentVersion || '';
+          const name = v.VersionName ? ` - ${v.VersionName}` : '';
+          const def = v.IsDefaultVersion ? ' (default)' : '';
+          return { label: `${ver}${name}${def}`, value: ver };
+        })
+        .filter((option) => Boolean(option.label));
 
       return {
         disabled: false,


packages/openops/src/lib/aws/ssm/runbook-version-property.ts
```diff
@@ -0,0 +1,53 @@
...
+        disabled: false,
+        options: opts,
+        placeholder: 'Select version',
+      };
+    } catch (error) {
+      return {
+        disabled: true,
+        options: [],
+        placeholder: 'Failed to load versions',
+        error: String(error),
```
MarceloRGonc, line 51: Same as the other PR

packages/openops/src/lib/aws/ssm/runbook-version-property.ts
```diff
@@ -0,0 +1,53 @@
...
+        auth: awsAuth,
+        region: awsRegion,
+        runbookName,
+      });
+
+      const opts = versions.map((v) => {
+        const ver = v.DocumentVersion || '';
+        const name = v.VersionName ? ` - ${v.VersionName}` : '';
+        const def = v.IsDefaultVersion ? ' (default)' : '';
+        return { label: `${ver}${name}${def}`, value: ver };
```
bigfluffycookie: what if documentVersion, VersionName, isDefaultVersion are all undefined? label will be an empty string, or it could be just '(default)' is this what we want? maybe we can add logs atleast if those are missing and we allow it

packages/openops/src/lib/aws/ssm/runbook-version-property.ts
rmn-snpk: It's fine, we use it this way for all aws errors

packages/openops/src/lib/aws/ssm/runbook-version-property.ts
rmn-snpk: I  added filtering fro empty labels. In case version is empty the default will be picked by AWS

packages/openops/src/lib/aws/ssm/runbook-version-property.ts
```diff
@@ -0,0 +1,55 @@
...
+      });
+
+      const opts = versions
+        .map((v) => {
+          const ver = v.DocumentVersion || '';
+          const name = v.VersionName ? ` - ${v.VersionName}` : '';
+          const def = v.IsDefaultVersion ? ' (default)' : '';
+          return { label: `${ver}${name}${def}`, value: ver };
+        })
+        .filter((option) => Boolean(option.label));
```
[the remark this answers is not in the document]
rmn-snpk: it's fine, empty value is picked up as default

MarceloRGonc (approved):

PR #1565, opened 2025-11-05
Add props processing when calling /options

rita-gorokhod:
Fixes OPS-2973


packages/engine/src/lib/helper/block-helper.ts
```diff
@@ -110,6 +111,26 @@ export const blockHelper = {
...
+
+      const { blockAction } = await blockLoader.getBlockAndActionOrThrow({
+        blockName: params.block.blockName,
+        blockVersion: params.block.blockVersion,
+        blocksSource,
+        actionName: params.actionOrTriggerName,
+      });
+
+      const { processedInput, errors } =
+        await propsProcessor.applyProcessorsAndValidators(
```
rita-gorokhod, line 123: same pattern that exists in other places: first call propsResolver, and then call propsProcessor

https://github.com/openops-cloud/openops/blob/5c2a12ea891ece855ee1a6d1cdca34b06b524405/packages/engine/src/lib/helper/trigger-helper.ts#L65

https://github.com/openops-cloud/openops/blob/5c2a12ea891ece855ee1a6d1cdca34b06b524405/packages/engine/src/lib/handler/block-executor.ts#L139

packages/engine/src/lib/helper/block-helper.ts
```diff
@@ -110,6 +111,26 @@ export const blockHelper = {
         unresolvedInput: params.input,
         executionState,
       });
+
+      const { blockAction } = await blockLoader.getBlockAndActionOrThrow({
+        blockName: params.block.blockName,
+        blockVersion: params.block.blockVersion,
+        blocksSource,
+        actionName: params.actionOrTriggerName,
+      });
```
MarceloRGonc, line 120: Before this method, we are doing 'getPropOrThrow', which also loads the block. Maybe we should change that method to 'getPropAndActionOrThrow'

PR #1566, opened 2025-11-05
Install default azure extensions

bigfluffycookie:
Fixes OPS-2673


.devcontainer/Dockerfile
```diff
@@ -39,6 +39,15 @@ RUN <<-```
   # Install Azure CLI
   curl -sL https://aka.ms/InstallAzureCLIDeb | bash
 
+  # Configure Azure CLI extensions
```
bigfluffycookie, line 46: idk what the common extensions are, so i just picked some that seemed to be, maybe we need to add more?

PR #1578, opened 2025-11-05
Support S3 in Cloudformation

rita-gorokhod:
image
image


ai-prompts/cloudformation.txt
```diff
@@ -0,0 +1,19 @@
+You are an AWS CloudFormation Assistant with deep expertise in CloudFormation syntax for AWS.
```
rita-gorokhod, line 1: I'm debating about this, but it sucks typing it, so it does help

PR #1581, opened 2025-11-06
Show network status for AI chats based on polling

alexandrudanpop:
Fixes OPS-2969, OPS-2968

### DIFF
--- a/ai-prompts/mcp.txt
+++ b/ai-prompts/mcp.txt
@@ -3,21 +3,16 @@ OpenOps is a No-Code FinOps automation platform. It provides customizable workfl
 
 Core responsibilities:
 1. Maintain consistent context across interactions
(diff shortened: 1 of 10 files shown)


packages/react-ui/src/app/common/hooks/use-network-status.ts
```diff
@@ -0,0 +1,50 @@
+import { platformApi } from '@/app/lib/platforms-api';
+import { useEffect, useState } from 'react';
+
+type NetworkStatus = 'good' | 'slow' | 'disconnected';
```
rmn-snpk, line 4: can we replace it with enum?

alexandrudanpop: TBH, I don't see an improvement here, and someone else might comment and say they prefer union :D

PR #1600, opened 2025-11-11
Fix slack interactions by normalising emoji text

ravikiranvm:
Fixes OPS-2996.

## Additional Notes

Button text is being normalised by Slack in the response payload and this lead to the conflict to resume the flow. Now we match the normalised text response from slack with the normalised original action text and then we return the original text in the updated message. 

For multi select, we are already sending the userSelection.value which slack doesn't normalise and need not be fixed. 

## For future reference

- I've first added emoji: false prop to SlackText and although this fixes the bug, Slack wouldn't render emojis in the UI when user uses emoji in colon notation like ":emoji:"
- I've then tried normalisation in request action and then compare it with original action in wait for interaction and then send the emojified string back to the user. 
- We've decided to send back the original text (either emoji or text notation) to the user.
- So I've moved the normalisation to wait for interaction module and return the original action to the user. For multi - select (Arrays), we already return the value prop which slack doesn't normalise.
- As as alternate fix, I've tried https://github.com/openops-cloud/openops/pull/1615, but this would require user to add value prop for actions when using send message action. So we ignore this method although it would've been the cleanest approach.

### DIFF
--- a/packages/blocks/slack/test/wait-for-interaction.test.ts
+++ b/packages/blocks/slack/test/wait-for-interaction.test.ts
@@ -1,21 +1,8 @@
 const slackUpdateMessageMock = jest.fn();
 
-function normalizeEmojiStringMock(str: string): string {
-  const emojiMap: Record<string, string> = {
-    '🎲': ':game_die:',
-    '✅': ':white_check_mark:',
-  };
-  let result = str;
-  for (const [emoji, shortcode] of Object.entries(emojiMap)) {
-    result = result.replaceAll(emoji, shortcode);
-  }
-  return result;
-}
-
 jest.mock('../src/lib/common/utils', () => ({
   ...jest.requireActual('../src/lib/common/utils'),
   slackUpdateMessage: slackUpdateMessageMock,
-  normalizeEmojiString: normalizeEmojiStringMock,
 }));
 
 import { StoreScope } from '@openops/blocks-framework';
@@ -98,7 +85,7 @@ describe('wait-for-interaction', () => {
           resumePayload: {
             queryParams: {
               userName: 'test_user',
-              actionType: ':game_die: Reroll',
+              actionType: 'button',
               actionClicked: JSON.stringify({
                 value: ':game_die: Reroll',
                 displayText: ':game_die: Reroll',
@@ -169,7 +156,7 @@ describe('wait-for-interaction', () => {
           resumePayload: {
             queryParams: {
               userName: 'test_user',
-              actionType: ':game_die: Reroll',
+              actionType: 'button',
               actionClicked: JSON.stringify({
                 value: ':game_die: Reroll',
                 displayText: ':game_die: Reroll',
@@ -207,7 +194,7 @@ describe('wait-for-interaction', () => {
           resumePayload: {
             queryParams: {
               userName: 'test_user',
-              actionType: 'Approve',
+              actionType: 'button',
               actionClicked: JSON.stringify({
                 value: 'Approve',
                 displayText: 'Approve',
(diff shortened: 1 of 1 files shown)


MarceloRGonc (approved):

packages/blocks/slack/src/lib/common/wait-for-interaction.ts
```diff
@@ -115,11 +115,13 @@ export async function onReceivedInteraction(
...
+  const emojifiedAction = Array.isArray(userSelection)
+    ? userSelection.map((opt) => emojifyString(opt.value))
+    : emojifyString(userSelection.value);
+
   return {
     user: resumePayload.userName,
-    action: Array.isArray(userSelection)
-      ? userSelection.map((opt) => opt.value)
-      : userSelection.value,
+    action: emojifiedAction,
```
ravikiranvm: We need to include the emojified string to be able to compare it in the following steps in the workflow.

packages/blocks/slack/jest.config.ts
```diff
@@ -6,6 +6,6 @@ export default {
   transform: {
     '^.+\\.ts$': ['ts-jest', { tsconfig: '<rootDir>/tsconfig.spec.json' }],
   },
-  moduleFileExtensions: ['ts', 'js', 'html'],
+  moduleFileExtensions: ['ts', 'js', 'html', 'json'],
```
ravikiranvm, line 9: The node-emoji package depends on emojilib, which imports a JSON data file (emojis.json) containing emoji mappings. This extension is required to resolve these JSON module imports.

packages/blocks/slack/src/lib/common/wait-for-interaction.ts
ravikiranvm: We don't need to normalise the userSelection because Slack already does that

packages/blocks/slack/src/lib/common/wait-for-interaction.ts
MarceloRGonc: That's debatable. That will be the output of the step. It should be exactly what the user wrote on the step definition. If the user entered an emoji, it's the emoji; if it's a string, it's the string.

ravikiranvm: Without this change, the output would not contain any emojis because Slack normalises them.

packages/blocks/slack/test/wait-for-action.test.ts
```diff
@@ -1,10 +1,9 @@
 const pauseMock = jest.fn();
 const slackUpdateMessageMock = jest.fn();
-jest.mock('../src/lib/common/utils', () => {
-  return {
-    slackUpdateMessage: slackUpdateMessageMock,
-  };
-});
+jest.mock('../src/lib/common/utils', () => ({
+  ...jest.requireActual('../src/lib/common/utils'),
```
bigfluffycookie, line 4: Can we not add tests?

ravikiranvm: This was part of a fix where I tried a different approach and added tests to this file. Later I removed the tests for trying the current approach. I can add tests for the logic I've added/modified.

bigfluffycookie: please add the tests. always good to verify assumptions.

packages/blocks/slack/test/wait-for-interaction.test.ts
```diff
@@ -0,0 +1,514 @@
+const slackUpdateMessageMock = jest.fn();
+
+function normalizeEmojiStringMock(str: string): string {
```
bigfluffycookie: Why are you mocking it instead of testing the real underlying normalizeEmojiString function?

ravikiranvm: Actually tests didn't use this function, I've removed it. I wrongly added it there.
The real func from `...jest.requireActual('../src/lib/common/utils')` is used by tests.

PR #1604, opened 2025-11-12
Refactor azure sub dropdown property and add multi select dropdown

bigfluffycookie:
Part of OPS-2860

### DIFF
--- a/packages/blocks/azure/src/lib/common-properties.ts
+++ b/packages/blocks/azure/src/lib/common-properties.ts
@@ -8,6 +8,7 @@ export interface SubscriptionDropdownConfig {
   description: string;
   required: boolean;
   multiSelect: boolean;
+  preselectAll?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


packages/blocks/azure/src/lib/common-properties.ts
```diff
@@ -3,93 +3,144 @@ import { getAzureSubscriptionsStaticDropdown } from '@openops/common';
 import { runCommand } from './azure-cli';
 import { getAzureErrorMessage } from './error-helper';
 
-export const subDropdown = Property.DynamicProperties({
-  displayName: '',
+export interface SubscriptionDropdownConfig {
```
bigfluffycookie, line 6: its an attempt to avoid the duplication of 2 basically identical properties, the alternative is to just copy paste the subDropdown implementation and simply using staticmultiselect instead of staticdropdown but i like this a bit better

MarceloRGonc (approved):

PR #1618, opened 2025-11-14
Improve /all-chats performance by replacing Redis SCAN with per-user chat index

ravikiranvm:
Fixes OPS-3001.
This change will require a migration script to move all chatIds to respective user index.

### Why /all-chats was slow

- Redis SCAN is O(n) and requires multiple passes.
- It scans all keys, not only chat keys.
- Each chat key triggered a separate GET, causing N+1 calls.

### How it’s now improved

- Each user has a dedicated Redis Set storing their chat IDs.
- Lookup becomes O(1) via SMEMBERS.

### DIFF
--- a/packages/react-ui/src/app/routes/openops-tables/index.tsx
+++ b/packages/react-ui/src/app/routes/openops-tables/index.tsx
@@ -2,8 +2,6 @@ import { t } from 'i18next';
 import { useLocation } from 'react-router-dom';
 
 import { flagsHooks } from '@/app/common/hooks/flags-hooks';
-import { platformHooks } from '@/app/common/hooks/platform-hooks';
-import { projectHooks } from '@/app/common/hooks/project-hooks';
 import { useDefaultSidebarState } from '@/app/common/hooks/use-default-sidebar-state';
 import { useCandu } from '@/app/features/extensions/candu/use-candu';
 import { FlagId } from '@openops/shared';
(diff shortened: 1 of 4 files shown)


packages/server/api/src/app/ai/chat/ai-chat.service.ts
```diff
@@ -176,22 +182,26 @@ export const getAllChats = async (
...
-      continue;
-    }
-    const longChatId = keyParts[2];
+  if (chatIds.length === 0) {
+    return [];
+  }
+
+  const contexts = await Promise.all(
+    chatIds.map((chatId) => getChatContext(chatId, userId, projectId)),
+  );
```
MarceloRGonc:
```suggestion
  const chatsOrNull = await Promise.all(
    chatIds.map(async (chatId): Promise<ChatSummary | null> => {
      const context = await getChatContext(chatId, userId, projectId);

      if (!context?.chatName) {
        return null;
      }

      return {
        chatId,
        chatName: context.chatName,
      };
    }),
  );
```

packages/server/api/src/app/ai/chat/ai-chat.service.ts
```diff
@@ -176,22 +182,26 @@ export const getAllChats = async (
...
+
```
MarceloRGonc, line 206:
```suggestion
  return chatsOrNull.filter(
    (chat): chat is ChatSummary => chat !== null,
  );
```

PR #1621, opened 2025-11-17
Adjust create database token API helper

cezudas:
Fixes OPS-3027.

### DIFF
--- a/packages/server/api/src/app/openops-tables/index.ts
+++ b/packages/server/api/src/app/openops-tables/index.ts
@@ -1,6 +1,6 @@
 import { addUserToWorkspace } from './add-user-workspace';
 import { createDatabase } from './create-database';
-import { createDatabaseToken } from './create-database-token';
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/openops-tables/database-token-service.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const TOKEN_NAME_PREFIX = 'Project_';
+
+export const databaseTokenService = {
+  async generateDatabaseToken(
+    projectId: string,
+    workspaceId: number,
+  ): Promise<DatabaseToken> {
+    const { token: systemToken } =
+      await authenticateDefaultUserInOpenOpsTables();
```
MarceloRGonc: This method should receive either the token or the user and the password should not use the default user. It will depend on the flow, but it's possible that the admin may not have access to the workspace.

cezudas: done

PR #1622, opened 2025-11-17
Prevent Ai chats from step settings to appear in Ai Assistant

ravikiranvm:
Fixes OPS-3058.
Deployed to UX


packages/react-ui/src/app/features/ai/lib/assistant-ui-chat-hook.ts
```diff
@@ -232,6 +232,10 @@ export const useAssistantChat = ({
       toast(errorToast);
     },
     onFinish: async () => {
+      if (chatMode !== ChatMode.Agent) {
```
ravikiranvm, line 235: This prevents calling chat-name endpoint to generate a chat name for AI conversations made from step-settings. When a chat doesn't have a name, it wouldn't be included in the chat history list.

PR #1624, opened 2025-11-17
Add list database token API helper

MarceloRGonc:
Part of OPS-3029.

### DIFF
--- a/packages/react-ui/src/app/routes/openops-tables/index.tsx
+++ b/packages/react-ui/src/app/routes/openops-tables/index.tsx
@@ -2,8 +2,6 @@ import { t } from 'i18next';
 import { useLocation } from 'react-router-dom';
 
 import { flagsHooks } from '@/app/common/hooks/flags-hooks';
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/openops-tables/list-database-tokens.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function listDatabaseTokens(
+  workspaceId: number,
+  token: string,
+): Promise<DatabaseToken[]> {
+  const tokens = await makeOpenOpsTablesGet<DatabaseToken>(
+    'api/database/tokens/',
+    createAxiosHeaders(token),
+  );
+
+  return tokens.flat().filter((t) => t.workspace === workspaceId);
```
cezudas, line 16: For some reason the endpoint returns an artificially nested list, looks like a Baserow bug to me.

MarceloRGonc: I think it's our method. But we already do this in other places.

PR #1628, opened 2025-11-17
Set tables workspace selection cookie upon user signup/signin

cezudas:
Part of OPS-3008.

### DIFF
--- a/packages/server/api/src/app/authentication/context/authentication-cookies.ts
+++ b/packages/server/api/src/app/authentication/context/authentication-cookies.ts
@@ -15,7 +15,7 @@ export function setAuthCookiesAndReply(
   const date = jwtDecode<{ exp: number }>(response.tablesRefreshToken);
   const cookieExpiryDate = new Date(date.exp * 1000);
 
-  let replyWithCookies = reply
+  return reply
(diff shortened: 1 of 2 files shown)


packages/shared/src/lib/authentication/dto/authentication-response.ts
```diff
@@ -8,4 +8,5 @@ export type AuthenticationResponse = UserWithoutPassword & {
   projectId: string;
   projectRole: ProjectMemberRole | null;
   tablesRefreshToken: string;
+  tablesWorkspaceId?: number;
```
MarceloRGonc: Why optional?

packages/server/api/src/app/authentication/context/authentication-cookies.ts
```diff
@@ -29,8 +29,24 @@ export function setAuthCookiesAndReply(
       httpOnly: false,
       expires: cookieExpiryDate,
       sameSite: 'lax',
-    })
-    .send(response);
+    });
+
+  if (response.tablesWorkspaceId !== undefined) {
```
MarceloRGonc: The value should not be optional. We generate the tokens so we have an workspaceId

MarceloRGonc (approved):

PR #1630, opened 2025-11-18
Rename AI Assistant to match the OpenOps edition

ravikiranvm:
Fixes OPS-3091.

## UI Changes

Screenshot 2025-11-20 at 12 54 00 PM
Screenshot 2025-11-20 at 12 53 31 PM

### DIFF
--- a/packages/ui-components/src/components/assistant-ui/assistant-top-bar.tsx
+++ b/packages/ui-components/src/components/assistant-ui/assistant-top-bar.tsx
@@ -34,7 +34,7 @@ const AssistantTopBar = ({
     text: title,
     speed: 50,
     chatId: chatId,
-    isDefault: isTitleDefault,
+    enableAnimation: !isTitleDefault,
   });
 
   return (
--- a/packages/ui-components/src/hooks/use-typing-animation.ts
+++ b/packages/ui-components/src/hooks/use-typing-animation.ts
@@ -4,25 +4,25 @@ type UseTypingAnimationOptions = {
   text: string;
   speed?: number;
   chatId?: string | null;
-  isDefault?: boolean;
+  enableAnimation?: boolean;
 };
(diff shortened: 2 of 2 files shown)


packages/ui-components/src/components/assistant-ui/assistant-top-bar.tsx
```diff
@@ -29,12 +28,14 @@ const AssistantTopBar = ({
   historyContent,
   children,
   chatId,
+  defaultTitle,
 }: AssistantTopBarProps) => {
+  const baseTitle = defaultTitle ?? t('OpenOps Assistant');
```
rmn-snpk: Why do we need `defaultTitle` , `title` and `t('OpenOps Assistant');`. May be default value for `title` is enough?

ravikiranvm: You're right. The problem was that the useTypingAnimation was designed to work with a static default name for the chat. Now we have dynamic default name based on the app's edition, this animation effect didn't work. So I've now refactored the animation logic to support dynamic default name for the assistant.

packages/ui-components/src/hooks/use-typing-animation.ts
```diff
@@ -3,41 +3,30 @@ import { useEffect, useRef, useState } from 'react';
 type UseTypingAnimationOptions = {
   text: string;
   speed?: number;
-  fromText?: string;
   chatId?: string | null;
-  defaultText?: string;
+  isDefault?: boolean;
```
rmn-snpk: I'd suggest using more generic name like `enableAnimation` or `isEnabled`

ravikiranvm: going with enableAnimation

PR #1633, opened 2025-11-19
Extract basic auth endpoints to methods

MarceloRGonc:
Part of OPS-2945.


packages/server/api/src/app/authentication/authentication.controller.ts
MarceloRGonc, line 1: Just moved the code. To prevent merge conflicts

PR #1638, opened 2025-11-19
Refactor for federated login to prevent conflicts with internal

alexandrudanpop:
Fixes OPS-2947


packages/shared/src/lib/flag/flag.ts
```diff
@@ -63,4 +63,5 @@ export enum FlagId {
   OAUTH_PROXY_URL = 'OAUTH_PROXY_URL',
   CANDU_CLIENT_TOKEN = 'CANDU_CLIENT_TOKEN',
   SAMPLE_DATA_SIZE_LIMIT_KB = 'SAMPLE_DATA_SIZE_LIMIT_KB',
+  FEDERATED_LOGIN_ENABLED = 'FEDERATED_LOGIN_ENABLED',
```
MarceloRGonc, line 66: FYI in the backend we don't have this defined for this repo.

alexandrudanpop: i know, I'm only adding it to avoid conflicts, and main logic is not here

PR #1646, opened 2025-11-20
Refactor step name generation to use alphabetical indexing

rmn-snpk:
Fixes OPS-3053.

### DIFF
--- a/packages/shared/src/lib/flows/test/flow-helper.test.ts
+++ b/packages/shared/src/lib/flows/test/flow-helper.test.ts
@@ -3,15 +3,17 @@ import { flowHelper } from '../flow-helper';
 
 describe('flowHelper', () => {
   describe('indexToAlphabetical', () => {
-    it('should convert boundary and multi-letter indices correctly', () => {
-      expect(flowHelper.indexToAlphabetical(0)).toBe('a');
-      expect(flowHelper.indexToAlphabetical(25)).toBe('z');
-      expect(flowHelper.indexToAlphabetical(26)).toBe('aa');
-      expect(flowHelper.indexToAlphabetical(27)).toBe('ab');
-      expect(flowHelper.indexToAlphabetical(51)).toBe('az');
-      expect(flowHelper.indexToAlphabetical(52)).toBe('ba');
-      expect(flowHelper.indexToAlphabetical(701)).toBe('zz');
-      expect(flowHelper.indexToAlphabetical(702)).toBe('aaa');
+    it.each<[number, string]>([
+      [0, 'a'],
+      [25, 'z'],
+      [26, 'aa'],
+      [27, 'ab'],
+      [51, 'az'],
+      [52, 'ba'],
+      [701, 'zz'],
+      [702, 'aaa'],
+    ])('converts %d to %s', (input, expected) => {
+      expect(flowHelper.indexToAlphabetical(input)).toBe(expected);
     });
   });
 


packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -1096,16 +1099,44 @@ function doesActionHaveChildren(
...
+ *
+ * @param index - Zero-based integer to convert.
+ * @returns Alphabetical label corresponding to the index.
+ * @example
+ * indexToAlphabetical(0)   // "a"
+ * indexToAlphabetical(25)  // "z"
+ * indexToAlphabetical(26)  // "aa"
+ * indexToAlphabetical(701) // "zz"
+ */
+function indexToAlphabetical(index: number): string {
```
cezudas, line 1115: nitpick: indexToAlphabetical assumes a non-negative index. Consider guarding

packages/shared/src/lib/flows/flow-helper.ts
```diff
@@ -1096,16 +1099,44 @@ function doesActionHaveChildren(
   return false;
 }
 
+/**
```
cezudas, line 1102: Nitpick: indexToAlphabetical function uses a standard but mathematically subtle algorithm for bijective base-26 conversion (specifically the n-- step inside the loop).

Suggestion: add tests that exercise indexToAlphabetical and findUnusedName directly (esp. boundary values like 0/25/26 and gap scenarios)

cezudas (approved):

packages/shared/src/lib/flows/flow-helper.ts
rmn-snpk: the guard is in the loop

`while (n > 0)`

packages/shared/src/lib/flows/flow-helper.ts
rmn-snpk: thanks, done

packages/shared/src/lib/flows/test/flow-helper.test.ts
```diff
@@ -2,6 +2,49 @@ import { Action } from '../actions/action';
 import { flowHelper } from '../flow-helper';
 
 describe('flowHelper', () => {
+  describe('indexToAlphabetical', () => {
+    it('should convert boundary and multi-letter indices correctly', () => {
+      expect(flowHelper.indexToAlphabetical(0)).toBe('a');
```
cezudas: you can use `each.test` pattern, the expectations are identical, different input and outputs.

packages/shared/src/lib/flows/test/flow-helper.test.ts
```diff
@@ -2,6 +2,49 @@ import { Action } from '../actions/action';
...
+      expect(flowHelper.indexToAlphabetical(26)).toBe('aa');
+      expect(flowHelper.indexToAlphabetical(27)).toBe('ab');
+      expect(flowHelper.indexToAlphabetical(51)).toBe('az');
+      expect(flowHelper.indexToAlphabetical(52)).toBe('ba');
+      expect(flowHelper.indexToAlphabetical(701)).toBe('zz');
+      expect(flowHelper.indexToAlphabetical(702)).toBe('aaa');
+    });
+  });
+
+  describe('findUnusedName', () => {
```
cezudas, line 20: it('should return the first alphabetical suffix when none exist',
 it('should fill the first gap in the existing names'
 it('should skip unrelated names and prefixes'
 it('should work with larger gaps and mixed ordering'
 
 have identical expectations, except different inputs and outputs. you can use `each.test` pattern

rmn-snpk: I would keep them separate

packages/shared/src/lib/flows/test/flow-helper.test.ts
rmn-snpk: good point, done

PR #1647, opened 2025-11-20
Add tablesDatabaseId and tablesDatabaseToken to block context to be later used by openops-tables

cezudas:
Part of OPS-3021.

### DIFF
--- a/packages/server/worker/src/lib/api/server-api.service.ts
+++ b/packages/server/worker/src/lib/api/server-api.service.ts
@@ -19,7 +19,6 @@ import {
   GetBlockRequestQuery,
   GetFlowVersionForWorkerRequest,
   PopulatedFlow,
-  Project,
   RemoveStableJobEngineRequest,
   UpdateRunProgressRequest,
   WorkerMachineHealthcheckRequest,
--- a/packages/engine/src/lib/handler/context/engine-constants.ts
+++ b/packages/engine/src/lib/handler/context/engine-constants.ts
@@ -39,8 +39,6 @@ export class EngineConstants {
   public static readonly BLOCK_SOURCES =
     process.env.OPS_BLOCKS_SOURCE ?? 'FILE';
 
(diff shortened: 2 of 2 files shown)


packages/blocks/framework/src/lib/context.ts
```diff
@@ -125,6 +125,11 @@ export type ServerContext = {
   apiUrl: string;
   publicUrl: string;
   token: string;
+  tablesDatabaseId: number;
+  tablesDatabaseToken: {
```
bigfluffycookie: is this always going to be the same as flow EncryptedStepOutput? maybe it should be a common type?

packages/server/worker/src/lib/engine/engine-runner.ts
```diff
@@ -27,7 +27,12 @@ import {
 } from '@openops/shared';
 import chalk from 'chalk';
 
-type EngineConstants = 'publicUrl' | 'internalApiUrl' | 'engineToken';
+type EngineConstants =
```
bigfluffycookie: this and the one in "flow-job-executor" are identical, maybe have a common type somewhere?

packages/blocks/framework/src/lib/context.ts
```diff
@@ -125,6 +125,11 @@ export type ServerContext = {
...
+    iv: string;
+    data: string;
+  };
```
MarceloRGonc: You can use the Type here

packages/server/worker/src/lib/api/server-api.service.ts
```diff
@@ -19,6 +19,7 @@ import {
   GetBlockRequestQuery,
   GetFlowVersionForWorkerRequest,
   PopulatedFlow,
+  Project,
```
MarceloRGonc: Is this used?

MarceloRGonc (approved):

PR #1653, opened 2025-11-21
Ensure we don't send strings in the error property

maor-rozenfeld:
Part of OPS-2843.

### DIFF
--- a/packages/openops/src/lib/command-wrapper.ts
+++ b/packages/openops/src/lib/command-wrapper.ts
@@ -42,10 +42,10 @@ export async function executeFile(
 
 async function getResult(childProcess: ChildProcess, fullCommand: string) {
   let stdout = '';
-  let error = '';
+  let errorMessage = '';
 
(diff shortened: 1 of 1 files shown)


packages/openops/src/lib/cloud-cli-common.ts
```diff
@@ -10,10 +10,12 @@ export function handleCliError({
   command: string;
   error: unknown;
 }): never {
-  logger.error(`${provider} CLI execution failed.`, {
-    command,
-    error,
-  });
+  logger.error(
+    `${provider} CLI execution failed.`,
```
rita-gorokhod, line 14: I'm pretty sure it's not the only place where we do it

maor-rozenfeld: It's pretty rare to throw strings as errors, I'll eventually find all cases

MarceloRGonc (approved):

PR #1654, opened 2025-11-24
Fix pagination logic to always fill the first page

rmn-snpk:
Fixes OPS-3092.

https://www.loom.com/share/6113676dd8df4a3bb0ddbb5e0c88a3fb

### DIFF
--- a/packages/server/api/test/integration/helper/pagination/paginator.integration.test.ts
+++ b/packages/server/api/test/integration/helper/pagination/paginator.integration.test.ts
@@ -73,6 +73,33 @@ const TestFlowVersionEntity = new EntitySchema({
   },
 });
 
+const FOUR_RUNS_TEST_DATA = [
+  {
+    id: 'run1',
+    created: '2025-01-01 08:51:00.880',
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/helper/pagination/paginator.ts
```diff
@@ -91,8 +91,22 @@ export default class Paginator<Entity extends ObjectLiteral> {
...
     builder: SelectQueryBuilder<Entity>,
   ): Promise<PagingResult<Entity>> {
-    const entities = await this.appendPagingQuery(builder).getMany();
-    const hasMore = entities.length > this.limit;
+    let entities = await this.appendPagingQuery(builder).getMany();
+    let hasMore = entities.length > this.limit;
+
+    if (
+      this.hasBeforeCursor() &&
+      !this.hasAfterCursor() &&
```
rmn-snpk, line 99: Based on the impl 

`hasBeforeCursor` means we move to the prev page
`this.hasAfterCursor()` means we move the the next page

PR #1670, opened 2025-11-25
Move tables admin auth to the api

MarceloRGonc:
Part of OPS-3123.

- It will be removed from the shared in a new PR

## Summary by CodeRabbit

* **New Features**
  * Implemented token caching for improved performance and reliability.

* **Refactor**
  * Switched internal authentication to use admin credentials across initialization, workspace, database, and seeding operations.
  * Exposed a new admin-auth entry point for shared usage.

* **Tests**
  * Added unit tests validating admin auth and caching behavior.
  * Updated integration mocks to reflect the new admin authentication pathway.

### DIFF
--- a/packages/openops/src/index.ts
+++ b/packages/openops/src/index.ts
@@ -44,6 +44,7 @@ export * from './lib/openops-tables/fields';
 export * from './lib/openops-tables/filters';
 export * from './lib/openops-tables/openops-tables-common-properties';
 export * from './lib/openops-tables/requests-helpers';
+export * from './lib/openops-tables/reset-user-password';
 export * from './lib/openops-tables/rows';
 export * from './lib/openops-tables/tables';
 export * from './lib/openops-tables/types';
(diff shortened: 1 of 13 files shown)


packages/server/api/src/app/database/migrations/1763394159990-AddTablesTokenToProject.ts
```diff
@@ -1,5 +1,5 @@
-import { authenticateDefaultUserInOpenOpsTables } from '@openops/common';
-import { encryptUtils } from '@openops/server-shared';
+import { authenticateUserInOpenOpsTables } from '@openops/common';
```
bigfluffycookie, line 1: Are you sure we can just edit an old migration? Im mostly worried about environments where this was already ran. Can i go ahead and assume the change is equivalent to a simple function rename?

bigfluffycookie: also actually second question, why in here we use authenticateUserInOpenOpsTables instead of authenticateAdminUserInOpenOpsTables?

MarceloRGonc: We can edit it. It will only run in environments where it hasn't run before. Here we are getting the values from the environment variables, so we have email and password. The authenticateAdminUserInOpenOpsTables will change in the last PR and here the code will stay the same.

PR #1677, opened 2025-11-26
Censor more sensitive data in logs

maor-rozenfeld:
## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive information (passwords, tokens, secrets, and keys) are now automatically censored in error logs to prevent accidental exposure of confidential data.

### DIFF
--- a/packages/server/api/src/app/helper/error-handler.ts
+++ b/packages/server/api/src/app/helper/error-handler.ts
@@ -27,6 +27,19 @@ export const errorHandler = async (
       query,
     }))(_request);
 
+    // check if the body in requestSummary has a secret property like password and censor it
+    const secretProperties = ['password', 'secret', 'token', 'key'];
+    if (requestSummary.body && typeof requestSummary.body === 'object') {
+      const censoredBody = { ...requestSummary.body };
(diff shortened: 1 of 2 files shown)


packages/server/api/src/app/helper/error-handler.ts
```diff
@@ -27,6 +27,19 @@ export const errorHandler = async (
       query,
     }))(_request);
 
+    // check if the body in requestSummary has a secret property like password and censor it
+    const secretProperties = ['password', 'secret', 'token', 'key'];
```
rita-gorokhod, line 31: is it case sensitive? i.e. Password?
and it might need to be "contains", to catch also `authToken` `resetPassword` etc

packages/server/shared/src/lib/logger/log-cleaner.ts
```diff
@@ -99,3 +99,19 @@ function stringify(value: any) {
     return `Logger error - could not stringify object. ${error}`;
   }
 }
+
+function isSecretKey(key: string): boolean {
+  const secretKeywords = ["password", "token", "key", "secret", "authorization"];
```
rita-gorokhod, line 104: don't repeat

PR #1681, opened 2025-11-26
Redact sensitive information

bigfluffycookie:
Fixes OPS-3082

## Summary by CodeRabbit

* **New Features**
  * Logging now automatically redacts sensitive information (passwords, auth tokens, etc.) from all log outputs — including nested objects, stringified JSON, error messages and stack traces — while preserving existing truncation and numeric rounding behavior. Public APIs and signatures remain unchanged.

* **Tests**
  * Added tests validating redaction across various data formats, nesting levels, circular structures, and error contexts.

### DIFF
--- a/packages/server/shared/src/lib/logger/log-cleaner.ts
+++ b/packages/server/shared/src/lib/logger/log-cleaner.ts
@@ -6,64 +6,62 @@ export const maxFieldLength = 2048;
 
 const REDACTED = '[REDACTED]';
 
-const SENSITIVE_FIELDS = [
+const SENSITIVE_PATTERNS = [
   'password',
-  'newPassword',
-  'oldPassword',
-  'currentPassword',
-  'confirmPassword',
   'token',
-  'accessToken',
-  'refreshToken',
-  'apiKey',
   'secret',
-  'privateKey',
   'authorization',
+  'apikey',
+  'privatekey',
   'cookie',
-  'sessionId',
+  'session',
   'passphrase',
 ];
 
-const SENSITIVE_FIELD_PATTERNS = SENSITIVE_FIELDS.map(
-  (field) => new RegExp(String.raw`"${field}"\s*:\s*"[^"]*"`, 'gi'),
+const SENSITIVE_FIELD_PATTERNS = SENSITIVE_PATTERNS.map(
+  (pattern) =>
+    new RegExp(String.raw`"[^"]*${pattern}[^"]*"\s*:\s*"[^"]*"`, 'gi'),
 );
 
 const isSensitiveField = (key: string): boolean => {
-  return SENSITIVE_FIELDS.some(
-    (field) => field.toLowerCase() === key.toLowerCase(),
-  );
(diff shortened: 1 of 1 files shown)


packages/server/shared/src/lib/logger/log-cleaner.ts
```diff
@@ -4,6 +4,72 @@ import { ApplicationError } from '@openops/shared';
 
 export const maxFieldLength = 2048;
 
+const REDACTED = '[REDACTED]';
+
+const SENSITIVE_FIELDS = [
```
bigfluffycookie: idk i asked the ai to come up with a list, maybe its a bit much

packages/server/shared/src/lib/logger/log-cleaner.ts
```diff
@@ -4,6 +4,84 @@ import { ApplicationError } from '@openops/shared';
...
+  (field) => new RegExp(String.raw`"${field}"\s*:\s*"[^"]*"`, 'gi'),
+);
+
+const isSensitiveField = (key: string): boolean => {
+  return SENSITIVE_FIELDS.some(
+    (field) => field.toLowerCase() === key.toLowerCase(),
+  );
+};
+
+const redactSensitiveFields = (obj: any, visited = new WeakSet()): any => {
```
bigfluffycookie, line 31: never heard of weakset, apparently its more memory efficient? maybe ai comments arent so bad after all

packages/server/shared/src/lib/logger/log-cleaner.ts
maor-rozenfeld: you should redact any key that includes the word "password" so you don't need all the various variations

packages/server/shared/src/lib/logger/log-cleaner.ts
```diff
@@ -34,14 +112,18 @@ export const cleanLogEvent = (logEvent: any) => {
...
     } else if (typeof value === 'number') {
       eventData[key] = Math.round(value * 100) / 100;
     } else if (typeof value === 'object') {
-      eventData[key] = stringify(value);
+      eventData[key] = stringify(redactSensitiveFields(value));
+    } else if (typeof value === 'string') {
+      eventData[key] = redactSensitiveDataInString(truncate(value));
     } else {
       eventData[key] = truncate(value);
     }
```
maor-rozenfeld, line 128: string is pretty much the only type that is left anyway

packages/server/shared/src/lib/logger/log-cleaner.ts
maor-rozenfeld: no idea but can you make sure this function can never throw an error? we simply can't allow the logger to fail. If there's an error then the field should just show the error message.

packages/server/shared/src/lib/logger/log-cleaner.ts
bigfluffycookie: hey you never know

MarceloRGonc (approved):

PR #1691, opened 2025-11-27
Add system prompt to clarify the scope of session-sql tool

ravikiranvm:
Fixes OPS-3138

## Additional Notes

Screenshot 2025-11-27 at 3 40 36 PM

Although it's not happening most of the times, LLM once included `session-sql` tool in the tool selection output even for a simple query like `do you have access to my tables?`

I've checked the description of this tool:
Screenshot 2025-11-27 at 3 44 48 PM

The description is kinda generic and may not be clear for LLM to use it correctly in our app. So I've modified the system prompt to clarify the scope of this tool

---

Alternative fix: Do we really need this `session-sql` tool? Can we exclude it from the tools selection?

### DIFF
--- a/packages/server/api/src/app/ai/mcp/llm-query-router.ts
+++ b/packages/server/api/src/app/ai/mcp/llm-query-router.ts
@@ -207,7 +207,7 @@ const getSystemPrompt = async (
     `IMPORTANT: Tables tools should always be included in the output if the user asks a question involving those table names: ${openopsTablesNames.join(
       ', ',
     )}. ` +
-    "NOTE: The 'session-sql' tool is ONLY for querying tables created by AWS cost/billing MCP tools and performing temporary analysis. It is NOT for user's database or table related queries. " +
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/ai/mcp/llm-query-router.ts
```diff
@@ -207,6 +207,7 @@ const getSystemPrompt = async (
     `IMPORTANT: Tables tools should always be included in the output if the user asks a question involving those table names: ${openopsTablesNames.join(
       ', ',
     )}. ` +
+    "NOTE: The 'session-sql' tool is ONLY for querying tables created by AWS cost/billing MCP tools and performing temporary analysis within the session. It is NOT for user's database or table related queries. " +
```
rita-gorokhod, line 210: I don't like this @ravikiranvm .
There might be other tools in the future that cause this problem.
I think a broader solution is needed, to say that certain tools do not mix with each other.
Please consult with Eran.

ravikiranvm: Okay, I will talk to Eran.

PR #1693, opened 2025-11-27
Use a Service token for the mcp tools

MarceloRGonc:
Part of OPS-3127.


packages/server/api/src/app/blocks/base-block-module.ts
```diff
@@ -230,6 +230,9 @@ const ListCategoriesRequest = {
 };
 
 const OptionsBlockRequest = {
+  config: {
+    allowedPrincipals: ALL_PRINCIPAL_TYPES,
+  },
```
MarceloRGonc, line 235: I used the same as the other endpoints in this file.

alexandrudanpop (approved):

PR #1699, opened 2025-12-01
Make Region property optional in Build ARN action

rmn-snpk:
Fixes OPS-3129

### DIFF
--- a/packages/blocks/aws/test/arn/build-arn-action.test.ts
+++ b/packages/blocks/aws/test/arn/build-arn-action.test.ts
@@ -44,7 +44,7 @@ describe('buildArnAction', () => {
         ...jest.requireActual('@openops/blocks-framework'),
         propsValue: {
           service: 'iam',
-          region: regionVal,
+          region: regionVal as any,
           accountId: '123456789012',
           resourceId: 'user/David',
         },


packages/blocks/aws/src/lib/actions/arn/build-arn-action.ts
```diff
@@ -33,7 +33,7 @@ export const buildArnAction = createAction({
       service,
       accountId,
       resource: resourceId,
-      region,
+      region: region ?? '',
```
rmn-snpk, line 36: Made the field optional, as using dynamic property would lead to wf migrations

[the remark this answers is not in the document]
rmn-snpk: It's fine to have empty region

packages/blocks/aws/test/arn/build-arn-action.test.ts
rita-gorokhod, line 25: @rSnapkoOpenOps please add a test for when the region is missing

rmn-snpk: fixed

packages/blocks/aws/test/arn/build-arn-action.test.ts
```diff
@@ -36,4 +36,18 @@ describe('buildArnAction', () => {
       'arn:aws:ec2:us-west-2:123456789012:i-1234567890abcdef0',
     );
   });
+
+  it('should build ARN without region for global services (e.g., IAM)', async () => {
+    const result = await buildArnAction.run({
+      ...jest.requireActual('@openops/blocks-framework'),
+      propsValue: {
+        service: 'iam',
+        region: undefined,
```
rita-gorokhod: please add a test-case for undefined null and ''

rmn-snpk: done

rita-gorokhod: thanks!

PR #1705, opened 2025-12-02
Check tables mcp endpoints list for length

bigfluffycookie:
Part of OPS-3176


packages/server/api/src/app/ai/mcp/tables-tools.ts
```diff
@@ -8,7 +9,7 @@ import { MCPTool } from './types';
 export async function getTablesTools(): Promise<MCPTool> {
   const { token } = await authenticateAdminUserInOpenOpsTables();
   const mcpEndpoint = await openopsTables.getMcpEndpointList(token);
-  if (!mcpEndpoint) {
+  if (isEmpty(mcpEndpoint)) {
```
rita-gorokhod: @bigfluffycookie please rename it to mcpEndpoints (plural) -- it's not clear from the var name that it's a list

PR #1708, opened 2025-12-02
Add env variable to not redact in environments where its set to false

bigfluffycookie:
Fixes OPS-3175


rita-gorokhod:
@bigfluffycookie why a new var and not check if it's local env?

bigfluffycookie:
> @bigfluffycookie why a new var and not check if it's local env?

I consulted with @maor-rozenfeld and he said its better to create a new one that conveys the purpose instead of relying on one that we already have. The only one we have is OPS_ENVIRONMENT and I agree that its not great

rita-gorokhod (approved):

maor-rozenfeld (approved):

PR #1710, opened 2025-12-02
Refactor basic authentication utils for better code reusability

cezudas:
Part of OPS-3009.

### DIFF
--- a/packages/server/api/src/app/authentication/utils.ts
+++ b/packages/server/api/src/app/authentication/utils.ts
@@ -58,9 +58,9 @@ export const assertPasswordMatches = async ({
   }
 };
 
-export async function buildAuthResponse(
+export function buildAuthResponse(
   projectContext: ProjectContext,
-): Promise<AuthenticationResponse> {
+): AuthenticationResponse {
   const userWithoutPassword = removePasswordPropFromUser(projectContext.user);
 
   return {


packages/server/api/src/app/authentication/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export async function authResponse(
+  user: User,
+  tablesRefreshToken: string,
+  getProjectAndToken: (
+    user: User,
+    tablesRefreshToken: string,
+  ) => Promise<ProjectContext>,
+): Promise<AuthenticationResponse> {
+  const projectContext = await getProjectAndToken(user, tablesRefreshToken);
```
MarceloRGonc: If you're moving this for reuse purposes, why receive this method as a parameter? We only need to receive the ProjectContext

cezudas: alright, done

packages/server/api/src/app/authentication/utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  AuthenticationResponse,
+  ErrorCode,
+  isNil,
+  User,
+  UserStatus,
+} from '@openops/shared';
+import { passwordHasher } from './basic/password-hasher';
+import { AssertPasswordsMatchParams, ProjectContext } from './types';
+
+export const removePasswordPropFromUser = (
```
MarceloRGonc: Is it used outside?

cezudas: it wasn't used, i removed the export

PR #1719, opened 2025-12-03
Fix chat name generation

rmn-snpk:
Fixes OPS-3173.

Deployed on UX

The idea is to let LLM decided whether it has enough context or no. In case AI decides that name is generated we set it as chat name. In case  there is no enough data, we wait for the next LLM response and try to generate the name once again 

## Summary by CodeRabbit

* **New Features**
  * Chat name generation now includes an `isGenerated` flag to indicate whether a name was successfully generated or a default was used.

* **Bug Fixes**
  * Improved chat name generation reliability with enhanced validation and structured output handling.

* **Chores**
  * Updated chat name API response format from `chatName` to include `name` and `isGenerated` fields for better clarity on generation status.


rmn-snpk:
@coderabbitai review

packages/server/api/src/app/ai/chat/ai-mcp-chat.controller.ts
```diff
@@ -247,27 +247,24 @@ export const aiMCPChatController: FastifyPluginAsyncTypebox = async (app) => {
...
       }
 
-      const userMessages = chatHistory.filter(
-        (msg): msg is UserModelMessage =>
-          msg &&
-          typeof msg === 'object' &&
-          'role' in msg &&
-          msg.role === 'user',
-      );
+      const generated = await generateChatName(chatHistory, projectId);
```
rmn-snpk, line 256: The filtering is removed, having only user messages for generation is not enough

packages/server/api/src/app/ai/chat/ai-chat.service.ts
```diff
@@ -84,28 +90,39 @@ export const generateChatIdForMCP = (params: {
...
+
+  const sanitizedMessages: ModelMessage[] =
+    sanitizeMessagesForChatName(messages);
+
+  const result = await generateObject({
     model: languageModel,
-    messages: prompt,
+    system: systemPrompt,
+    messages: sanitizedMessages,
+    schema: generatedChatNameSchema,
```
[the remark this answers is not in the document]
rmn-snpk: fixed

packages/server/api/src/app/ai/chat/ai-chat.service.ts
```diff
@@ -84,28 +91,43 @@ export const generateChatIdForMCP = (params: {
...
-  const response = await generateText({
+
+  const sanitizedMessages: ModelMessage[] =
+    sanitizeMessagesForChatName(messages);
+
+  if (isEmpty(sanitizedMessages)) {
+    return { name: null, isGenerated: false };
+  }
+
+  const result = await generateObject({
```
alexandrudanpop: As LLMs don't always generate correct structured output, you can make it a bit more reliable using the approach below 
https://github.com/openops-cloud/openops/blob/9b6afa3e511f947fc3178c74ae288b3274ed1f9e/packages/server/api/src/app/ai/mcp/llm-query-router.ts#L120

alexandrudanpop: Also, you could wrap in try catch and return {isGenerated: false} in case there is an error and log the error

packages/server/api/src/app/ai/chat/ai-chat.service.ts
```diff
@@ -84,28 +91,43 @@ export const generateChatIdForMCP = (params: {
   });
 };
 
+const generatedChatNameSchema = z.object({
+  name: z
+    .string()
+    .max(100)
+    .nullable()
+    .describe('Conversation name or null if it was not generated'),
+  isGenerated: z.boolean().describe('Whether the name was generated or not'),
```
rita-gorokhod, line 102: @rSnapkoOpenOps why do you need isGenerated? 
you can deduce that from checking if (name). The less logic the LLM needs to do, the better.

rmn-snpk: Yes, I've tried that, but ai still adds dummy names

cezudas (approved):

PR #1720, opened 2025-12-03
Refactor seedTemplateTablesService for better separation of responsibilities

cezudas:
Part of OPS-3009.

### DIFF
--- a/packages/react-ui/src/app/features/ai/lib/enrich-context.ts
+++ b/packages/react-ui/src/app/features/ai/lib/enrich-context.ts
@@ -23,11 +23,9 @@ export const createAdditionalContext = (
     flowVersionId: flowVersion.id,
     runId,
     currentStepId: stepData?.id ?? '',
-    currentStepName: stepData?.name ?? '',
     steps: [
       {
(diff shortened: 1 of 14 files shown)


packages/server/api/src/app/database/seeds/get-admin-token-and-database.ts
```diff
@@ -0,0 +1,34 @@
...
+    throw new Error(`Admin user not found for email: ${email}`);
+  }
+
+  const project = await projectService.getOneForUser(user);
+
+  if (!project) {
+    throw new Error(`No project found for user: ${email}`);
+  }
+
+  return project;
```
bigfluffycookie, line 24: doesnt it need to be  return project.tablesDatabaseId?

cezudas: this return stmt is for getAdminUserProject helper.
We used to have `getProjectTablesDatabaseId`  helper (in `packages/server/api/src/app/openops-tables/template-tables/seed-tables-for-templates.ts`).

PR #1721, opened 2025-12-03
Log request validation errors at debug level

ravikiranvm:
Fixes OPS-2954.

### DIFF
--- a/packages/server/api/src/app/helper/error-handler.ts
+++ b/packages/server/api/src/app/helper/error-handler.ts
@@ -1,5 +1,6 @@
 import { logger } from '@openops/server-shared';
 import { ApplicationError, ErrorCode } from '@openops/shared';
+import type { ErrorObject } from 'ajv';
 import { FastifyError, FastifyReply, FastifyRequest } from 'fastify';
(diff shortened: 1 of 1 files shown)


packages/server/api/src/app/helper/error-handler.ts
```diff
@@ -53,6 +63,17 @@ const isApplicationError = (error: unknown): error is ApplicationError => {
   }
 };
 
+type FastifyValidationError = FastifyError & { validation?: ErrorObject[] };
+
+const isValidationError = (error: unknown): error is FastifyValidationError => {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
```
MarceloRGonc, line 71:
```suggestion
    isObject(error) &&
```

ravikiranvm: I've tried this;
I think `isObject` doesn't work for Error instances.

MarceloRGonc: You are right it fails on this validation:

```
Object.getPrototypeOf(error) === Object.prototype
```

PR #1739, opened 2025-12-05
Add fixed salt for admin password

MarceloRGonc:
Part of OPS-3201

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -2,7 +2,7 @@ name: openopsdev
 services:
   tables:
     container_name: tables
-    image: public.ecr.aws/openops/openops-tables:0.2.12
+    image: public.ecr.aws/openops/openops-tables:0.2.13
     environment:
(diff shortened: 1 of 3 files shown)


packages/server/shared/src/lib/system/system.ts
```diff
@@ -100,6 +100,7 @@ const systemPropDefaultValues: Partial<Record<SystemProp, string>> = {
   [AppSystemProp.TELEMETRY_MODE]: 'COLLECTOR',
   [AppSystemProp.TELEMETRY_COLLECTOR_URL]: 'https://telemetry.openops.com/save',
   [SharedSystemProp.ENABLE_HOST_VALIDATION]: 'true',
+  [AppSystemProp.OPENOPS_ADMIN_PASSWORD_SALT]: '$2b$10$6zuoB5d8Dz9bzV91gpuynO',
```
MarceloRGonc, line 103: Same default value as in the Tables repository

PR #1847, opened 2026-01-16
Update dependabot.yml to allow only security vulnerabilities

rita-gorokhod:
Fixes OPS-3425


maor-rozenfeld (changes requested):
If you don't want non-security updates by Dependabot, you can simply delete this file. However we will then be stuck with ancient package versions and it will be a nightmare to upgrade when the time comes.

PR #1961, opened 2026-02-18
Resolve conditional steps in benchmark wizard

ravikiranvm:
Part of OPS-3650.

### DIFF
--- a/packages/server/api/src/app/benchmark/provider-adapter.ts
+++ b/packages/server/api/src/app/benchmark/provider-adapter.ts
@@ -19,7 +19,12 @@ export type WizardStepOptionsSource =
 
 export type WizardStepConditional = {
   when: string;
-  skipToStep: string;
+  onSuccess?: {
+    optionsSource?: WizardStepOptionsSource;
+  };
+  onFailure?: {
+    skipToStep?: string;
+  };
 };
 
 export type WizardConfigStep = {
@@ -43,6 +48,7 @@ export type ProviderAdapter = {
     method: string,
     context: WizardContext,
   ): Promise<BenchmarkWizardOption[]>;
+  evaluateConditional(when: string, context: WizardContext): Promise<boolean>;
 };
 
 const providers = new Map<string, ProviderAdapter>();
--- a/packages/server/api/src/app/benchmark/providers/aws/aws-conditional-step-resolver.ts
+++ b/packages/server/api/src/app/benchmark/providers/aws/aws-conditional-step-resolver.ts
@@ -1,0 +1,19 @@
+import { throwValidationError } from '../../errors';
+import type { WizardContext } from '../../provider-adapter';
+
+export async function evaluateConditional(
+  when: string,
+  context: WizardContext,
+): Promise<boolean> {
+  switch (when) {
+    case 'hasMultipleAccounts':
+      return hasMultipleAccounts(context);
+    default:
+      throwValidationError(`Unknown AWS conditional method: ${when}`);
+  }
+}
+
+async function hasMultipleAccounts(_context: WizardContext): Promise<boolean> {
+  // TODO: Evaluate from connection (e.g. roles.length > 1) and return false when no connection in context. For now always show the accounts step.
+  return true;
+}
--- a/packages/server/api/src/app/benchmark/providers/aws/aws.json
+++ b/packages/server/api/src/app/benchmark/providers/aws/aws.json
@@ -13,12 +13,20 @@
       "id": "accounts",
       "title": "Which accounts should we include in the report?",
       "selectionType": "multi-select",
-      "optionsSource": { "type": "dynamic", "method": "getConnectionAccounts" },
+      "optionsSource": "",
+      "nextStep": "regions",
       "conditional": {
-        "when": "connection.supportsMultiAccount",
-        "skipToStep": "regions"
-      },
-      "nextStep": "regions"
+        "when": "hasMultipleAccounts",
+        "onSuccess": {
+          "optionsSource": {
+            "type": "dynamic",
+            "method": "getConnectionAccounts"
+          }
+        },
+        "onFailure": {
+          "skipToStep": "regions"
+        }
+      }
     },
     {
       "id": "regions",
--- a/packages/server/api/src/app/benchmark/wizard.service.ts
+++ b/packages/server/api/src/app/benchmark/wizard.service.ts
@@ -10,6 +10,8 @@ import {
   type StaticOptionValue,
   type WizardConfig,
   type WizardConfigStep,
+  type WizardContext,
+  type WizardStepOptionsSource,
 } from './provider-adapter';
 import './register-providers';
 
@@ -17,14 +19,16 @@ function getStepProgress(
   config: WizardConfig,
   stepToReturn: WizardConfigStep,
 ): { totalSteps: number; stepIndex: number } {
-  const stepsWithOptions = config.steps.filter((s) => s.optionsSource);
+  const stepsWithOptions = config.steps.filter(
+    (s) => s.optionsSource || s.conditional?.onSuccess?.optionsSource,
+  );
   const totalSteps = stepsWithOptions.length;
   const stepIndex =
     stepsWithOptions.findIndex((s) => s.id === stepToReturn.id) + 1;
   return { totalSteps, stepIndex };
 }
 
-function resolveNextStepId(
+function getNextStepIdFromConfig(
   step: WizardConfigStep,
   config: WizardConfig,
 ): string | null {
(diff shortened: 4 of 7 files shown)


packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -55,22 +88,34 @@ function computeWizardStepResponse(
...
     } else {
-      const nextStepDef = steps.find((s) => s.id === nextStepId);
+      const resolvedNextStepId = await resolveNextStepId(
+        nextStepIdFromConfig,
+        config,
+        providerAdapter,
+        context,
+      );
+      const nextStepDef = steps.find((s) => s.id === resolvedNextStepId);
       if (!nextStepDef) {
```
bigfluffycookie, line 103: you already throw in getNextStepIdFromConfig

packages/server/api/src/app/benchmark/providers/aws/aws.json
```diff
@@ -13,12 +13,20 @@
       "id": "accounts",
       "title": "Which accounts should we include in the report?",
       "selectionType": "multi-select",
-      "optionsSource": { "type": "dynamic", "method": "getConnectionAccounts" },
+      "optionsSource": "",
```
bigfluffycookie, line 16: remove this, you should only have it on steps without the conditionals

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -36,16 +40,45 @@ function resolveNextStepId(
...
 }
 
-function computeWizardStepResponse(
+async function resolveNextStepId(
+  nextStepIdFromConfig: string | null,
+  config: WizardConfig,
+  providerAdapter: ProviderAdapter,
+  context: WizardContext,
+): Promise<string | null> {
+  if (!nextStepIdFromConfig) {
```
bigfluffycookie, line 52: why even allow to add in no next stepId? just make   nextStepIdFromConfig: string mandatory

packages/server/api/src/app/benchmark/providers/aws/aws-conditional-step-resolver.ts
```diff
@@ -0,0 +1,19 @@
...
+  switch (when) {
+    case 'hasMultipleAccounts':
+      return hasMultipleAccounts(context);
+    default:
+      throwValidationError(`Unknown AWS conditional method: ${when}`);
+  }
+}
+
+async function hasMultipleAccounts(_context: WizardContext): Promise<boolean> {
+  // TODO: Evaluate from connection (e.g. roles.length > 1) and return false when no connection in context. For now always show the accounts step.
```
ravikiranvm, line 17: Will implement this in the follow up PR

packages/server/api/src/app/benchmark/provider-adapter.ts
```diff
@@ -43,6 +48,7 @@ export type ProviderAdapter = {
     method: string,
     context: WizardContext,
   ): Promise<BenchmarkWizardOption[]>;
+  evaluateConditional(when: string, context: WizardContext): Promise<boolean>;
```
ravikiranvm, line 51: im indecisive between `evaluateConditional` and `evaluateConditionalStep`

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -36,16 +40,45 @@ function resolveNextStepId(
...
+    context,
+  );
+  if (shouldShowStep) {
+    return nextStepIdFromConfig;
+  }
+  const skipToStepId = step.conditional.onFailure?.skipToStep;
+  if (!skipToStepId) {
+    return nextStepIdFromConfig;
+  }
+  return resolveNextStepId(skipToStepId, config, providerAdapter, context);
```
ravikiranvm, line 73: This is recursive because what if the resolved step is a conditional step? 
But I don't think it is a common scenario; I can change it after inputs.

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -87,21 +132,17 @@ function staticValuesToOptions(
 async function resolveOptions(
   providerAdapter: ProviderAdapter,
   step: WizardConfigStep,
-  request: BenchmarkWizardRequest,
-  projectId: string,
+  context: WizardContext,
```
ravikiranvm, line 135: instead we build the context and pass it as a whole

packages/server/api/test/unit/benchmark/providers/aws/aws-conditional-step-resolver.test.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('evaluateConditional', () => {
+  const projectId = 'project-123';
+  const provider = 'aws';
+
+  it('returns true for hasMultipleAccounts', async () => {
+    const result = await evaluateConditional('hasMultipleAccounts', {
+      projectId,
+      provider,
+    });
+    // Returns true until we implement real evaluation (see hasMultipleAccounts in aws-conditional-step-resolver).
```
ravikiranvm, line 12: Will modify it once the real evaluation functions is available in the next PR

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -55,22 +88,34 @@ function computeWizardStepResponse(
...
       }
       stepToShow = nextStepDef;
     }
   } else {
     stepToShow = steps[0];
   }
 
-  const nextStep = resolveNextStepId(stepToShow, config);
+  const nextStepIdFromConfig = getNextStepIdFromConfig(stepToShow, config);
+  const nextStep = await resolveNextStepId(
```
bigfluffycookie, line 113: This is now wrong.

Imagine this:
User is on step1 and clicks next
you look up step1s nextStep which is →  step2

step2 has a conditional → evaluates it → condition is FALSE → step2 should not be shown, you recursively check if step3 should be shown (this is correct in theory but honestly the recursion will just confuse us all more)
stepToShow= step3
nextStepId = step4

step2 has a conditional → evaluates it →  condition is TRUE →  step2 should be shown with its options
stepToShow = step2
nextStepId = step3

BUT what happens now is you go in resolveNextStepId() with step3.id, and you go through the whole loop. no matter whats the case if step2 conditon evaluates to false it means you HAVE to skip to the next step.

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -113,19 +154,21 @@ export async function resolveWizardNavigation(
...
   const providerAdapter = getProvider(normalizedProvider);
   const config = providerAdapter.config;
+  const context = {
+    benchmarkConfiguration: request.benchmarkConfiguration,
+    projectId,
+    provider: config.provider,
+  };
 
-  const { stepToShow, nextStep } = computeWizardStepResponse(
+  const { stepToShow, nextStep } = await computeWizardStepResponse(
```
bigfluffycookie, line 163: im sorry i missed it earlier, but it led me to be confused in some places now.

in here nextStep seems to be nextStepID
whereas stepToShow is an actual step.

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -36,16 +40,45 @@ function resolveNextStepId(
...
+    return nextStepIdFromConfig;
+  }
+  const skipToStepId = step.conditional.onFailure?.skipToStep;
+  if (!skipToStepId) {
+    return nextStepIdFromConfig;
+  }
+  return resolveNextStepId(skipToStepId, config, providerAdapter, context);
+}
+
+async function computeWizardStepResponse(
```
bigfluffycookie, line 76: i think the problem is that we are doing 2 things here. 

1. we navigate to the next step
2. we look ahead at whats the next step id for the response to the frontend

in one case nextStep is the next step to navigate to and then later the next step is the nextsteps.nextstep to return

packages/server/api/src/app/benchmark/wizard.service.ts
```diff
@@ -36,16 +40,45 @@ function resolveNextStepId(
...
+async function resolveNextStepId(
+  nextStepIdFromConfig: string | null,
+  config: WizardConfig,
+  providerAdapter: ProviderAdapter,
+  context: WizardContext,
+): Promise<string | null> {
+  if (!nextStepIdFromConfig) {
+    return null;
+  }
+  const step = config.steps.find((s) => s.id === nextStepIdFromConfig);
```
bigfluffycookie, line 55: this is a bit funny, its needed bc of the recursion probably but in essence what you do is 

1. you get a stepId
2. you call getNextStepIdFromConfig - which gets the steps[stepId]
3. you throw if there is no step
4. you return the stepId
5. you call this function with the ID to then get the step AGAIN from steps
6. you throw again if there is no step

bigfluffycookie (dismissed):

bigfluffycookie (commented):
Wrong approve

packages/server/api/src/app/benchmark/provider-adapter.ts
```diff
@@ -19,7 +19,12 @@ export type WizardStepOptionsSource =
 
 export type WizardStepConditional = {
   when: string;
-  skipToStep: string;
+  onSuccess?: {
+    optionsSource?: WizardStepOptionsSource;
+  };
+  onFailure?: {
+    skipToStep?: string;
+  };
```
MarceloRGonc, line 27: Aren't these fields mandatory?

packages/server/api/src/app/benchmark/provider-adapter.ts
MarceloRGonc: evaluateCondition or resolveCondition

packages/server/api/src/app/benchmark/providers/aws/aws.json
ravikiranvm: I think we decided keep it empty string for conditional steps? But I can remove it

packages/server/api/src/app/benchmark/wizard.service.ts
ravikiranvm: yeah, here nextStep is the ID. I'm using nextStep here to match the usage in json file. I will see if I can make is more clear to read.

packages/server/api/src/app/benchmark/wizard.service.ts
ravikiranvm: if the condition is TRUE, we show the step. 
I've named the method as `hasMultipleAccounts` and the success path would be to resolve options for the step.

packages/server/api/src/app/benchmark/providers/aws/aws.json
bigfluffycookie: no, for conditional steps you have it inside the condition, for non conditional steps you have it outside

packages/server/api/src/app/benchmark/wizard.service.ts
bigfluffycookie: you are right, i wrote my comment the wrong we around, i edited to fix it. The point I made is still correct though

Like maybe write a test case with 4 steps, where step 2 evaluates to TRUE and step 3 evaluates to FALSE. See what you get back from this method matches what you expect.

I suspect you get {stepToShow: step2, nextStepId: step4 }

ravikiranvm:
Opened another PR that implements the same ticket https://github.com/openops-cloud/openops/pull/1986

PR #2050, opened 2026-03-03
Fix benchmark cleanup webhook 500 on non-local environments

cezudas:
Fixes OPS-3829.

### DIFF
--- a/packages/server/api/src/app/benchmark/attach-benchmark-flows.service.ts
+++ b/packages/server/api/src/app/benchmark/attach-benchmark-flows.service.ts
@@ -1,3 +1,4 @@
+import { networkUtls } from '@openops/server-shared';
 import {
   BenchmarkProviders,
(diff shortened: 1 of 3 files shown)


packages/server/api/src/app/benchmark/attach-benchmark-flows.service.ts
```diff
@@ -38,7 +38,7 @@ async function buildPayloadForWebhook(params: {
     );
   }
 
-  const webhookBaseUrl = await webhookUtils.getWebhookPrefix();
+  const webhookBaseUrl = `${networkUtls.getInternalApiUrl()}v1/webhooks`;
```
cezudas, line 41: Use the internal URL so the engine (running inside the cluster) can reach the API server directly without going through the external nginx ingress, which fails due to hairpin NAT on non-local environments.

PR #2207, opened 2026-04-08
feat: add MiniMax provider support

octo-patch:
## Summary

- Add MiniMax as a new AI provider using the OpenAI-compatible API
- Add `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` models to the provider's model list
- Register the new provider in `AiProviderEnum` and `PROVIDER_MAP`
- Add unit tests for the new provider

## Implementation Details

MiniMax exposes an OpenAI-compatible chat API at `https://api.minimax.io/v1`. This PR implements a dedicated MiniMax provider using `@ai-sdk/openai-compatible` (already a project dependency), following the same pattern as other providers in this codebase.

Users can configure MiniMax by setting:
- Provider: `MiniMax`
- API Key: `MINIMAX_API_KEY`
- Model: `MiniMax-M2.7` (flagship) or `MiniMax-M2.7-highspeed` (faster variant)

## API Reference

- Chat (OpenAI Compatible): https://platform.minimax.io/docs/api-reference/text-openai-api

## Test plan

- [x] Unit tests updated and passing (`packages/openops/test/ai/providers.test.ts`)
- [x] Integration test verified — real API call to `api.minimax.io` returns valid response
- [x] Provider registered in `AiProviderEnum` and `PROVIDER_MAP`
- [x] Models list includes `MiniMax-M2.7` and `MiniMax-M2.7-highspeed`


alexandrudanpop:
Thanks for the contribution, @octo-patch !

I can see this issue when testing it in the main chat, you can try a prompt like: 
`What are my available tables`?

Looks related to generating JSON output from our query router. Could you please check?

```
 AI SDK Warning (minimax.chat / MiniMax-M2.7): The feature "responseFormat" is not supported. JSON response format schema is only supported with structuredOutputs
[API]
[API] [07:43:23 UTC] ERROR: Failed to generate chat name
[API]     errorStack: "AI_NoObjectGeneratedError: No object generated: could not parse the response.\n    at parseAndValidateObjectResult (/openops/node_modules/ai/src/generate-object/parse-and-validate-object-result.ts:33:11)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at parseAndValidateObjectResultWithRepair (/openops/node_modules/ai/src/generate-object/parse-and-validate-object-result.ts:88:12)\n    at fn (/openops/node_modules/ai/src/generate-object/generate-object.ts:432:24)\n    at /openops/node_modules/ai/src/telemetry/record-span.ts:32:24\n    at generateObject (/openops/node_modules/ai/src/generate-object/generate-object.ts:264:12)"
[API]     errorMessage: "No object generated: could not parse the response."
[API]     errorName: "AI_NoObjectGeneratedError"
[API]     errorContext: "{\"cause\":{\"name\":\"AI_JSONParseError\",\"cause\":{},\"text\":\"<think>\nThe user is asking me to generate a chat name based on the conversation. Let me analyze what the conversation was about:\n\n1. User said \\"hi\\" - just a greeting\n2. User asked \\"what are my available tables?\\" - asking about OpenOps tables\n3. I responded asking for clarification\n4. User clarified \\"openops tables\\"\n5. I explained I don't have direct access to query OpenOps internal tables\n6. User asked \\"what tools can you access\\"\n7. I had an error (billing related)\n8. Now user wants me to generate a chat name\n\nThe main topic seems to be about:\n- Exploring OpenOps capabilities\n- Understanding available tables and tools in OpenOps\n- User trying to learn what they can do/access in OpenOps\n\nA good name should be specific and capture the essence of this conversation. Something like:\n- \\"OpenOps capabilities inquiry\\"\n- \\"OpenOps tables and tools\\"\n- \\"Exploring OpenOps access\\"\n- \\"OpenOps features question\\"\n- \\"Available OpenOps tools\\"\n\nI think \\"OpenOps features and tables\\" or \\"OpenOps capabilities inquiry\\" would work well. Let me keep it under 5 words.\n\n\\"OpenOps features and tools\\" - 4 words\n\\"OpenOps tables and tools\\" - 4 words\n\\"OpenOps capabilities question\\" - 3 words\n\\"Exploring OpenOps access\\" - 3 words\n\nI think \\"OpenOps tables and tools\\" captures the specific questions asked well.\n</think>\n\n{\n  \\"isGenerated\\": true,\n  \\"name\\": \\"OpenOps tables and tools\\"\n}\"},\"text\":\"<think>\nThe user is asking me to generate a chat name based on the conversation. Let me analyze what the conversation was about:\n\n1. User said \\"hi\\" - just a greeting\n2. User asked \\"what are my available tables?\\" - asking about OpenOps tables\n3. I responded asking for clarification\n4. User clarified \\"openops tables\\"\n5. I explained I don't have direct access to query OpenOps internal tables\n6. User asked \\"what tools can you access\\"\n7. I had an error (billing related)\n8. Now user wants me to generate a chat name\n\n
```

octo-patch:
Sorry for the delay here, @alexandrudanpop! Sharing what I found and where I think this should land.

**Root cause.** `MiniMax-M2.7` is a thinking model — every reply starts with a `<think>...</think>` block before the actual content. MiniMax's OpenAI-compatible `/v1/chat/completions` endpoint also does not honor `response_format` (per [MiniMax-AI/MiniMax-M2.5#4](https://github.com/MiniMax-AI/MiniMax-M2.5/issues/4), which still applies to M2.7), so the AI SDK warns `"responseFormat" is not supported` and falls back to manual JSON parsing — which then trips on the leading `<think>` block. The chat-name `generateObject` call you're hitting is exactly that path.

**Why I do not want to flip `supportsStructuredOutputs: true`.** That flag tells the SDK "this model returns conformant JSON," but the underlying endpoint will still ignore `response_format` and still leak `<think>` content. It would silence the warning without fixing the parse error.

**What I think the right fix is, in order of preference:**

1. **Strip the `<think>` block before JSON parsing** for minimax models. The SDK's `transformRequestBody` doesn't help here (it's request-side); we'd need to wrap the model with a custom `extractReasoningMiddleware` (built into `ai` v6) tagged with `<think>`, which moves the reasoning out of the text channel so `generateObject` only sees the JSON tail.
2. **For object-generation paths specifically**, prefer `MiniMax-M2.7-highspeed` — it does not emit `<think>` and produces clean JSON-friendly output.

Approach 1 is the more robust fix and keeps both models usable for both `generateText` and `generateObject`. Happy to roll that into this PR (small `wrapLanguageModel(... extractReasoningMiddleware({ tagName: 'think' }))` in `minimax.ts` plus a unit test), or split it into a follow-up if you'd rather keep this PR focused on "add provider, ship, iterate." Let me know which you prefer and I'll push within a day.

alexandrudanpop:
> Sorry for the delay here, @alexandrudanpop! Sharing what I found and where I think this should land.
> 
> **Root cause.** `MiniMax-M2.7` is a thinking model — every reply starts with a `<think>...</think>` block before the actual content. MiniMax's OpenAI-compatible `/v1/chat/completions` endpoint also does not honor `response_format` (per [MiniMax-AI/MiniMax-M2.5#4](https://github.com/MiniMax-AI/MiniMax-M2.5/issues/4), which still applies to M2.7), so the AI SDK warns `"responseFormat" is not supported` and falls back to manual JSON parsing — which then trips on the leading `<think>` block. The chat-name `generateObject` call you're hitting is exactly that path.
> 
> **Why I do not want to flip `supportsStructuredOutputs: true`.** That flag tells the SDK "this model returns conformant JSON," but the underlying endpoint will still ignore `response_format` and still leak `<think>` content. It would silence the warning without fixing the parse error.
> 
> **What I think the right fix is, in order of preference:**
> 
> 1. **Strip the `<think>` block before JSON parsing** for minimax models. The SDK's `transformRequestBody` doesn't help here (it's request-side); we'd need to wrap the model with a custom `extractReasoningMiddleware` (built into `ai` v6) tagged with `<think>`, which moves the reasoning out of the text channel so `generateObject` only sees the JSON tail.
> 2. **For object-generation paths specifically**, prefer `MiniMax-M2.7-highspeed` — it does not emit `<think>` and produces clean JSON-friendly output.
> 
> Approach 1 is the more robust fix and keeps both models usable for both `generateText` and `generateObject`. Happy to roll that into this PR (small `wrapLanguageModel(... extractReasoningMiddleware({ tagName: 'think' }))` in `minimax.ts` plus a unit test), or split it into a follow-up if you'd rather keep this PR focused on "add provider, ship, iterate." Let me know which you prefer and I'll push within a day.

The plan sounds ok for me, but we would need to make it in a generic way (both 1 and 2). For point 2 for example currently we don't have a special setting, so the generateObject call happens with  the same model. Not sure if it's a good idea to hardcode the secondary model, as it will get stale as we have new model releases. So I would propose to have it in the AI settings configuration something like secondary model with short explanation what it does (generate chat name, used for code generation block, etc.).

@octo-patch  happy to assist if you need more information how to adapt the PR

octo-patch:
Thanks for the steer @alexandrudanpop, picking this back up before it gets auto-closed.

Just pushed [a75b7a4](https://github.com/octo-patch/openops/commit/a75b7a47c4) which addresses **point 1 (extractReasoningMiddleware)** in a generic, provider-agnostic way:

- Wrap every model returned from `getAiProviderLanguageModel` with `extractReasoningMiddleware({ tagName: 'think' })` from `ai` v6.
- Applied unconditionally — for models that do not emit `<think>` it is a no-op, so there is no per-provider branching to maintain.
- This fixes the chat-name `generateObject` parse error you reported (and the same class of error in `llm-query-router.ts`, which uses the same JSON-parsing path).
- Unit tests in `packages/openops/test/ai/providers.test.ts` updated to assert the wrap is applied with the expected middleware.

For **point 2 (secondary model in AI settings)**, I'd love a quick steer before I push it:

The clean way I see is: add `secondaryModel` + `secondaryCustomModel` form fields to `aiAuth` in `packages/openops/src/lib/ai/ai-auth.ts`, pass an optional `useSecondaryModel` flag through `getLLMConfig` in `ai-chat.service.ts`, and have the two utility call sites (`generateChatName` and `routeQuery` in `llm-query-router.ts`) opt into it. The frontend side mostly comes along for free because `aiAuth` is rendered from the declarative props.

A couple of open questions:

1. **Backwards compatibility** — leave `secondaryModel` optional and fall back to the primary model when unset, so existing connections keep working untouched. OK with that default?
2. **Scope of utility tasks** — only `generateChatName` and `routeQuery` for now? `code.service.ts` and `ai-step.ts` are user-facing generation paths so I assume they should keep using the primary model.
3. **PR boundary** — happy to land it in this same PR if you prefer one round of review, or split into a follow-up so #2207 stays scoped to the original "add MiniMax provider" + the generic reasoning fix. Whichever helps you ship faster.

Once you call it I'll push within a day.

MarceloRGonc:
> Thanks for the steer @alexandrudanpop, picking this back up before it gets auto-closed.
> 
> Just pushed [a75b7a4](https://github.com/octo-patch/openops/commit/a75b7a47c4) which addresses **point 1 (extractReasoningMiddleware)** in a generic, provider-agnostic way:
> 
> * Wrap every model returned from `getAiProviderLanguageModel` with `extractReasoningMiddleware({ tagName: 'think' })` from `ai` v6.
> * Applied unconditionally — for models that do not emit `<think>` it is a no-op, so there is no per-provider branching to maintain.
> * This fixes the chat-name `generateObject` parse error you reported (and the same class of error in `llm-query-router.ts`, which uses the same JSON-parsing path).
> * Unit tests in `packages/openops/test/ai/providers.test.ts` updated to assert the wrap is applied with the expected middleware.
> 
> For **point 2 (secondary model in AI settings)**, I'd love a quick steer before I push it:
> 
> The clean way I see is: add `secondaryModel` + `secondaryCustomModel` form fields to `aiAuth` in `packages/openops/src/lib/ai/ai-auth.ts`, pass an optional `useSecondaryModel` flag through `getLLMConfig` in `ai-chat.service.ts`, and have the two utility call sites (`generateChatName` and `routeQuery` in `llm-query-router.ts`) opt into it. The frontend side mostly comes along for free because `aiAuth` is rendered from the declarative props.
> 
> A couple of open questions:
> 
> 1. **Backwards compatibility** — leave `secondaryModel` optional and fall back to the primary model when unset, so existing connections keep working untouched. OK with that default?
> 2. **Scope of utility tasks** — only `generateChatName` and `routeQuery` for now? `code.service.ts` and `ai-step.ts` are user-facing generation paths so I assume they should keep using the primary model.
> 3. **PR boundary** — happy to land it in this same PR if you prefer one round of review, or split into a follow-up so [feat: add MiniMax provider support #2207](https://github.com/openops-cloud/openops/pull/2207) stays scoped to the original "add MiniMax provider" + the generic reasoning fix. Whichever helps you ship faster.
> 
> Once you call it I'll push within a day.

Hi,

Sorry for the slow reply on our end — and thanks for picking this back up.

**Point 1:** I was doing more of an overall test rather than testing against that specific model, since I don’t currently have access to it.

One thing to get the build green: in ai v6, wrapLanguageModel expects a concrete LanguageModelV3, but our provider factories — and the AiProvider interface — currently return LanguageModel, which is effectively the LanguageModelV2 | LanguageModelV3 union. That union is not assignable there.

I think the cleanest fix is to type createLanguageModel as LanguageModelV3 from @ai-sdk/provider across the providers and in the interface, instead of casting at the call site. That should clear the build without introducing an as cast around the wrapper.

In general though, the approach looks good to me.

**Point 2:**  Honestly, I’m not sure we need this yet. It feels like it adds some complexity around when and why each model gets used. But to answer your questions:

Backwards compatibility: Yes, keep it optional and fall back to the primary model when unset.
Scope: generateChatName makes sense. I’m less sure about routeQuery: routing probably works better when the same, or at least very similar, model that runs the task also makes the routing call. A different secondary model might route in ways that don’t line up with the primary model’s behavior. Let’s hold off on that one for now.
PR boundary: Separate follow-up, please. Let’s land this PR with MiniMax support and the generic reasoning fix first, then handle the secondary model separately.

Thanks again — once point 1 builds cleanly, I think we’re good to merge it. 🚀

PR #2221, opened 2026-04-15
Add new System Job and allow using jobId for in memory jobs

bigfluffycookie:
Part of OPS-4023

### DIFF
--- a/packages/server/api/src/app/helper/system-jobs/common.ts
+++ b/packages/server/api/src/app/helper/system-jobs/common.ts
@@ -27,7 +27,7 @@ type IssuesReminderSystemJobData = {
 
 type CampaignCompletionSystemJobData = {
   campaignId: string;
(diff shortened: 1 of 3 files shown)


packages/server/api/src/app/helper/system-jobs/common.ts
```diff
@@ -70,5 +77,6 @@ type UpsertJobParams<T extends SystemJobName> = {
 export type SystemJobSchedule = {
   init(): Promise<void>;
   upsertJob<T extends SystemJobName>(params: UpsertJobParams<T>): Promise<void>;
+  removeJob(jobId: string): Promise<void>;
```
bigfluffycookie, line 80: upsert doesnt actually upsert, if the job already exists it skips it so we need a way of removing it.

MarceloRGonc (approved):

PR #2224, opened 2026-04-15
Implement sorting functionality for  project tables

rmnsnpk:
Fixes OPS-4153

image


alexandrudanpop:
@rmnsnpk i think for flows and runs, sorting should not be enabled in the homepage.

packages/shared/src/lib/flow-run/dto/list-flow-runs-request.ts
```diff
@@ -3,6 +3,18 @@ import { OpenOpsId } from '../../common/id-generator';
...
+export enum FlowRunSortBy {
+  FLOW_NAME = 'flowName',
+  STATUS = 'status',
+  TRIGGER_SOURCE = 'triggerSource',
+  CREATED = 'created',
+}
+
+export enum FlowRunSortDirection {
+  ASC = 'asc',
+  DESC = 'desc',
```
alexandrudanpop: this enum can be extracted, we have same enum for others (AppConnectionSortDirection, FlowSortDirection)

packages/ui-components/src/ui/data-table.tsx
```diff
@@ -151,8 +160,30 @@ export function DataTable<
...
   const [searchParams, setSearchParams] = useSearchParams();
-  const startingCursor = searchParams.get('cursor') || undefined;
-  const startingLimit = searchParams.get('limit') || '10';
+  const startingCursor = syncWithSearchParams
+    ? searchParams.get('cursor') || undefined
+    : undefined;
+  const startingLimit =
+    syncWithSearchParams && searchParams.get('limit')
+      ? searchParams.get('limit') || '10'
+      : '10';
```
alexandrudanpop: would be cool to extract constant

packages/ui-components/src/ui/data-table.tsx
```diff
@@ -230,9 +268,13 @@ export function DataTable<
...
     columns,
     manualPagination: true,
+    enableSorting,
     getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel(),
+    onSortingChange: setSorting,
     state: {
       columnVisibility,
+      sorting,
     },
```
alexandrudanpop, line 305: I think we need to set `manualSorting: true`

https://tanstack.com/table/v8/docs/guide/sorting#manual-server-side-sorting

alexandrudanpop:
@rmnsnpk I think it looks really great, left a few mostly estetic comments

packages/shared/src/lib/flow-run/dto/list-flow-runs-request.ts
rmnsnpk: fixed

packages/ui-components/src/ui/data-table.tsx
rmnsnpk: fixed

packages/ui-components/src/ui/data-table.tsx
rmnsnpk: Done

alexandrudanpop (approved):

PR #2236, opened 2026-04-17
Update dependencies for CI audit fix

alexandrudanpop:
Fixes OPS-4155

Deployed to UX


package.json
```diff
@@ -212,7 +212,7 @@
     "isolated-vm": "5.0.1",
     "js-yaml": "4.1.0",
     "jsdom": "23.0.1",
-    "jshint": "2.13.6",
+    "jshint": "0.5.9",
```
MarceloRGonc: We need to investigate why this version is decreasing.

alexandrudanpop: It was not used. From my checks, it had a lodash dependency with a security issue. Anyway, I removed it.

MarceloRGonc (approved):

PR #2250, opened 2026-04-23
Add NestedMultiSelect component

bigfluffycookie:
Fixes OPS-4178


packages/blocks/cloudability/src/index.ts
```diff
@@ -6,6 +6,9 @@ import { snoozeRecommendationAction } from './lib/actions/snooze-recommendations
 import { unsnoozeRecommendationAction } from './lib/actions/unsnooze-recommendations-action';
 import { cloudabilityAuth } from './lib/auth';
 
+export { getRecommendationTypesByVendor } from './lib/common/common-properties';
```
bigfluffycookie: my previous export didnt work the way i wanted it to so we need to do this

packages/ui-components/src/components/nested-multi-select/nested-multi-select.tsx
```diff
@@ -0,0 +1,175 @@
+import * as React from 'react';
```
bigfluffycookie, line 1: reusable component for nested multi select in the wizard.

Image

packages/ui-components/src/components/nested-multi-select/nested-multi-select.tsx
```diff
@@ -0,0 +1,200 @@
...
+      onValueChange(newValue);
+    } else {
+      onValueChange({
+        ...value,
+        [groupId]: allItemIds,
+      });
+    }
+  };
+
+  const handleItemToggle = (groupId: string, itemId: string) => {
```
alexandrudanpop: wrap functions with useCallback

packages/blocks/cloudability/src/index.ts
alexandrudanpop: are we sure we want to import in the server from the blocks & not export this to somewhere common instead?

alexandrudanpop (approved):

PR #2254, opened 2026-04-27
Fix Redshift endpoint in Cloudability block

ravikiranvm:
Fixes OPS-4203.

### DIFF
--- a/packages/blocks/cloudability/src/lib/common/common-properties.ts
+++ b/packages/blocks/cloudability/src/lib/common/common-properties.ts
@@ -1,6 +1,6 @@
 import { Property } from '@openops/blocks-framework';
 import { Vendor, getRecommendationTypesByVendor } from '@openops/common';
-import { Duration } from './recommendations-api';
+import { CostBasis, Duration } from './recommendations-api';
 
 export function getVendorsProperty() {
   return {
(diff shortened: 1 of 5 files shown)


packages/blocks/cloudability/src/lib/common/common-properties.ts
```diff
@@ -86,3 +86,41 @@ export function getRecommendationDurationProperty() {
...
+            options: [],
+            placeholder: 'Select a vendor and recommendation type first',
+          };
+        }
+
+        if (vendor === Vendor.AWS && recommendationType === 'redshift') {
+          return {
+            disabled: true,
+            options: [],
+            placeholder: 'Cost basis is not available for AWS Redshift',
```
ravikiranvm, line 111: I think `disabling` this field for Redshift is okay with this placeholder instead of hiding this field entirely.

bigfluffycookie (approved):
looks good, btw for the other actions in cloud ability did you check if its an issue there?

ravikiranvm:
> looks good, btw for the other actions in cloud ability did you check if its an issue there?

I've checked the `snooze` endpoints and the docs didn't mention of a different API route for Redshift. But do you suggest I check those actions manually?

PR #2255, opened 2026-04-28
Sort Benchmark providers based on selectability

ravikiranvm:
Part of OPS-4221

### Items will be sorted in this order:
- First priority is given to providers that are enabled and connections are available.
- Second priority is to providers that are enabled but no connections are saved yet.
- Then everything else

Once this is merged, I will re-use the helper function for templates selection step in Campaigns.

Before this fix:
Screenshot 2026-04-28 at 2 34 15 PM

After:
Screenshot 2026-04-28 at 2 33 54 PM

### DIFF
--- a/packages/react-ui/src/app/lib/provider-sorting.ts
+++ b/packages/react-ui/src/app/lib/provider-sorting.ts
@@ -1,0 +1,29 @@
+type SortBySelectabilityOptions<T> = {
+  isSelectable: (item: T) => boolean;
+  isConnected: (item: T) => boolean;
+};
+
+export function sortBySelectability<T>(
+  items: readonly T[],
+  { isSelectable, isConnected }: SortBySelectabilityOptions<T>,
+): T[] {
+  return items
+    .map((item, index) => ({
+      item,
+      index,
+      selectable: isSelectable(item),
+      connected: isConnected(item),
+    }))
+    .sort((a, b) => {
+      if (a.selectable !== b.selectable) {
+        return a.selectable ? -1 : 1;
+      }
+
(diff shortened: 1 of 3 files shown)


packages/react-ui/src/app/lib/provider-sorting.ts
```diff
@@ -0,0 +1,29 @@
...
+): T[] {
+  return items
+    .map((item, index) => ({
+      item,
+      index,
+      selectable: isSelectable(item),
+      connected: isConnected(item),
+    }))
+    .sort((a, b) => {
+      if (a.selectable !== b.selectable) {
```
ravikiranvm, line 18: Selectable = Templates available for the provider (or) Benchmark is enabled for this provider
First we sort on `selectability`

packages/react-ui/src/app/lib/provider-sorting.ts
```diff
@@ -0,0 +1,29 @@
...
+        return a.selectable ? -1 : 1;
+      }
+
+      if (a.selectable && a.connected !== b.connected) {
```
ravikiranvm, line 22: From the selectable providers, we check if a provider has active connection. Then we show them at the top of the list.

packages/react-ui/src/app/lib/provider-sorting.ts
```diff
@@ -0,0 +1,29 @@
...
+        return a.connected ? -1 : 1;
+      }
+
+      return a.index - b.index;
```
ravikiranvm, line 26: If there any tie breaker, we keep the original order from the static list.

bigfluffycookie (approved):

PR #2259, opened 2026-05-01
Move service now functions to shared

bigfluffycookie:
Part of OPS-4218


packages/openops/src/lib/servicenow/get-tables.ts
```diff
@@ -1,33 +1,38 @@
...
   name: string;
   label: string;
-}
+};
 
 export async function getServiceNowTables(
   auth: ServiceNowAuth,
+  search?: string,
 ): Promise<ServiceNowTable[]> {
+  const queryParams: Record<string, string> = {
```
bigfluffycookie: allows us to directly search query servicenow if it table wasnt in the first 10k results.

packages/blocks/servicenow/src/lib/get-tables.ts
```diff
@@ -1,33 +1,38 @@
...
-        username: auth.username,
-        password: auth.password,
-      }),
+      ...generateAuthHeader(auth),
       Accept: 'application/json',
     },
-    queryParams: {
-      sysparm_fields: 'name,label',
-      sysparm_limit: '10000',
-      sysparm_query: 'nameSTARTSWITHx_',
```
bigfluffycookie, line 29: we didnt need this

packages/openops/src/lib/servicenow/get-tables.ts
```diff
@@ -0,0 +1,43 @@
...
+} from './auth';
+
+export type ServiceNowTable = {
+  name: string;
+  label: string;
+};
+
+export async function getServiceNowTables(
+  auth: ServiceNowAuth,
+  options?: { search?: string; query?: string; limit?: number },
```
bigfluffycookie, line 15: for blocks we wanna keep the behavior as it was before, but for campaigns we need to be able to do a search with the table name so i extended it

MarceloRGonc (approved):

PR #2261, opened 2026-05-04
Add support for assuming AWS role via Azure VM managed identity

MarceloRGonc:
Part of OPS-4254.


packages/openops/src/lib/aws/get-client.ts
bigfluffycookie, line 34: We will only hit the third else if when AWS_ENABLE_IMPLICIT_ROLE is true AND AWS_USE_AZURE_MANAGED_IDENTITY is true. Is this the intended behavior?

packages/openops/src/lib/aws/auth.ts
```diff
@@ -345,6 +349,14 @@ For large or complex setups, enhanced features are available, including:
       return fieldValidation;
     }
 
+    const hasCredentials = auth.accessKeyId && auth.secretAccessKey;
```
bigfluffycookie: The validation order looks different here vs in the get-client, which means it could pass when saving the connection but fail when running the block.

Validation says: "You need Azure enabled to skip providing credentials"

Runtime says: "You just need implicit role enabled to skip providing credentials"

packages/openops/src/lib/aws/get-client.ts
MarceloRGonc: Yes

packages/openops/src/lib/aws/auth.ts
MarceloRGonc: Removed

bigfluffycookie (approved):

PR #2263, opened 2026-05-05
Add flows bulk actions

rmnsnpk:
it is possible to delete the workflow in case it's enabled, so I think we should keep this behaviour for bulk operations 

image

Fixes OPS-4321


packages/server/api/src/app/flows/flow/flow.service.ts
```diff
@@ -453,6 +453,60 @@ export const flowService = {
...
+    await Promise.all(
+      flowIds.map((id) =>
+        this.delete({
+          id,
+          projectId,
+          userId,
+        }),
+      ),
+    );
+  },
```
alexandrudanpop, line 482: would need validation:
- load all requested flows for the project first, verify the count matches the requested IDs (this checks the flows are in the respective project)
- no flow is internal (function could be accessed from other places)
- no flow is enabled (toast to explain in the UI)

rmnsnpk: We allow user to delete single workflow when it's enabled. Are you sure we want to disable it for bulk?

packages/react-ui/src/app/routes/flows/index.tsx
```diff
@@ -71,7 +115,130 @@ const FlowsPage = () => {
...
+  const completeBulkAction = useCallback(
+    (resetSelection: () => void) => {
+      resetSelection();
+      resetSelectedRows();
+      onTableRefresh();
+    },
+    [onTableRefresh, resetSelectedRows],
+  );
+
+  const moveSelectedFlows = useCallback(
```
alexandrudanpop, line 136: move to Uncategoriezed does not work, I think we could remove that option (it does not exist also on flow level in the 3dots menu)

rmnsnpk: thanks! it's fixed

packages/server/api/src/app/flows/flow/flow.service.ts
alexandrudanpop: I would add it for sure to prevent mistakes.

rmnsnpk: Sure, done

image

alexandrudanpop (approved):

PR #2264, opened 2026-05-06
Modify `Custom Azure API Call` action to retry on 429 errors

ravikiranvm:
Fixes OPS-4255.

## Additional Notes

Here's what we are doing:
- Some Azure endpoints are throttling for the calls we make from Azure Benchmark flows.
- So we do need to catch some retry headers from Azure API calls and then wait for the duration mentioned in the headers and then retry. 
- So we do not throw on 429 errors and then honour the wait time.


packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -1,15 +1,30 @@
...
+  HttpHeaders,
+  HttpRequest,
+} from '@openops/blocks-common';
 import { Property } from '@openops/blocks-framework';
 import { azureAuth, getUseHostSessionProperty } from '@openops/common';
 import { getAzureAccessToken } from '../auth/get-azure-access-token';
 import { getSubscriptionsDropdownForHostSession } from '../common-properties';
 
+const DEFAULT_RETRY_DELAY_MS = 60000;
+const MAX_RETRY_ATTEMPTS = 4;
```
ravikiranvm, line 14: If we get 429 repeatedly even after delay; we limit the consecutive retries to 4.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -1,15 +1,30 @@
...
+const COST_MANAGEMENT_RETRY_AFTER_HEADER_PATTERN =
```
ravikiranvm, line 15: Cost Management endpoint returned different types of retry headers patterns like this. So we're using regex.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -1,15 +1,30 @@
...
+  /^x-ms-ratelimit-microsoft\.costmanagement.*-retry-after$/i;
+const AZURE_RETRY_AFTER_HEADERS = new Set([
```
ravikiranvm, line 17: These are the headers we noticed while testing by Dariusz

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -54,4 +69,100 @@ export const customAzureApiCallAction = createCustomApiCallAction({
       'Content-Type': 'application/json',
     };
   },
+  requestHandler: sendWithRetry,
 });
+
+async function sendWithRetry(
```
ravikiranvm, line 75: Instead of standard http request we use this handler function with retry logic. I've extended `createCustomApiCallAction` to use this handler

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -54,4 +69,100 @@ export const customAzureApiCallAction = createCustomApiCallAction({
...
+
+  const retryDelaysMs = retryValues
+    .map((value) => parseRetryDelayMs(value))
+    .filter((value): value is number => value !== null);
+
+  if (retryDelaysMs.length === 0) {
+    return null;
+  }
+
+  return Math.max(...retryDelaysMs);
```
ravikiranvm, line 139: If a response has many retry headers (entity, subscription, tenant) we pick the max delay time and use it.

packages/blocks/common/src/lib/http/core/http-error.ts
```diff
@@ -33,6 +34,7 @@ export class HttpError extends Error {
   get response() {
     return {
       status: this._err?.response?.status || 500,
+      headers: (this._err?.response?.headers as HttpHeaders | undefined) ?? {},
```
ravikiranvm, line 37: Without this in http response, we cannot fix this issue.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -54,4 +69,100 @@ export const customAzureApiCallAction = createCustomApiCallAction({
...
+  throw new Error(`Failed to send Azure API request to ${request.url}`);
+}
+
+export function getRetryDelayMs(
+  headers: HttpHeaders | undefined,
+  attempt: number,
+): number {
+  const retryDelayMs = getHeaderRetryDelayMs(headers);
+  if (retryDelayMs === null) {
+    return Math.pow(2, attempt - 1) * DEFAULT_RETRY_DELAY_MS;
```
ravikiranvm, line 103: if no retry-headers are matched but the error is 429, we wait for 60 secs and retry

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -54,4 +69,100 @@ export const customAzureApiCallAction = createCustomApiCallAction({
...
+
+      if (attempt === MAX_RETRY_ATTEMPTS) {
+        throw error;
+      }
+
+      await sleep(getRetryDelayMs(error.response.headers, attempt));
+    }
+  }
+
+  throw new Error(`Failed to send Azure API request to ${request.url}`);
```
bigfluffycookie, line 94: do we ever reach this code?

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -54,4 +69,100 @@ export const customAzureApiCallAction = createCustomApiCallAction({
...
+}
+
+function parseRetryDelayMs(headerValue: string): number | null {
+  const trimmedValue = headerValue.trim();
+  if (trimmedValue.length === 0) {
+    return null;
+  }
+
+  const retryAfterSeconds = Number(trimmedValue);
+  if (!Number.isFinite(retryAfterSeconds)) {
```
bigfluffycookie, line 159: maybe lets also guard against negative numbers and the number 0

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -54,4 +69,100 @@ export const customAzureApiCallAction = createCustomApiCallAction({
...
+    return null;
+  }
+
+  return retryAfterSeconds * 1000;
```
bigfluffycookie, line 163: maybe we should cap it to a limit? i mean we are bound to the retry value we receive, probably should have a max of 10 minutes.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
bigfluffycookie: hm, but i think the math here is off? the first time you wait 60 seconds, the second 120, the third 240 and the forth time its up to 480 seconds.

is this on purpose? if yes document it atleast in the ticket as there it says specifically wait for 60 seconds.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -1,15 +1,30 @@
...
+  HttpError,
+  HttpHeaders,
+  HttpRequest,
+} from '@openops/blocks-common';
 import { Property } from '@openops/blocks-framework';
 import { azureAuth, getUseHostSessionProperty } from '@openops/common';
 import { getAzureAccessToken } from '../auth/get-azure-access-token';
 import { getSubscriptionsDropdownForHostSession } from '../common-properties';
 
+const DEFAULT_RETRY_DELAY_MS = 60000;
```
bigfluffycookie, line 13: What happens when the user enables retry on failure on the action? do we do 4*4 retries?

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
```diff
@@ -1,15 +1,30 @@
-import { createCustomApiCallAction } from '@openops/blocks-common';
+import {
```
bigfluffycookie, line 1: what about the other azure actions, is it not a problem there?

bigfluffycookie (changes requested):

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
ravikiranvm: Yes, this is on purpose. On consecutive 429 failures, we wait more exponentially. I will document it.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
ravikiranvm: Sounds good

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
ravikiranvm: Yes, in the worst case it would be 16 retries if the user enables `Retry on failure` toggle. The worst case would be when the 4 consecutive 429 retries failed.

bigfluffycookie: and this is by design? as in we want this behavior?

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
ravikiranvm: There are two Azure actions that depend on CLI and I think it wouldn't be a cleaner way to retry in CLI where we might not have headers like for an HTTP call. Also currently we're not using Azure graph query action in Benchmark. Marco suggested that we document this retry on 429 behaviour somewhere for future reference. 

I'll check with Marco if I should also modify Graph Query action to retry on 429 errors.

packages/blocks/azure/src/lib/actions/custom-azure-api-action.ts
ravikiranvm: Maybe I will check about this also with Marco

PR #2273, opened 2026-05-08
Update Claude / Agent MD and add React skill

alexandrudanpop:
Fixes OPS-4263


AGENTS.md
```diff
@@ -7,15 +7,25 @@
 - Update documentation for user-facing changes
```
alexandrudanpop, line 7: modiefied agents.md to be the same as claude.md

MarceloRGonc (approved):

PR #2280, opened 2026-05-11
Add 429 retry behaviour to engine's error handler

ravikiranvm:
Fixes OPS-4255.

## Additional Notes

- Some Azure endpoints are throttling for the calls we make from Azure Benchmark flows.
- We do need to catch some retry headers from Azure API calls and then wait for the duration mentioned in the headers and then retry.
- So we do not throw on 429 errors and then honour the wait time.

### DIFF
--- a/packages/blocks/common/src/lib/http/core/azure-retry-delay.ts
+++ b/packages/blocks/common/src/lib/http/core/azure-retry-delay.ts
@@ -1,0 +1,67 @@
+import { HttpHeaders } from './http-headers';
+
+const MAX_RETRY_DELAY_MS = 10 * 60 * 1000;
+const COST_MANAGEMENT_RETRY_AFTER_HEADER_PATTERN =
+  /^x-ms-ratelimit-microsoft\.costmanagement.*-retry-after$/i;
+
+const AZURE_RETRY_AFTER_HEADERS = new Set([
+  'x-ms-ratelimit-microsoft.consumption-retry-after',
+  'x-ms-ratelimit-retailprices-retry-after',
+]);
+
+export function getAzureRetryDelayMs(
+  headers: HttpHeaders | undefined,
+): number | undefined {
+  if (!headers) {
+    return undefined;
+  }
+
+  const retryValues: string[] = [];
+
+  for (const [headerName, headerValue] of Object.entries(headers)) {
+    if (
+      AZURE_RETRY_AFTER_HEADERS.has(headerName.toLowerCase()) ||
+      COST_MANAGEMENT_RETRY_AFTER_HEADER_PATTERN.test(headerName)
+    ) {
+      retryValues.push(...normalizeHeaderValues(headerValue));
+    }
+  }
+
+  if (retryValues.length === 0) {
+    return undefined;
+  }
+
+  const retryDelaysMs = retryValues
+    .map((value) => parseRetryDelayMs(value))
+    .filter((value): value is number => value !== undefined);
+
+  if (retryDelaysMs.length === 0) {
+    return undefined;
+  }
+
+  return Math.max(...retryDelaysMs);
+}
+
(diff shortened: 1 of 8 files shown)


packages/engine/src/lib/helper/error-handling.ts
```diff
@@ -139,6 +148,87 @@ const executionFailedWithRetryableError = (
...
+): number {
+  const retryMetadata = getRetryMetadata(executionState, action);
+  if (retryMetadata?.type === AZURE_429_RETRY_TYPE) {
+    return (
+      sanitizeRetryAfterMs(retryMetadata.retryAfterMs) ??
+      DEFAULT_AZURE_429_RETRY_DELAY_MS
+    );
+  }
+
+  return (
```
ravikiranvm, line 183: Except for Azure 429 like headers, existing retry behaviour is not changed.

packages/engine/src/lib/helper/error-handling.ts
```diff
@@ -139,6 +148,83 @@ const executionFailedWithRetryableError = (
   return flowExecutorContext.verdict === ExecutionVerdict.FAILED;
 };
 
+export function getBlockRetryMetadata(
+  action: BlockAction,
+  error: unknown,
+): StepRetryMetadata | undefined {
+  if (!isAzureCustomApiAction(action) || !isHttpErrorLike(error)) {
```
ravikiranvm, line 155: Marco said that the 429 retry behaviour would only be required for Azure block actions. And for now, we're only adding that to Azure Custom API Action.

packages/blocks/common/src/lib/http/core/azure-retry-delay.ts
```diff
@@ -0,0 +1,67 @@
+import { HttpHeaders } from './http-headers';
+
+const MAX_RETRY_DELAY_MS = 10 * 60 * 1000;
+const COST_MANAGEMENT_RETRY_AFTER_HEADER_PATTERN =
+  /^x-ms-ratelimit-microsoft\.costmanagement.*-retry-after$/i;
```
ravikiranvm, line 5: Cost Management endpoint returned different types of retry headers patterns like this. So we're using regex.

packages/blocks/common/src/lib/http/core/azure-retry-delay.ts
```diff
@@ -0,0 +1,67 @@
...
+
+const AZURE_RETRY_AFTER_HEADERS = new Set([
```
ravikiranvm, line 7: These are the headers we noticed while testing by Dariusz

packages/blocks/common/src/lib/http/core/azure-retry-delay.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const retryDelaysMs = retryValues
+    .map((value) => parseRetryDelayMs(value))
+    .filter((value): value is number => value !== undefined);
+
+  if (retryDelaysMs.length === 0) {
+    return undefined;
+  }
+
+  return Math.max(...retryDelaysMs);
```
ravikiranvm, line 42: If a response has many retry headers (entity, subscription, tenant) we pick the max delay time and use it.

packages/engine/src/lib/handler/block-executor.ts
```diff
@@ -277,6 +278,7 @@ const executeAction: ActionHandler<BlockAction> = async ({
       .setVerdict(ExecutionVerdict.RUNNING, undefined);
   } catch (e) {
     const handledError = handleExecutionError(e);
+    const retryMetadata = getBlockRetryMetadata(action, e);
```
ravikiranvm, line 281: We include this retryMetadata in the failed step output.

bigfluffycookie (approved):

PR #2287, opened 2026-05-12
Fix helper in engine's error handler

ravikiranvm:
Fixes OPS-4255.


packages/engine/src/lib/helper/error-handling.ts
```diff
@@ -1,6 +1,6 @@
 import { SharedSystemProp, system } from '@openops/server-shared';
 import type { StepRetryMetadata } from '@openops/shared';
-import { ActionType, BlockAction, CodeAction, isObject } from '@openops/shared';
```
ravikiranvm, line 3: I used this existing `isObject` shared helper func and later didn't test the behaviour locally. Now when I tested it, I've observed that for our use-case this helper is not helpful. We need to check for a class instance and not a plain objct.

bigfluffycookie (approved):

PR #2292, opened 2026-05-14
Fix missing test step viewer labels

ravikiranvm:
Fixes OPS-4318

## Additional Notes

From `dev` env

Screenshot 2026-05-14 at 5 10 37 PM
Screenshot 2026-05-14 at 5 10 49 PM

PR #2293, opened 2026-05-14
Update translation bundle and declare sideEffects in ui-components

alexandrudanpop:
Fixes OPS-4320


alexandrudanpop:
Verified the fix mechanism with a controlled 4-way build comparison.

**Setup:** Built `react-ui` four times — clearing both `dist/packages/react-ui` and `node_modules/.vite/packages/react-ui` between runs — toggling two independent variables:
- `"sideEffects": ["*.css", "*.scss"]` declared in `packages/ui-components/package.json` (yes/no)
- The in-component `useTranslation()` fix from #2292 (yes/no)

**Results:**

| Build | `sideEffects` | Hook fix | # JS chunks | Entry chunk | Where `test-step-data-viewer` lands |
|-------|---------------|----------|-------------|-------------|--------------------------------------|
| A | ✅ | ✅ | 117 | 2,940 KB | inside entry chunk |
| B | ❌ | ✅ | 120 | 1,735 KB | separate `cloud-user-api-*.js` (1,329 KB), static import |
| C | ✅ | ❌ | 117 | 2,938 KB | inside entry chunk |
| D | ❌ | ❌ | 120 | 1,736 KB | separate `cloud-user-api-*.js` (1,326 KB), static import |

\`sideEffects\` alone reshapes the chunk graph: rolldown stops splitting \`ui-components\` modules into a 1.3 MB static-dependency chunk and inlines them into the entry chunk body.

**Why that fixes OPS-4318:** In Build D, the entry chunk starts with \`import { ... } from "./cloud-user-api-*.js"\`. Per ESM semantics, that dep chunk's body evaluates *before* the entry chunk's body. Its body contains \`test-step-data-viewer.tsx\`'s module-scope \`const VIEW_INPUT_OPTIONS = [{ label: t('Input'), … }]\`, which runs while \`i18next.translator\` is still undefined → \`t()\` returns \`undefined\` → labels are frozen as missing. The entry body then runs \`i18next.init()\` — too late.

In Build C (this PR, hook fix reverted), rolldown inlines \`test-step-data-viewer.tsx\` into the entry chunk body. Inside that body, modules evaluate in \`main.tsx\`'s declaration order: \`./i18n\` → App → transitive \`test-step-data-viewer\`. By the time \`t('Input')\` runs, \`translator\` exists and returns the key fallback \`"Input"\`. Bug gone.

**Why the array form, not \`false\`:** \`ui-components\` does have real side effects — \`src/tailwind.css\`, \`src/styles/global.css\`, and \`import '@xyflow/react/dist/style.css'\` in \`flow-canvas.tsx\`. Confirmed post-build that all three are still in the output CSS (xyflow rules present, total CSS unchanged at ~112 KB).

**Lint / tests / build (\`--skip-nx-cache\`):**
- \`npx nx lint ui-components\` — 0 errors
- \`npx nx test ui-components\` — 188/188
- \`npx nx lint react-ui\` — 0 errors
- \`npx nx build react-ui\` — 3.05s, no new warnings

This is defense-in-depth alongside #2292: the hook fix neutralizes one offender; \`sideEffects\` prevents rolldown's chunk-splitting heuristic from re-introducing the same race for any other module-scope \`t()\` in \`ui-components\`.

packages/react-ui/public/locales/en/translation.json
ravikiranvm, line 1: I think we decided to not make changes to this file to avoid sync conflicts in internal repo. Do you still want to update this file?

ravikiranvm (approved):

PR #2298, opened 2026-05-17
Add action for bulk update rows to tables block

bigfluffycookie:
Fixes OPS-4325

### DIFF
--- a/packages/blocks/openops-tables/src/actions/update-records-batch-action.ts
+++ b/packages/blocks/openops-tables/src/actions/update-records-batch-action.ts
@@ -2,7 +2,6 @@ import { BlockAuth, createAction, Property } from '@openops/blocks-framework';
 import {
   batchUpdateRows,
   getFields,
-  getPrimaryKeyFieldFromFields,
   getTableIdByTableName,
   OpenOpsField,
   openopsTablesDropdownProperty,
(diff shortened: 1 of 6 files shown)


packages/blocks/openops-tables/src/actions/update-records-batch-action.ts
```diff
@@ -0,0 +1,81 @@
...
+  resolveTokenProvider,
+  TokenOrResolver,
+} from '@openops/common';
+import { cacheWrapper } from '@openops/server-shared';
+
+export const updateRecordsBatchAction = createAction({
+  auth: BlockAuth.None(),
+  name: 'update_records_batch',
+  description:
+    'Update multiple existing records in an OpenOps table. Note: rowId must be the internal Baserow row ID (integer from the "ID" field), not a custom primary key.',
```
ravikiranvm: I think we don't need the 'Note' here. It looks too descriptive in the UI. 

Image

Maybe the description of 'item' property is enough or if that's not clear enough, maybe make it more clear.

packages/blocks/openops-tables/src/actions/update-records-batch-action.ts
```diff
@@ -0,0 +1,81 @@
...
+  displayName: 'Update Records Batch',
+  isWriteAction: true,
+  props: {
+    tableName: openopsTablesDropdownProperty(),
+    items: Property.Json({
+      displayName: 'Items',
+      required: true,
+      description:
+        'An array of objects with rowId (Baserow internal row ID as integer) and fields keyed by table field names. Example: [{ rowId: 123, fields: { Owner: "user@example.com" } }]',
```
ravikiranvm, line 25: How do we get the Baserow internal row ID?

bigfluffycookie: Its automatically included when you get a baserow object, so when we do get records its already there

ravikiranvm (approved):

PR #2312, opened 2026-05-26
Include inherited ServiceNow table fields in field lookup

ravikiranvm:
Fixes OPS-4414 and OPS-4431

## Problem
I was not able to see some service now table fields like `Description` and `Short Description`. It was due to that these fields were inherited from parent tables and we were looking for only the fields defined in the given table.

ServiceNow doesn't have an endpoint to return all the table fields their UI shows. There is one undocumented endpoint that returns all table fields that are shown on ServiceNow UI `/ui/meta/{table} `. But it is not officially available in their docs. 

The current PR approach is how people usually solve this problem. Look [ServiceNow REST API: Get list of column names](https://stackoverflow.com/questions/56994277/servicenow-rest-api-get-list-of-column-names)

## Fix
- Walk the ServiceNow `sys_db_object` table hierarchy (the table of all tables in SN) when loading dictionary fields
- Merge inherited fields from parent tables so child tables (e.g. `incident`) expose fields like `description`
- Deduplicate by element name, preferring the most specific table in the hierarchy


packages/blocks/servicenow/src/lib/table-dropdown-property.ts
```diff
@@ -20,9 +20,7 @@ export function servicenowTableDropdownProperty(): DropdownProperty<
       }
 
       try {
-        const tables = await getServiceNowTables(auth as ServiceNowAuth, {
-          query: 'nameSTARTSWITHx_',
-        });
+        const tables = await getServiceNowTables(auth as ServiceNowAuth);
```
ravikiranvm, line 23: This reverts the change committed in https://github.com/openops-cloud/openops/pull/2259/changes#diff-46cca212ec4c6c22b05d80a485896ee77a3eae825266ba8f602e1d8e27248b8cL24. This fixes the tables dropdown issue and fixes [OPS-4431](https://linear.app/openops/issue/OPS-4431/table-names-are-not-showing-up-in-the-tables-dropdown-property-in)

packages/openops/src/lib/servicenow/get-table-fields.ts
```diff
@@ -19,10 +20,141 @@ export type ServiceNowTableField = {
...
 export async function getServiceNowTableFields(
   auth: ServiceNowAuth,
   tableName: string,
 ): Promise<ServiceNowTableField[]> {
+  let tableNames: string[];
+
+  try {
+    tableNames = await getServiceNowTableHierarchy(auth, tableName);
+  } catch (error) {
+    tableNames = [tableName];
```
ravikiranvm, line 158: Marco is not sure yet of what table would we have permissions for. 
So if the user doesn't have permission to the 'sys_db_object', we fall-back to the original approach.

packages/blocks/servicenow/src/actions/create-fields-properties.ts
```diff
@@ -39,6 +39,8 @@ async function createFieldsProperties(
       properties: {
         fieldName: Property.StaticDropdown<string>({
           displayName: 'Field name',
+          description:
+            'Includes inherited fields from parent tables if your account has read access to sys_db_object.',
```
ravikiranvm: This is to let user know if he is confused about the missing fields.

MarceloRGonc (approved):

PR #2316, opened 2026-05-29
Move ServiceNow choice values to shared and add state fields helper

ravikiranvm:
Moves `getServiceNowChoiceValues` from the ServiceNow block into `@openops/common` so it can be reused across packages, and adds `getServiceNowStateFields` a helper that returns the columns on a given table (including parent-table columns) usable as a status field.

Part of OPS-4424


packages/openops/src/lib/servicenow/get-state-fields.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Returns the columns on the given table (including parent-table columns)
+ * that are usable as a status field — i.e. those whose values come from a
+ * fixed choice list, so we can show them in a dropdown and reliably map
+ * polled values back to OpenOps statuses.
+ *
+ * Picks columns whose `internal_type` is `choice`, or whose sys_dictionary
+ * `choice` flag is `'1'` (dropdown) or `'3'` (dropdown without --None--).
+ * `'2'` (suggestion) is excluded because it allows free-text values.
+ */
+export async function getServiceNowStateFields(
```
ravikiranvm, line 21: We need this new function to let user pick a column in the table which will be used for mapping OpenOps opportunity state with ticket state in ServiceNow. 

Screenshot 2026-06-01 at 2 53 51 PM

MarceloRGonc (approved):

PR #2415, opened 2026-07-29
Add an OAuth authorization server for external agents

MarceloRGonc:
Fixes OPS-4673.

Adds an OAuth 2.1 authorization server to the Node API so external agents — Claude Code, Codex, Claude.ai/ChatGPT connectors, M365 Copilot, partner CLIs — can hold a self-service, revocable credential instead of sharing one static `AUTH_TOKEN`. Works when SSO is on and password login is disabled.

Off by default: every `/v1/oauth/*` route is unregistered unless `OPS_OAUTH_ENABLED=true`, so this is inert until an operator turns it on.

Design and rationale: `docs/oauth-design.md`. Local runbook: `docs/oauth-manual-testing.md`.

## What's here

**Authorization server** (`packages/server/api/src/app/oauth/`) — PKCE S256-only, exact redirect matching, refresh rotation with family reuse detection, RFC 8707 resource indicators, RFC 8693 token exchange, RFC 7009 revocation, RFC 9207 `iss`, RFC 7591 dynamic registration, RFC 8414 + OIDC-shaped discovery, RFC 9728 protected-resource metadata.

**A dedicated RS256 keypair** with a real JWKS and `kid` dispatch, generated on first boot and stored encrypted. Deliberately isolated from the pre-existing HS256 `OPS_JWT_SECRET` trust domain, so an OAuth token can never be confused with a session token.

**API enforcement** — `extractPrincipal` dispatches on the signing algorithm and enforces the audience *positively*: a token minted for the MCP resource cannot authenticate anywhere in the API, including the websocket path. OAuth principals are `SERVICE`, never `USER`.

**Per-connection grants** — one row per completed authorization, so connecting the same agent twice gives two independently revocable connections. Revocation cascades to that connection's refresh tokens only.

**Project as a selector, not a grant of authority** — every token carries a required `project_id`, re-authorized against membership on every request. A connection can move between the user's projects by naming one when it gets a token (`project_id` on the refresh and token-exchange grants); it can never reach further than the user could in the browser.

**UI** — consent as a dialog over a new **Settings → Connected apps** page, which is also where connections are reviewed and disconnected. Hidden behind a `CONNECTED_APPS_ENABLED` flag when OAuth is off.

**AI chat compatibility** — the stdio spawn now passes the tool allow-list as a file (`OPENOPS_MCP_ROUTES`), matching the rewritten MCP server in `openops-mcp`. `INCLUDED_PATHS` stays the single source of truth for what the chat exposes.

## Additional Notes

This replaces the `feat/mcp-oauth-authentication` spike in `openops-internal`. It is a fresh implementation informed by an adversarial security audit of that spike; `docs/oauth-design.md` lists the findings (H1–H4, M1–M8, L1–L6) and where each is addressed. Highlights:

- Consent was forgeable from URL parameters. It is now a server-side pending-authorization record; the screen renders the client name from the database and references an opaque `request_id`.
- Deny was an open redirect. Both outcomes now go through the server and are validated against registered URIs.
- Refresh reuse went undetected. Tokens form families, and reuse revokes the family.
- One HS256 secret signed everything, advertised behind a `jwks_uri` that served nothing.

Deviations from the design, and cases where the design was wrong, are recorded in the doc rather than quietly resolved — including one reversal: an intermediate decision to build no project switching at all, which turned out to remove the capability rather than defer it to enterprise, because `/switch-project` is unreachable for a `SERVICE` principal.

Two behaviours are load-bearing and pinned by tests, because a plausible-looking change to either would silently make `project_id` decorative:

- the OAuth principal type must stay `SERVICE` (`test/unit/oauth/oauth-principal.test.ts`)
- a refresh must keep the project the connection is currently in, not the one it started in (`test/unit/oauth/tokens.service.test.ts`)

**Follow-ups, not in this PR:** the resource server's project-switch tool and deployment packaging both live in `openops-mcp`. `docs/oauth-design.md` notes that whatever holds the switch selection must not be a process-local map — that is where audit finding M5 came from, and `stateless_http` would make it diverge across replicas.

## Testing Checklist

Automated: 894 server-api unit tests (293 for OAuth), 17 OAuth integration tests that initialise a real database, 504 react-ui, 197 ui-components, 149 shared. Lint and typecheck clean; all affected projects build.

The integration tests exist because the unit tests use in-memory repositories, which cannot show the guarantees that depend on database semantics: that single-use consumption of codes, pending records and refresh tokens is atomic under concurrency, that revocation cascades to one connection only, and that the cleanup job deletes what expired and nothing else. Writing them found a real defect — date cutoffs bound as ISO strings are compared *textually* by drivers that store a different textual format, so a cleanup pass was deleting live rows.

Manual, end to end against a running instance and a real MCP client:

- discovery → dynamic registration → browser consent → code → token → tool call reaching the API as the authorizing user
- an `mcp`-audience token refused by the API (401) and only usable after exchange — the no-token-passthrough rule
- a replayed authorization code, a replayed refresh token (family revoked), and a revoked connection cut off on its next request
- two connections for one client, one revoked, the other still working
- a genuine project switch across two projects, on both the refresh and exchange paths, refused as `invalid_target` for a project the user is not a member of
- a wrong project on refresh not consuming the refresh token, so a bad guess cannot brick a working agent

Backwards compatibility: additive migration, no existing table or column is altered. The MCP spawn variables were renamed, and the rewritten server accepts the previous names as fallbacks, so the two repositories can roll out independently.

## Visual Changes

Two new surfaces, both under **Settings → Connected apps**:

- **The consent dialog**, shown over that page when the URL carries a `request_id`. Names the application, lists what it will be able to do — including that it can act in any project the user has access to — and warns against continuing unless the user started the flow themselves. Cancel and dismissal both deny, so the waiting application is always told the outcome.
- **The connections list**, one row per authorization: application name, how it connects (via the MCP server or the API), when it was connected, when it was last used, and a red Disconnect behind a confirmation. Styled after the enterprise integrations card.

> Screenshots to be attached — they are on the branch author's machine and cannot be uploaded from the terminal.


MarceloRGonc:
## Claude PR Review — #2415: Add an OAuth authorization server for external agents

**Author:** MarceloRGonc | **Base:** `main` | **Labels:** (none) | **Status:** draft

### Verdict

SHIP — no blockers or risks survived validation; the two remaining items are a latent i18n note and a documentation question.

### Summary

Adds an OAuth 2.1 authorization server (PKCE S256-only, RS256/JWKS in a separate trust domain, refresh rotation with family reuse detection, RFC 8693 token exchange) behind `OPS_OAUTH_ENABLED`, plus a Connected Apps settings surface. The deep candidates a review of this shape usually surfaces — the header-over-cookie precedence flip, process-local grant-snapshot cache staleness across replicas, non-transactional grant+token creation — all turned out to be deliberate, documented in `docs/oauth-design.md`, and correctly bounded, so they are listed under "Verified acceptable" rather than as findings.

### Findings

**[NOTE]** `packages/react-ui/src/app/common/components/project-settings-layout.tsx:31` — module-scope `t('Connected apps')`; a new instance of the OPS-4318 anti-pattern (safe today, latent under chunking changes). See inline comment.

**[QUESTION]** `packages/server/api/src/app/oauth/oauth-cleanup-job.ts:63` — `REVOKED_RETENTION_DAYS = 7` vs the 30-day refresh TTL bounds reuse detection to 7 days after rotation; is that window intentional, and worth a sentence in the doc? See inline comment.

### Testing gaps

`docs/oauth-design.md` (L5) says the header-over-cookie precedence flip in `access-token-authn-handler.ts` is covered by "regression tests for the app's cookie-based flows", but nothing in `packages/server/api/test/` exercises `AccessTokenAuthnHandler` or sends a cookie through auth. The flip itself checks out — it is necessary for audience separation, and no in-repo caller sends both a cookie and an Authorization header — but the promised test does not exist yet.

### Verified acceptable (checked, deliberately not raised)

- Cross-replica revocation lag of up to ~60s from the process-local grant-snapshot cache — documented tradeoff in `docs/oauth-design.md`; worst case is one extra stacked cache window, not the full access-token TTL.
- Orphaned active grant if code redemption fails between grant insert and refresh-token insert — cosmetic, cleaned within 30 days by the dead-grant pass.
- `/v1/oauth/grants` N+1 client lookups — bounded by the caller's own grant count, indexed PK reads.
- Unauthenticated writes on `/register` and `/authorize` — rate-limited by default, same posture as the existing sign-up endpoints, bounded by hourly cleanup.

PR #2456, opened 2026-09-01
Forward container credential variables to the AWS CLI

### LINKED ISSUE #2455: AWS CLI blocks fail with NoCredentials under implicit role on ECS/Fargate
AWS CLI blocks cannot use the implicit role on ECS/Fargate (or EKS): the CLI is spawned with an environment that drops the only variables those platforms use to expose a role, so every command fails with `NoCredentials`.

`OPS_AWS_ENABLE_IMPLICIT_ROLE=true`, AWS connection with empty access key / secret, running on ECS Fargate with a task role attached.

## Expected Behavior

`aws compute-optimizer get-ec2-instance-recommendations` runs under the ECS task role, the same way it runs under the instance role on EC2.

## Actual Behavior

```
Failed to run the aws command: 'aws compute-optimizer get-ec2-instance-recommendations'.
Error: {"exitCode":253,"stdOut":"","stdError":"aws: [ERROR]: An error occurred (NoCredentials):
Unable to locate credentials. You can configure credentials by running \"aws login\"."}
```

CloudTrail shows no corresponding API call at all — the request is never signed or sent. SDK-based AWS blocks using the *same* task role work fine, because they run in-process.

## Cause

`packages/blocks/aws/src/lib/actions/cli/aws-cli.ts` builds the child environment from scratch:

```ts
const envVars: any = {
  AWS_DEFAULT_REGION: region,
  PATH: process.env['PATH'] ?? '',
};
```

and `executeFile()` in `packages/openops/src/lib/command-wrapper.ts` **replaces** the environment rather than merging it:

```ts
const options: ExecFileOptions = { env: envVariables };
```

So `aws` sees only `AWS_DEFAULT_REGION`, `PATH` and `HOME`.

On EC2 that is harmless: the CLI reaches the instance role over IMDS at a fixed link-local address and needs nothing from the environment — which is why the feature is documented as an EC2 feature and why nobody has hit this before. ECS and Fargate have no IMDS credential provider; task-role credentials are discoverable **only** via `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` (or `..._FULL_URI` + the authorization token variables). EKS IRSA is equivalent, via `AWS_WEB_IDENTITY_TOKEN_FILE` / `AWS_ROLE_ARN`. All of them are scrubbed.

## Steps to Reproduce

1. Deploy the engine on ECS/Fargate with a task role and `OPS_AWS_ENABLE_IMPLICIT_ROLE=true`.
2. Create an AWS connection leaving access key ID and secret access key empty.
3. Run any workflow using the **AWS CLI** block.

Reproducible directly inside the engine container, which isolates it from anything OpenOps-specific:

```console
# normal environment
$ aws sts get-caller-identity
{ "Arn": "arn:aws:sts::<account>:assumed-role/<task-role>/<task-id>" }        # rc=0

# the environment OpenOps builds for the CLI
$ env -i PATH=$PATH HOME=$HOME AWS_DEFAULT_REGION=us-east-1 aws sts get-caller-identity
aws: [ERROR]: An error occurred (NoCredentials): Unable to locate credentials.  # rc=253
```

## Your environment

Release version: 0.6.25 (`public.ecr.aws/openops/openops-engine:0.6.25`, also present on `main`)
Operating System: ECS Fargate, Linux/x86_64, `us-east-1`
Additional information: `OPS_EXECUTION_MODE=SANDBOX_CODE_ONLY`; engine, app, tables and analytics as separate ECS services

## Additional Context

Verified fix: forwarding `AWS_CONTAINER_CREDENTIALS_*` and `AWS_WEB_IDENTITY_*` from `process.env` when falling back to the implicit role. With a `PATH` shim that re-injects `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` before `exec`ing the real binary, the same failing command returns real recommendations in the same container. PR follows.

AgustinRamiroDiaz:
Fixes #2455.

Under implicit role, the AWS CLI block runs `aws` with an environment that contains only `AWS_DEFAULT_REGION`, `PATH` and `HOME` — `aws-cli.ts` builds `envVars` from scratch and `executeFile()` sets `env` rather than merging it. The CLI therefore has no way to discover the role on any platform that exposes it through the environment, and fails with `NoCredentials` (exit 253).

This forwards the variables the CLI needs for that discovery when, and only when, we fall back to the implicit role:

- `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`, `AWS_CONTAINER_CREDENTIALS_FULL_URI`, `AWS_CONTAINER_AUTHORIZATION_TOKEN`, `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE` — ECS and Fargate task roles
- `AWS_ROLE_ARN`, `AWS_ROLE_SESSION_NAME`, `AWS_WEB_IDENTITY_TOKEN_FILE` — EKS IRSA / Pod Identity

## Additional Notes

**Why it has not surfaced before.** On EC2 the CLI reaches the instance role over IMDS at a fixed link-local address and needs nothing from the environment, which is the deployment shape the feature is documented against. Fargate has no IMDS credential provider, so the scrub is fatal there. The asymmetry is confusing in practice: SDK-based AWS blocks work on the same task role, because they run in-process where the variables are intact — only the CLI blocks fail.

**Scope.** Deliberately narrow. Only the implicit-role branch is touched, only variables that are already in `process.env` are forwarded, and nothing is forwarded when explicit credentials are supplied (they take precedence in the chain anyway, but keeping the environment minimal seemed better than relying on that). A variable absent from `process.env` is not added, so non-container deployments are byte-for-byte unchanged.

**Possibly worth a follow-up, not done here:** `azure-cli.ts` and the Google Cloud CLI wrapper go through the same `executeFile()` and so have the same blind spot for managed identity / workload identity. I have not tested either, so I have left them alone.

## Testing Checklist

Tested on a real ECS Fargate deployment of 0.6.25 (engine as its own service, task role attached, `OPS_AWS_ENABLE_IMPLICIT_ROLE=true`). Inside the engine container, reproducing the environment the block builds:

```console
$ env -i PATH=$PATH HOME=$HOME AWS_DEFAULT_REGION=us-east-1 \
    aws compute-optimizer get-ec2-instance-recommendations
aws: [ERROR]: An error occurred (NoCredentials): Unable to locate credentials.

$ env -i PATH=/tmp/bin:$PATH HOME=$HOME AWS_DEFAULT_REGION=us-east-1 \
    aws compute-optimizer get-ec2-instance-recommendations
{ "instanceRecommendations": [ { "currentInstanceType": "t3.large", "finding": "OPTIMIZED", ... } ] }
```

where `/tmp/bin/aws` is a wrapper that re-exports `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` and execs the real binary — the same variable this PR forwards.

Unit tests: two added (forwarding under implicit role; no forwarding with explicit credentials), and the existing cases pin the unchanged behaviour. The suite clears these variables around each test so a developer machine that has them set cannot leak into the assertions. `packages/blocks/aws`: 176 tests, all passing.

### DIFF
--- a/packages/blocks/aws/src/lib/actions/cli/aws-cli.ts
+++ b/packages/blocks/aws/src/lib/actions/cli/aws-cli.ts
@@ -1,6 +1,41 @@
 import { runCliCommand } from '@openops/common';
 import { SharedSystemProp, system } from '@openops/server-shared';
 
+/**
+ * Variables the AWS CLI uses to find credentials it has not been handed
+ * explicitly. The CLI runs with a replaced environment (see executeFile), so
+ * under implicit role these have to be forwarded deliberately.
+ *
+ * On EC2 nothing needs forwarding: the CLI reaches the instance role over IMDS
+ * at a fixed link-local address. Every other role-based deployment is
+ * discoverable only through the environment -- ECS and Fargate task roles
+ * through the container credentials variables, EKS IRSA through the web
+ * identity ones -- so without this, implicit role fails there with
+ * "Unable to locate credentials".
+ */
+const AMBIENT_CREDENTIAL_ENV_VARS = [
+  'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI',
+  'AWS_CONTAINER_CREDENTIALS_FULL_URI',
+  'AWS_CONTAINER_AUTHORIZATION_TOKEN',
+  'AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE',
+  'AWS_ROLE_ARN',
+  'AWS_ROLE_SESSION_NAME',
+  'AWS_WEB_IDENTITY_TOKEN_FILE',
+];
+
+function getAmbientCredentialEnvVars(): Record<string, string> {
+  const envVars: Record<string, string> = {};
+
+  for (const name of AMBIENT_CREDENTIAL_ENV_VARS) {
+    const value = process.env[name];
+    if (value) {
+      envVars[name] = value;
+    }
+  }
+
+  return envVars;
+}
+
 export async function runCommand(
   command: string,
   region: string,
@@ -26,6 +61,8 @@ export async function runCommand(
     throw new Error(
       'AWS credentials are required, please provide accessKeyId and secretAccessKey',
     );
+  } else {
+    Object.assign(envVars, getAmbientCredentialEnvVars());
   }
 
   return await runCliCommand(command, 'aws', envVars);
--- a/packages/blocks/aws/test/cli/aws-cli.test.ts
+++ b/packages/blocks/aws/test/cli/aws-cli.test.ts
@@ -20,9 +20,38 @@ const credential = {
   sessionToken: 'some token',
 };
 
+const AMBIENT_CREDENTIAL_ENV_VARS = [
+  'AWS_CONTAINER_CREDENTIALS_RELATIVE_URI',
+  'AWS_CONTAINER_CREDENTIALS_FULL_URI',
+  'AWS_CONTAINER_AUTHORIZATION_TOKEN',
+  'AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE',
+  'AWS_ROLE_ARN',
+  'AWS_ROLE_SESSION_NAME',
+  'AWS_WEB_IDENTITY_TOKEN_FILE',
+];
+
 describe('awsCli', () => {
+  // The machine running the tests may itself have some of these set, which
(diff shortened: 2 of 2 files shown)


MarceloRGonc:
As discussed in the issue (https://github.com/openops-cloud/openops/issues/2455), we will not add this behaviour. See https://github.com/openops-cloud/openops/issues/2455#issuecomment-5508458713

